PR #11, opened 2017-08-31
NEPT-1103: Update ECL library to 0.12.0 #7

ademarco:
#7


voidtek:
The laster version of ECL library is "0.12.0".
It includes a lot of new things.

ademarco:
@voidtek thanks for the hint! I've upgraded it to `0.12.0`, I'll keep the version fixed instead of `<1.0.0` so that we can better control component upgrade, since a new ECL version is still luckily to break the theme  at this stage.

PR #19, opened 2018-02-07
OPENEUROPA-12: Include Europa Component Library Twig loader as composer dependency.

drupol:


drupol:
@ademarco , I guess we need to merge the task runner in before having green light in Travis.

modules/oe_theme_helper/oe_theme_helper.services.yml
```diff
@@ -0,0 +1,10 @@
+parameters:
+  oe_theme_helper.twig_loader.namespaces: ['ec-europa', 'ecl']
+  oe_theme_helper.twig_loader.theme: 'oe_theme'
+  oe_theme_helper.twig_loader.directory: 'templates/components'
+services:
+  oe_theme_helper.twig_loader.component_library:
+    class: Drupal\oe_theme_helper\Loader\ComponentLibraryLoader
+    arguments: ['%oe_theme_helper.twig_loader.namespaces%', '@app.root', '%oe_theme_helper.twig_loader.theme%', '%oe_theme_helper.twig_loader.directory%', '@theme_handler']
+    tags:
+      - { name: twig.loader, priority: -100 }
```
dxvargas: It's missing a new line at the end.

composer.json
```diff
@@ -6,10 +6,11 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "ec-europa/oe_core": "*"
+        "ec-europa/oe_core": "*",
```
dxvargas: We should remove the dependency from oe_core, right?

runner.yml.dist
```diff
@@ -8,7 +8,7 @@ drupal:
     user: "root"
     password: ""
   post_install:
-    - "./vendor/bin/drush en oe_core oe_theme_test -y"
+    - "./vendor/bin/drush en oe_theme_test oe_theme_helper -y"
```
dxvargas: I wouldn't always enable oe_theme_test. I understand that for development phase it's handy, but still I would keep it out.

composer.json
ademarco: Yes that's correct, we should remove this.

runner.yml.dist
ademarco: Actually about this: we should consider patching core with [this patch](https://www.drupal.org/project/drupal/issues/474684#comment-12419246) so that we can just list the theme helper module as a dependency of the theme and having it enabled when enabling the theme itself (i.e. we could then remove the change above).

aritomelo: Ok We will address them and check if the patch can be applied.

modules/oe_theme_helper/oe_theme_helper.services.yml
drupol: It hasn't been detected by oe-code-review yet, I opened an issue about this: https://github.com/ec-europa/oe-code-review/issues/16

runner.yml.dist
drupol: Patch tested and validated: https://www.drupal.org/project/drupal/issues/474684#comment-12470938

modules/oe_theme_helper/oe_theme_helper.services.yml
drupol: In the meantime, I fixed it.

PR #20, opened 2018-02-07
OPENEUROPA-194: Added navigation components

imanoleguskiza:
Styled primary and secondary local tasks using ECL lists navigation component.


ademarco:
@imanoleguskiza you can rebase this on current master since #18  has been merged.

ademarco:
We can now add Behat tests in this way:

1. check if tabs and secondary tabs are available in [oe_theme_test.theme_preview.yml](https://github.com/ec-europa/oe_theme/blob/master/modules/oe_theme_test/oe_theme_test.theme_preview.yml), if not we add them
2. Add a Behat scenario that will assert that ECL tabs markup is present on that page

That would be a good enough bases for now.

oe_theme.theme
```diff
@@ -4,3 +4,33 @@
  * @file
  * Functions to support theming.
  */
+
+/**
+ * Implements template_preprocess_menu_local_tasks().
+ */
+function oe_theme_preprocess_menu_local_tasks(&$variables) {
+
```
drupol: Extraneous empty line.

PR #23, opened 2018-02-11
OPENEUROPA-193: Add header and related default blocks

ademarco:


behat.yml.dist
```diff
@@ -0,0 +1,30 @@
...
+  suites:
+    default:
+      paths:
+        - %paths.base%/tests/features
+      contexts:
+        - Drupal\DrupalExtension\Context\MinkContext
+        - Drupal\DrupalExtension\Context\DrupalContext
+        - Drupal\DrupalExtension\Context\MarkupContext
+        - Drupal\Tests\oe_theme\Behat\TransformationContext:
+            elements:
```
imanoleguskiza: I'm a bit confused at why do we need to create our own transformation context. Doesn't the drupal-extension already provide the same feature?

ademarco: Not exactly: Drupal extension allows to specify regions, here we are providing human readable names for elements the presence of which needs to be checked within a region.

netlooker:
+1

PR #26, opened 2018-02-12
OPENEUROPA-163: Integrate form-related ECL components

ademarco:


templates/form/form.html.twig
```diff
@@ -0,0 +1,15 @@
...
+ * Theme override for a 'form' element.
+ *
+ * Available variables
+ * - attributes: A list of HTML attributes for the wrapper element.
+ * - children: The child elements of the form.
+ *
+ * @see template_preprocess_form()
+ */
+#}
+<form{{ attributes.addClass() }}>
```
ademarco: Are we missing a class here?

templates/form/input--checkbox.html.twig
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Available variables:
+ * - attributes: A list of HTML attributes for the input element.
+ * - children: Optional additional rendered elements.
+ *
+ * @see template_preprocess_input()
+ */
+#}
+
+<input{{ attributes.addClass('ecl-checkbox__input ecl-u-sr-only') }} />{{ children }}
```
ademarco: We chain two `.addClass()` when more than one class is being added.

templates/form/input--radio.html.twig
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Available variables:
+ * - attributes: A list of HTML attributes for the input element.
+ * - children: Optional additional rendered elements.
+ *
+ * @see template_preprocess_input()
+ */
+#}
+
+<input{{ attributes.addClass('ecl-radio__input ecl-u-sr-only') }} />{{ children }}
```
ademarco: We chain two `.addClass()` when more than one class is being added.

ademarco:
Looks good, let's make sure we cover with tests what you did then you can leave out the rest, like radio, and open a follow up on Jire. Thanks!

ademarco:
Password field does not seems to be styled yet, we should also fix spaces between input fields and labels and between form elements:

image

PR #29, opened 2018-02-14
OPENEUROPA-301: Add header placeholder blocks to demo module

ademarco:


modules/oe_theme_demo/oe_theme_demo.module
```diff
@@ -11,3 +11,12 @@
 function oe_theme_demo_theme() {
   return \Drupal::service('plugin.manager.oe_theme_demo.demo_blocks')->hookTheme();
 }
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function oe_theme_demo_theme_suggestions_block_alter(array &$suggestions, array $variables) {
+  if (isset($variables['elements']['#base_plugin_id']) && $variables['elements']['#base_plugin_id'] === 'oe_theme_demo_block') {
+    $suggestions[] = 'block__bare';
```
imanoleguskiza: What is the use of adding this suggestion?

ademarco: That's to make sure we loose the wrapping DIVs from the demo blocks.

netlooker:
👍

PR #30, opened 2018-02-16
OPENEUROPA-309: Theme "Links block" paragraph type in OE Theme

ademarco:


tests/Kernel/ParagraphsTest.php
```diff
@@ -0,0 +1,93 @@
...
+    $this->installSchema('system', ['sequences']);
+    $this->installConfig(['oe_paragraphs']);
+  }
+
+  /**
+   * Test links block paragraph rendering.
+   */
+  public function testLinksBlock() {
+    $paragraph = Paragraph::create(['type' => 'oe_paragraphs_links_block']);
+    $paragraph->field_oe_paragraphs_text = 'Title';
```
brummbar: To improve readability, we can move these field values inside the array specified in the `create()` method above.

PR #39, opened 2018-02-23
OPENEUROPA-326: Theme "Accordion" paragraph type

ademarco:


netlooker:
I've noticed at least twice that the solution that we are using for scaffolding is not yet fully stable. I had to run the Travis CI once again manually to get a green light. Also on my machine, I had some small issues which I solved by repeating some steps.

PR #40, opened 2018-02-23
OPENEUROPA-318: integrate messages components

brummbar:


tests/Kernel/StatusMessagesTest.php
```diff
@@ -0,0 +1,208 @@
...
+
+    foreach ($data as $type => $messages) {
+      $modifier = self::MESSAGE_TYPES[$type]['modifier'];
+      $heading = self::MESSAGE_TYPES[$type]['heading'];
+
+      $wrapper = $crawler->filter('div.ecl-message--' . $modifier);
+      $this->assertCount(1, $wrapper, sprintf('Wrong number of wrappers found for "%s" messages.', $type));
+
+      $title = $wrapper->filter('div.ecl-message__title');
+      $this->assertCount(1, $title, sprintf('Wrong number of headings found for "%s" messages.', $type));
```
ademarco: This is nice, could we port this to the other tests as well? Like now we do:

```
    // Assert list contains ECL classes.
    $actual = $crawler->filter('ul.ecl-navigation-list.ecl-navigation-list--tabs')->count();
    $this->assertEquals(1, $actual);
```
While it could actually simply be
```
    $this->assertCount(1, $actual);
```

drupol: Comment was adressed in the last commit.

tests/Kernel/StatusMessagesTest.php
```diff
@@ -63,10 +63,10 @@ public function testStatusMessages(array $data): void {
       $heading = self::MESSAGE_TYPES[$type]['heading'];
 
       $wrapper = $crawler->filter('div.ecl-message--' . $modifier);
-      $this->assertCount(1, $wrapper, sprintf('Wrong number of wrappers found for "%s" messages.', $type));
+      $this->assertEquals(1, $wrapper->count(), sprintf('Wrong number of wrappers found for "%s" messages.', $type));
 
       $title = $wrapper->filter('div.ecl-message__title');
-      $this->assertCount(1, $title, sprintf('Wrong number of headings found for "%s" messages.', $type));
+      $this->assertEquals(1, $title->count(), sprintf('Wrong number of headings found for "%s" messages.', $type));
```
ademarco: Ah no sorry! I meant the contrary: to convert all other tests to use `assertCount()`, not to remove it from here.

PR #41, opened 2018-02-26
OPENEUROPA-329: Theme Quote paragraph.

aritomelo:
- Theming and create test for Quote paragraph


oe_theme.theme
```diff
@@ -66,6 +66,14 @@ function oe_theme_preprocess_paragraph__oe_links_block(&$variables) {
   }
 }
 
+/**
+ * Implements hook_preprocess_paragraph().
+ */
+function oe_theme_preprocess_paragraph__oe_quote(&$variables) {
+	// Temporary solution to avoid render issue with ECL component that expecting a string.
+  $variables['quoteBody'] = $variables['content']['field_oe_text_long'][0]['#text'];
```
ademarco: Preprocess is not necessary, we can just render `content.field_oe_text_long` in the template.

templates/paragraphs/paragraph--oe-quote.html.twig
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @see template_preprocess_paragraph()
+ *
+ * @ingroup themeable
+ */
+#}
+{% include '@ecl/ecl-blockquotes' with {
+    'body': quoteBody,
+    'author': content.field_oe_text,
+} %}
```
ademarco: Until INNO-942 let's use here the actual twig template, as in:

1. We don't use `include`
2. We add here the actual Twig template from `ecl-blockquote`, maybe simplifying it a bit
3. We add a todo comment linking to INNO-942

oe_theme.theme
aritomelo: @ademarco I do agree with you, however, the preprocessor was just for testing while we were solving the ECL rending issue when the data is an array type.

Now that we are using the plain text I will test and the preprocessor will be removed.

tests/Kernel/ParagraphsTest.php
```diff
@@ -136,6 +137,31 @@ public function testAccordions() {
     $this->assertEquals(1, $actual);
   }
 
+  /**
+   * Test quote paragraph rendering.
+   */
+  public function testQuotes() {
+
+    $attribution = 'Quote author goes here';
+    $body = 'Quote body goes here';
```
drupol: I would prefer to use a dataprovider in this case, I think it would be better to do that all the time to avoid mixing the logic of the test AND the test case data.
@ademarco what's your point of view on this ?

aritomelo: @ademarco and @drupol I fine to create a new public method for the data provider, however, it seems me that it is useful only when the purpose is to test several cases. Let's discuss it...

drupol: This is true as well.
I was higlighting this in a matter of code consistency.
I leave @ademarco decide on this.

PR #42, opened 2018-03-07
OPENEUROPA-341: Create theme overview page.

imanoleguskiza:


runner.yml.dist
```diff
@@ -8,7 +8,7 @@ drupal:
     user: "root"
     password: ""
   post_install:
-    - "./vendor/bin/drush en oe_theme_test -y"
+    - "./vendor/bin/drush en oe_theme_demo -y"
```
brummbar: The `oe_theme_demo` module has configuration that requires the `oe_theme` theme to be already installed, so this line should be moved after line 14.

modules/oe_theme_demo/src/Plugin/Styleguide/EclStyleGuide.php
```diff
@@ -0,0 +1,111 @@
...
+   * {@inheritdoc}
+   */
+  public function items() {
+    $items = [];
+    if ($this->moduleHandler->moduleExists('oe_paragraphs')) {
+
+      // Prepare quote paragraph.
+      $paragraph = Paragraph::create([
+        'type' => 'oe_quote',
+        'field_oe_text' => 'Quote author',
```
brummbar: The `$generator` service is used to generate random data. If we don't use it (one idea would be that we can't use it if we want to run visual regression on this page) then we have to remove it from the injected dependencies. Otherwise use here `$this->generator->sentence()` for example.

modules/oe_theme_demo/src/Plugin/Styleguide/EclStyleGuide.php
```diff
@@ -0,0 +1,111 @@
...
+      $container->get('module_handler')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function items() {
+    $items = [];
+    if ($this->moduleHandler->moduleExists('oe_paragraphs')) {
```
brummbar: Maybe we can add the dependency to the `oe_paragraphs` module? Being a demo module, it makes sense.

modules/oe_theme_demo/src/Plugin/Styleguide/EclStyleGuide.php
```diff
@@ -0,0 +1,111 @@
...
+
+      // Prepare quote paragraph.
+      $paragraph = Paragraph::create([
+        'type' => 'oe_quote',
+        'field_oe_text' => 'Quote author',
+        'field_oe_text_long' => 'Quote body',
+      ]);
+      $paragraph->save();
```
brummbar: It seems that you don't need to save the paragraph in order to render it.

modules/oe_theme_demo/oe_theme_demo.links.menu.yml
```diff
@@ -65,3 +65,8 @@ oe_theme_demo.main.contacts:
   route_name: "<none>"
   menu_name: "main"
   weight: 3
+oe_theme_demo.main.styleguide:
```
ademarco: Nice one! :+1:

PR #48, opened 2018-03-12
OPENEUROPA-366: Theming radio and checkbox

aritomelo:
- To use ECL checkbox and radio component was needed to use the preprocessor to render the label as a parent (wrapper) for the entire checkbox markup

   - Also to hide the label, it was created a new template and register it using hook suggestion alter

   - An INNO ticket has been open, to allow extra attributes, once DRUPAL requires some extra classes not available on the current version, those class will allow full javascript functionality

    - Kernel test


templates/form/input--radio.html.twig
```diff
@@ -10,5 +10,10 @@
...
 #}
-
-<input{{ attributes.addClass('ecl-radio__input').addClass('ecl-u-sr-only') }} />{{ children }}
+{% include '@ecl/ecl-forms-radios' with {
+    'id': attributes.id,
+    'name': attributes.name,
+    'value': attributes.value,
+    'checked': attributes.checked,
+    'label': title
+} %}
```
voidtek: According to https://ec-europa.github.io/europa-component-library/components/detail/ecl-forms-radios, some parameters are missing:
    - "is_disabled" (boolean): is disabled (optional, default: false)
    - "has_error" (boolean): has error (optional, default: false)
    - "extra_class" (string): extra CSS classes (optional, default: '')

templates/form/input--checkbox.html.twig
```diff
@@ -11,4 +11,10 @@
...
 #}
 
-<input{{ attributes.addClass('ecl-checkbox__input').addClass('ecl-u-sr-only') }} />{{ children }}
+{% include '@ecl/ecl-forms-checkboxes' with {
+    'id': attributes.id,
+    'name': attributes.name,
+    'value': attributes.value,
+    'checked': attributes.checked,
+    'label': title
+} %}
```
voidtek: According to https://ec-europa.github.io/europa-component-library/components/detail/ecl-forms-checkboxes, some parameters are missing:
    - "is_disabled" (boolean): is disabled (optional, default: false)
    - "has_error" (boolean): has error (optional, default: false)
    - "extra_class" (string): extra CSS classes (optional, default: '')

voidtek:
The phpunit tests for all states ('checked', 'disabled', 'error') for the checkboxes and the radios are missing.

oe_theme.theme
```diff
@@ -116,3 +116,52 @@ function oe_theme_preprocess_menu_local_tasks(&$variables) {
...
+ *   Set of available variables.
+ */
+function _oe_theme_preprocess_input_label_wrapper(array &$variables) {
+  $element = &$variables['element'];
+
+  if (isset($element['#title']) && $element['#title'] !== '') {
+    $variables['title'] = ['#markup' => $element['#title']];
+    $variables['attributes']['checked'] = isset($element['#checked']) ? $element['#checked'] : FALSE;
+    $variables['has_error'] = isset($element['#errors']) ? $element['#errors'] : FALSE;
+    $variables['is_disabled'] = isset($element['#disabled']) ? $element['#disabled'] : FALSE;
```
voidtek: These lines are duplicated.

templates/form/input--radio.html.twig
```diff
@@ -6,9 +6,20 @@
...
  */
 #}
-
-<input{{ attributes.addClass('ecl-radio__input').addClass('ecl-u-sr-only') }} />{{ children }}
+{% include '@ecl/ecl-forms-radios' with {
+    'id': attributes.id,
+    'name': attributes.name,
+    'value': attributes.value,
+    'checked': attributes.checked is defined,
+    'is_disabled': attributes.disabled is defined,
```
drupol: Nice!

oe_theme.theme
```diff
@@ -116,3 +116,46 @@ function oe_theme_preprocess_menu_local_tasks(&$variables) {
...
+ * Implements hook_preprocess_input__radio().
+ */
+function oe_theme_preprocess_input__radio(&$variables) {
+  _oe_theme_preprocess_input_label_wrapper($variables);
+}
+
+/**
+ * Implements hook_preprocess_input__checkbox().
+ */
+function oe_theme_preprocess_input__checkbox(&$variables) {
```
drupol: If the label processing is the same for all form elements, can't we use `oe_theme_preprocess_input(&$variables)` ?

tests/Kernel/RenderingTest.php
```diff
@@ -43,7 +44,26 @@ public function buildForm(array $form, FormStateInterface $form_state, array $st
...
+        $form_state->setError($element, t('Validation error on @label', ['@label' => $label]));
+      }
+
+      foreach (Element::children($element) as $key) {
+        // Recursively call this closure on all the children elements.
+        $add_errors($element[$key]);
+      }
+    };
+
+    $add_errors($form['test']);
```
drupol: I like it :)

oe_theme.theme
brummbar: We need to move the label element to the input template only for checkbox and radio elements. Truth is that the name `_oe_theme_preprocess_input_label_wrapper()` is not the best to represent this common function, but I couldn't come with anything better to improve it.

drupol: The function `oe_theme_preprocess_input__[something](&$variables)` is a variant of `oe_theme_preprocess_input(&$variables)`, this is why I was wondering if all the inputs had to be updated.
If it was the case, just naming the function `oe_theme_preprocess_input(&$variables)` would have been enough to alter them all.
It's also a good news that Drupal is using variant in theme calls, it's actually a very good news.

Regarding the name of the common function, I don't have a better candidate, I've always been very bad choosing arbitrary names :)

PR #49, opened 2018-03-13
OPENEUROPA-305: Provide a release package including compiled CSS.

pfrenssen:


.travis.yml
```diff
@@ -37,3 +37,16 @@ script:
...
 notifications:
   email: false
+
+before_deploy:
+  ./scripts/travis-ci/create-release-tarball.sh
+
+deploy:
+  provider: releases
+  api_key:
+    secure: JGvlz3E89uDsv/CMspHrFvkGUyc7Bs+qEFYIt01V5DCJ4DIc7KDiAt2zE8KHrw8M5Q5sXkbs6kQfvuB+XY0aMP73vlHdKzg+ZNpUu+Ikmt8uOt8UYZtQQkwOZcD2hsTj/1SzDb+cARQEW7hLDQrp5qL9y6cuPS9aplusY+qaepHPPixGgROpZ/uuj/mtFYJwdN++RS5H33KqhkrDGjbtJo0j1j78l5nSoeXIxqJ9cE2pf2Cb8CZsN8wEiJycLKbI3LKitc7WCODNTvwk2SKUTYm3F8WtQjfBbbghjHn6zFOXucMX1raRJ40LpLNEvd3/3SENWIL0QRF5PQcudZjh9G8Y2pHcSky+rFE9/wCejOKj3Ug54kj9Hsa3Uyqjlyyk0v6xNT8NVpCZsDCstrhuhcHgdWNBBPGQyNc9O+DxVu+VMFUFXFc8lXzLWE+nePBv01WDlv0ywPoiLNN/j68FZNrgbIfsX+56jqQY1Rh3FCtsGpWsYkixQsUsR4hHqdG+9atZGMRmAITksWTMsjVapBZxVf0PB1MuCQSsobTMdi/p+XQXTgDfEucy3qTJNjWFj7oDg8DD+AnsCwwjHMl6FO6loTK5k+ji7ArRHKXxIpm977NkdFGzLCUc2M2/SJ3EylJjadfv43gNgZOz4hje8NJ43kgxKS0r3naVIqppa3I=
```
pfrenssen: This token will need to be replaced with one generated by the repository owner.

How to do this:
* Install the [Travis CLI](https://github.com/travis-ci/travis.rb).
* Run the following command:
```
$ travis setup releases --force
```
* Follow the instructions. You will be asked for your Github username and password.
* Check the `.travis.yml` file. It will be updated with the new encrypted key in the `deploy` section.
* The file will be mangled, so don't commit it, just copy the new encrypted key, undo all the changes (`git checkout .travis.yml`) and paste the new key in the original file.

scripts/travis-ci/create-release-tarball.sh
```diff
@@ -0,0 +1,57 @@
...
+# Create release folder.
+mkdir ${RELEASE_PATH}
+
+# Undo vandalism to composer.json.
+git checkout composer.json
+rm composer.lock
+
+# Install production dependencies.
+# Todo: Not needed if we exclude the vendor folder.
+sudo chmod -R a+w build/
```
pfrenssen: This code block downloads the production dependencies so they can be included in the archive. I put this here for completeness and because it is similar to how Drupal core is packaged.

However, I personally think having the dependencies inside the theme is not needed. Since we are using a composer based workflow in all our projects the dependencies will be automatically downloaded. I don't even know if the old Composer Manager workflow with separate vendor folders per project is still supported in Drupal 8.

Let's discuss this, and we can decide to keep it or remove it.

scripts/travis-ci/create-release-tarball.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Copy production files.
+# Todo: exclude the vendor folder?
+cp -r \
+  config/ \
+  css/ \
+  js/ \
+  modules/ \
+  templates/ \
+  vendor/ \
```
pfrenssen: If we decide to not keep the locally installed dependencies then this line should also be removed.

pfrenssen:
To see the packaging in action, I have been doing dry runs in my fork while working on this: https://github.com/pfrenssen/oe_theme/releases

The latest release 8.x-1.8 is generated with the code that is included in this PR.

scripts/travis-ci/create-release-tarball.sh
ademarco: Yeah we should remove the vendor folder, we will have Composer taking care of that for us.

PR #52, opened 2018-03-21
OPENEUROPA-531: form elements should be tested for rendering inside a form

brummbar:


aritomelo:
PR can be merged!

Happy to see type hints!!!

PR #59, opened 2018-04-12
OPENEUROPA-355: Theming confirm password.

aritomelo:
It was needed to attach 'ecl-help-block' however the element is rendered by javascript so the approach took was via styling.


sass/components/_form.scss
```diff
@@ -0,0 +1,22 @@
...
+
+    .is-good,
+    .is-strong {
+      background-color: $ecl-color-success;
+    }
+  }
+
+  /* @see @ec-europa/ecl-forms-help-blocks/_help-blocks.scss */
+  .password-strength .password-strength__title,
+  .password-confirm {
```
richardcanoe: ecl-help-block should also be added to extend password-suggestions.

oe_theme.theme
```diff
@@ -155,6 +155,11 @@ function _oe_theme_preprocess_input_label_wrapper(array &$variables) {
  * label is already rendered in the input template.
  */
 function oe_theme_preprocess_form_element(&$variables) {
+  $variables['is_password_confirm'] = FALSE;
+  if (in_array('password_confirm', explode('[', $variables['element']['#name']))) {
+    $variables['is_password_confirm'] = TRUE;
+  }
+
```
ademarco: Is this necessary at all?

aritomelo: No, it isn't. I will push a fix asap.

ademarco: Ok thanks, I'll reopen the ticket then.

PR #61, opened 2018-04-13
OPENEUROPA-382: Only required ECL templates are copied.

richardcanoe:
Only templates included in oe_theme twig templates are included.

Check using `find . -type f -exec grep -H '@ecl/ecl-' {} \;` in the templates folder


ecl-builder.config.js
```diff
@@ -34,7 +34,18 @@ module.exports = {
     { from: path.resolve(source_folder, 'ecl-forms-selects/images'), to: path.resolve(__dirname, 'images')},
     { from: path.resolve(source_folder, 'ecl-logos/images'), to: path.resolve(__dirname, 'images')},
     { from: path.resolve(source_folder, 'ecl-messages/images'), to: path.resolve(__dirname, 'images')},
-    { from: path.resolve(source_folder), patterns: '**/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-accordions/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-breadcrumbs/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-buttons/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-fields/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-forms*/*.twig', to: path.resolve(__dirname, 'templates/components') },
```
ademarco: This might be a little too broad, with that we get things like the `<form>` tag and other templates we don't really use.

I'd just add those that we explicitly require, plus those required in turn, e.g. some form components might embed the label component via `{% embed '@ec-europa/ecl-forms-labels'...`

ecl-builder.config.js
```diff
@@ -34,7 +34,18 @@ module.exports = {
...
+    { from: path.resolve(source_folder), patterns: 'ecl-link-blocks/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-links/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-logos/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-messages*/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-navigation*/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-page*/*.twig', to: path.resolve(__dirname, 'templates/components') },
```
ademarco: I'd just include `ecl-pager` here.

ecl-builder.config.js
```diff
@@ -34,7 +34,18 @@ module.exports = {
...
+    { from: path.resolve(source_folder), patterns: 'ecl-breadcrumbs/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-buttons/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-fields/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-forms*/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-link-blocks/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-links/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-logos/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-messages*/*.twig', to: path.resolve(__dirname, 'templates/components') },
+    { from: path.resolve(source_folder), patterns: 'ecl-navigation*/*.twig', to: path.resolve(__dirname, 'templates/components') },
```
ademarco: We can restrict this too as it would include `ecl-navigation-inpages` which we don't need.

PR #66, opened 2018-04-23
OPENEUROPA-543: Theme tables.

netlooker:
I have prepared a small patch with the helper controller https://gist.github.com/netlooker/c76fc96b6e829cd3501bc49bca87c455 which can be used for reviewing purposes.
I have also created the custom class for the Kernel testing as I am almost sure that we will extend tests in future. For the moment I focused to cover current acceptance criteria.


templates/tables/table.html.twig
```diff
@@ -0,0 +1,104 @@
+{#
+/**
+ * @file
+ * Theme override to display a table.
```
ademarco: What we do is removing the whole comment and just add a reference to the original template, like [this templates](https://github.com/openeuropa/oe_theme/blob/master/templates/content/node.html.twig) for an example.

templates/tables/table.html.twig
```diff
@@ -0,0 +1,104 @@
...
+ * - empty: The message to display in an extra row if table does not have
+ *   any rows.
+ * - no_striping: A boolean indicating that the row should receive no striping.
+ * - header_columns: The number of columns in the header.
+ *
+ * @see template_preprocess_table()
+ */
+#}
+{{ attach_library('oe_theme/tables') }}
+<table{{ attributes.addClass('ecl-table') }}>
```
ademarco: We could actually extend the original template and just add the class to the attributes, something like:

```
{% extends "@stable/content/node.html.twig" %}
{% set display_submitted = false %}
```

PR #68, opened 2018-04-30
OPENEUROPA-298: Commit hash updated for oe_profile.

richardcanoe:
Includes oe_profile profile for future site building requirements.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
         "symfony/dom-crawler": "^3",
         "webflo/drupal-core-require-dev": "^8.5",
         "drupal/styleguide": "~1.0-alpha3",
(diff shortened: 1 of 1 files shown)


modules/oe_theme_demo/config/optional/block.block.oe_theme_demo_header_language_switcher.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
```
ademarco: @richardcanoe the demo language switcher block needs to be removed as it is now provided by OE Multilingual. See #70

brummbar:
Good for me!

PR #69, opened 2018-05-03
OPENEUROPA-560: Port CI/CD setup to Drone

ademarco:


.blackfire.yml
```diff
@@ -0,0 +1,23 @@
+tests:
```
ademarco: We can remove this file for now as we don't run blackfire tests.

composer.json
```diff
@@ -6,24 +6,27 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
+        "drupal/core": "8.6.x-dev",
```
ademarco: It's ok to have `"drupal/core": "~8.6",` since we have `minimum-stability: dev` composer will download 8.6.x, if no stable release is found.

scripts/drone/.htaccess
```diff
@@ -0,0 +1,164 @@
+#
```
ademarco: We can remove this file.

scripts/drone/create-release-tarball.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh -e
```
ademarco: We can remove this file as now release archive is generated by the task runner.

composer.json
pfrenssen: :+1:  I was curious about how Composer would handle this situation, that is good to know!

scripts/drone/create-release-tarball.sh
pfrenssen: Good point, I hope one of the colleagues will be able to take care of this since I am out of time and will be on holiday for the remainder of the sprint.

Medi4tor:
Reviewing: Everything is green in drone CI. Nothing to add.

PR #72, opened 2018-05-18
OPENEUROPA-580: Use original language and add pt and mt in language switcher

Medi4tor:
# OPENEUROPA-580

## Description
Use original language in language switcher and add 'pt' and 'mt'

### DIFF
--- a/tests/features/theme-showcase.feature
+++ b/tests/features/theme-showcase.feature
@@ -47,11 +47,28 @@ Feature: Theme showcase
     Then the "language switcher overlay" is not visible
 
   @javascript
-  Scenario: Site visitor can change language using the language switcher
+  Scenario: Site visitor can see and change language using the language switcher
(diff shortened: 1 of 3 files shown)


tests/features/theme-showcase.feature
```diff
@@ -47,11 +47,28 @@ Feature: Theme showcase
...
     When I open the language switcher dialog
-    And I click "Polish"
+    Then I should see the link "Български"
+    And I should not see the link "Bulgarian"
+    Then I should see the link "Português"
+    And I should not see the link "Portuguese"
+    Then I should see the link "Malti"
+    And I should not see the link "Maltese"
+
+      # Change languages.
```
ademarco: We don't need to comment Behat scenarios, if steps are unclear we can create custom ones that are easier to understand for business.

PR #73, opened 2018-05-22
OPENEUROPA-581: Update ECL library to 1.0.0.

voidtek:
## Description

Update ECL library to 1.0.0.


ademarco:
When building I get the error below, which refers to JS compilation and, indeed, I don't get the compiled `./js/base.js` after that:

```
...
0.0.1 build:scripts /home/node/app
[build:scripts] > ecl-builder scripts
[build:scripts] 
[build:scripts] (node:222) UnhandledPromiseRejectionWarning: BrowserslistError: Unknown version 66 of android
[build:scripts]     at error (/home/node/app/node_modules/@ecl/builder/node_modules/rollup/dist/rollup.js:224:15)
[build:scripts]     at Object.error (/home/node/app/node_modules/@ecl/builder/node_modules/rollup/dist/rollup.js:17174:21)
[build:scripts]     at /home/node/app/node_modules/@ecl/builder/node_modules/rollup/dist/rollup.js:17183:29
[build:scripts]     at <anonymous>
[build:scripts]     at process._tickCallback (internal/process/next_tick.js:160:7)
[build:scripts]     at Function.Module.
```

ademarco:
Had a quick chat with @yhuard and he suggests you have a look at https://github.com/ec-europa/europa-component-library/blob/master/src/systems/ec/ec-preset/ec-preset-website/ecl-builder.config.js for reference, we should port ours to that one.

ecl-builder.config.js
```diff
@@ -22,32 +22,22 @@ module.exports = {
...
-    { from: path.resolve(source_folder), patterns: 'ecl-logos/*.twig', to: path.resolve(__dirname, 'templates/components') },
-    { from: path.resolve(source_folder), patterns: 'ecl-messages/*.twig', to: path.resolve(__dirname, 'templates/components') },
-    { from: path.resolve(source_folder), patterns: 'ecl-navigation-lists/*.twig', to: path.resolve(__dirname, 'templates/components') },
-    { from: path.resolve(source_folder), patterns: 'ecl-navigation-menus/*.twig', to: path.resolve(__dirname, 'templates/components') },
-    { from: path.resolve(source_folder), patterns: 'ecl-pagers/*.twig', to: path.resolve(__dirname, 'templates/components') },
-    { from: path.resolve(source_folder), patterns: 'ecl-skip-links/*.twig', to: path.resolve(__dirname, 'templates/components') },
-    { from: path.resolve(source_folder), patterns: 'ecl-lang-select-sites/*.twig', to: path.resolve(__dirname, 'templates/components') },
-    { from: path.resolve(source_folder), patterns: 'ecl-language-list/*.twig', to: path.resolve(__dirname, 'templates/components') },
-    { from: path.resolve(source_folder), patterns: 'ecl-dialogs/*.twig', to: path.resolve(__dirname, 'templates/components') },
```
ademarco: Not all of the deleted components seems to be there in the new version, for example I cannot find:

```
'ecl-accordions/*.twig'
'ecl-breadcrumbs/*.twig'
'ecl-buttons/*.twig'
'ecl-fields/*.twig'
```

voidtek: These components change the name: 
'ec-component-accordion/*.twig'
'ec-component-breadcrumb/*.twig'
'ec-component-buttons/*.twig'
'ec-component-fields/*.twig'

imanoleguskiza: If I see the old implementation we were only taking the twig templates of the components we have implemented. In the new implementation we are taking all of the twig templates. Since this is not mentioned in the ticket, I think we should keep it the same right?

modules/oe_theme_demo/templates/demo-block-header-search.html.twig
```diff
@@ -1,7 +1,7 @@
...
-  </label>
-  <button class="ecl-button ecl-button--form ecl-search-form__button" type="submit">Search</button>
-</form>
+{% include '@ecl/generic-component-search-form' with {
+  aria_label: 'Search this website'|t,
+  button: { 'label': 'Search' }|merge(button|default({})),
+  input: { 'id': 'global-search' }|merge(input|default({})),
+  system: 'ec',
+  extra_classes: 'ecl-site-header__search demo-block-header-search'
+} only %}
```
imanoleguskiza: Missing empty line

modules/oe_theme_demo/templates/demo-block-header-site-switcher.html.twig
```diff
@@ -1,6 +1,9 @@
...
-</div>
+{% include '@ecl/ec-component-site-switcher' with {
+  political: { href: '#', label: 'Commission and its priorities' },
+  info: {
+    href: '#',
+    label: 'Policies, information and services',
+    is_active: true,
+  },
+  extra_classes: 'ecl-site-switcher--header demo-block-site-switcher'
+} %}
```
imanoleguskiza: Missing empty line

ecl-builder.config.js
voidtek: My idea is to optimize and implement a very generic approach. For example, for images, it is not possible with the actual runner to put a generic step.
In addition, if we have to put all the necessary components and dependencies, we will add more than 20 lines in the runner, I think it's useless and not optimized.

imanoleguskiza: I honestly don't mind it much, I was more worried with not being consistent with the previous implementation. I'm happy with it as it is

ecl-builder.config.js
```diff
@@ -1,53 +1,58 @@
...
+    { from: path.resolve(nodeModules, '@ecl/generic-component-social-icon/images'), to: path.resolve(outputFolder, 'images')},
+    { from: path.resolve(nodeModules, '@ecl/ec-component-form-checkbox/images'), to: path.resolve(outputFolder, 'images')},
+    { from: path.resolve(nodeModules, '@ecl/ec-component-form-feedback-message/images'), to: path.resolve(outputFolder, 'images')},
+    { from: path.resolve(nodeModules, '@ecl/ec-component-form-radio/images'), to: path.resolve(outputFolder, 'images')},
+    { from: path.resolve(nodeModules, '@ecl/ec-component-form-select/images'), to: path.resolve(outputFolder, 'images')},
+    { from: path.resolve(nodeModules, '@ecl/ec-component-logo/images'), to: path.resolve(outputFolder, 'images')},
+    { from: path.resolve(nodeModules, '@ecl/ec-component-message/images'), to: path.resolve(outputFolder, 'images')},
+    { from: path.resolve(nodeModules, '@ecl/ec-component-social-icon/images'), to: path.resolve(outputFolder, 'images')},
+    { from: path.resolve(nodeModules, '@ecl'), patterns: '*/*.twig', to: path.resolve(outputFolder, 'templates/components') },
```
ademarco: This will copy all templates under our `templates/components` while we should only copy over what we need. It might be enough to just remove this line:
```
+    { from: path.resolve(nodeModules, '@ecl'), patterns: '*/*.twig', to: path.resolve(outputFolder, 'templates/components') },
```

ecl-builder.config.js
```diff
@@ -1,53 +1,58 @@
...
+    { from: path.resolve(nodeModules, '@ecl'), patterns: '**/variants.json', to: path.resolve(outputFolder, 'templates/components') }
```
ademarco: Is this useful for us? Where do we use `variants.json`?

ademarco:
@voidtek  It seems that JS stops working after enabling `oe_paragraphs_demo` module.

ademarco:
@voidtek actually it always works as anonymous, even with `oe_paragraphs_demo` enabled and content created but it stops as soon as we create a content item AND ONLY when logged in.

PR #78, opened 2018-06-08
OPENEUROPA-615: Add templates for splash page and css for body in sass.

Medi4tor:
### Description

Add templates for splash page and  background for body in sass.

### Change log

- Added:
templates/layout/page--select-language.html.twig
sass/components/_splash-page.scss
- Changed:
oe_theme.theme
sass/style.scss

### Commands

 npm run build
 drush cr


sass/components/_splash-page.scss
```diff
@@ -0,0 +1,3 @@
+.page-splash {
+  background-color: #004494;
```
ademarco: We should use ECL SASS variables for colors.

templates/html/html--select-language.html.twig
```diff
@@ -0,0 +1,32 @@
...
+#}
+<!DOCTYPE html>
+<html{{ html_attributes }}>
+  <head>
+    <head-placeholder token="{{ placeholder_token }}">
+    <title>{{ head_title|safe_join(' | ') }}</title>
+    <css-placeholder token="{{ placeholder_token }}">
+    <js-placeholder token="{{ placeholder_token }}">
+  </head>
+  <body{{ attributes.addClass('language-' ~ html_attributes.lang).addClass('ecl-typography').addClass('page-splash') }}>
```
ademarco: We should rather use the `page-splash` via a preprocess so we don't need to override the html template.

PR #96, opened 2018-06-24
OPENEUROPA-694: Make sure that in page navigation link translations are correctly displayed

ademarco:


oe_theme.theme
```diff
@@ -533,6 +533,11 @@ function oe_theme_preprocess_paragraph__oe_content_row__variant_inpage(&$variabl
     $field_render[$delta]['#theme_wrappers']['container'] = [
       '#attributes' => ['id' => $unique_id],
     ];
+
```
sandervd: oe_theme_preprocess_paragraph__oe_content_row__variant_inpage
-> This function should move to oe_paragraphs.
The theme should not be aware of the fieldnames, and the logic doesn't belong here.
Coupling can be avoided by changing 
$unique_id = Html::getUniqueId('ecl-inpage-' . $sub_paragraph->id());
to $unique_id = Html::getUniqueId('inpage-' . $sub_paragraph->id());

sandervd:
The functionality works, but is implemented in the theme.
I would propose to move it to the module declaring the paragraph type.

oe_theme.theme
ademarco: We can move part of it in oe_paragraphs indeed, we have other refactoring of this sort to do in the theme, would it be ok to open a followup and unblock the PR?

sandervd: Created OPENEUROPA-697 for follow-up, to unblock delivery.

PR #101, opened 2018-07-04
OPENEUROPA-582: Update images on docker-compose and drone file.

voidtek:


pfrenssen:
This build is failing.

composer.json
```diff
@@ -66,6 +67,7 @@
     },
     "config": {
         "sort-packages": true,
+        "preferred-install": "source",
```
ademarco: @voidtek any reason why we need to specify `"preferred-install": "source"`?

ademarco: @voidtek any reason why we need to specify `"preferred-install": "source"`?

composer.json
```diff
@@ -54,7 +54,8 @@
         "composer-exit-on-patch-failure": true,
         "patches": {
             "drupal/core": {
-                "https://www.drupal.org/project/drupal/issues/2943172": "https://www.drupal.org/files/issues/2018-07-05/2943172-kernel-test-base-3.patch"
+                "https://www.drupal.org/project/drupal/issues/2943172": "https://www.drupal.org/files/issues/2018-07-05/2943172-kernel-test-base-3.patch",
+                "https://www.drupal.org/project/drupal/issues/2599228": "https://www.drupal.org/files/issues/2018-05-17/2599228-51.patch"
```
ademarco: We can remove this patch since it get it in from the multilingual.

voidtek: I removed the patch.

PR #103, opened 2018-07-05
OPENEUROPA-384: Add extra attributes to checkbox and radios

imanoleguskiza:


drupol:
Hello @imanoleguskiza,

The branch was out of date, I rebased it.

oe_theme.theme
```diff
@@ -183,6 +183,25 @@ function _oe_theme_preprocess_input_label_wrapper(array &$variables): void {
...
+  ];
+  $extra_attributes = array_diff_key($extra_attributes, $predefined_attributes);
+  foreach ($extra_attributes as $key => $value) {
+    if (is_string($value)) {
+      $variables['input_attributes'][] = [
+        'name' => $key,
+        'value' => $value,
+      ];
+    }
+  }
```
ademarco: We should cover this with a Kernel test, it should not be that difficult as we can use `./tests/fixtures/rendering.yml`, we already have tests there for checkboxes, like the one below:

```yaml
- array:
    '#type': checkbox
    '#title': Checkbox
    '#set_validation_error': true
  assert:
    contains:
      - 'Checkbox'
    count:
      'label.ecl-form-label.ecl-checkbox.ecl-checkbox--has-error': 1
```

PR #107, opened 2018-07-09
OPENEUROPA-714: Update ECL version to 1.2.0

drupol:


tests/fixtures/rendering.yml
```diff
@@ -318,7 +318,7 @@
       '.ecl-page-header__title h1': "Business, Economy, Euro"
       '.ecl-page-header__intro': "EU economy, finance and the euro, and practical information for EU businesses and entrepreneurs on product safety, environmental regulations, trade with non-EU countries and competition rules."
       '.ecl-breadcrumb__segment--first a[href="http://www.example.com/"]': "European Commission"
-      '.ecl-breadcrumb__segment--last a[href="http://www.example.com/announcements/"]': "Announcements"
+      '.ecl-breadcrumb__segment--last span': "Announcements"
```
ademarco: ECL 1.2.0 specifications expect the last breadcrumb link to be the current page title. This will not impact this test specifically bu we will need to make sure that the page title is displayed in the breadcrumb.

We can add the page title without its link in the breadcrumb preprocess and add a test for that.

imanoleguskiza: @voidtek and I added a separate test to check wether the page title is added as a span to the breadcrumb or not: https://github.com/openeuropa/oe_theme/blob/OPENEUROPA-714/tests/Kernel/BreadcrumbTest.php#L95

PR #108, opened 2018-07-12
OPENEUROPA-722: Make sure the default body classes are output

pfrenssen:
The theme misses the important body classes that are used to identify whether the user is logged in, visiting an admin page, the node type etc. Without these it will not be possible to write effective JS, and the functional tests are also failing.

Ref. `template_preprocess_html()`
Ref. `html.html.twig` in the Classy theme.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "drupal-composer/drupal-scaffold": "^2.2",
         "drupal/config_devel": "~1.2",
         "drupal/console": "^1",
-        "drupal/drupal-extension": "^3.4",
+        "drupal/drupal-extension": "^4.0",
         "drush/drush": "^9",
         "nikic/php-parser": "~3",
         "openeuropa/code-review": "~0.3",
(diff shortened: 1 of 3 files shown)


templates/html/html.html.twig
```diff
@@ -14,7 +22,7 @@
     <css-placeholder token="{{ placeholder_token }}">
     <js-placeholder token="{{ placeholder_token }}">
   </head>
-  <body{{ attributes.addClass('language-' ~ html_attributes.lang).addClass('ecl-typography') }}>
+  <body{{ attributes.addClass('language-' ~ html_attributes.lang).addClass('ecl-typography').addClass(body_classes) }}>
```
brummbar: We could merge these two classes added here to the newly created `body_classes` array.

pfrenssen:
Thanks @brummbar, great suggestion!

I've added the test and rebased the PR so that the test runs first. It nicely demonstrates that the body classes are not present, since in order to detect if a user is logged in, it verifies the body classes first, and then resorts to finding a logout link. It can't do that:

```
--- Failed steps:
001 Scenario:                                                      # tests/features/authentication.feature:7
      Given I am logged in as a user with the "authenticated" role # tests/features/authentication.feature:10
        Unable to determine if logged in because 'log_out' link cannot be found for user 'fBiZSXqm' with role 'authenticated' (Exception)
```

composer.json
```diff
@@ -57,6 +57,14 @@
                 "https://www.drupal.org/project/drupal/issues/2943172": "https://www.drupal.org/files/issues/2018-07-05/2943172-kernel-test-base-3.patch"
             }
         },
+        "patches-ignore": {
```
Medi4tor: Hi,
Why do we have this ignore section in this commit? 
Is this related to the body class?

tests/features/authentication.feature
```diff
@@ -0,0 +1,10 @@
+@api
+Feature: Authentication
+  In order to allow personalized actions to be taken
```
Medi4tor: I guess that was just a missing test... nothing related with body classes? You confirm?

composer.json
imanoleguskiza: The issue with the patches across components will be solved in a follow up ticket, so for now I'm ok with merging this PR.

PR #110, opened 2018-07-17
OPENEUROPA-721: Update code-review and grump.yml.dist file.

imanoleguskiza:


composer.json
```diff
@@ -21,7 +21,7 @@
         "drupal/drupal-extension": "^3.4",
         "drush/drush": "^9",
         "nikic/php-parser": "~3",
-        "openeuropa/code-review": "^0.2",
+        "openeuropa/code-review": "^0.3",
```
pfrenssen: For dependencies that are on version 0 it is better to use the tilde so that future versions are automatically considered when they are released. So a better version range would be `~0.3`.

PR #117, opened 2018-07-31
OPENEUROPA-772: Move styleguide into require-dev.

richardcanoe:
Test using europa-demo branch OPENEUROPA-772


.gitignore
```diff
@@ -18,9 +18,8 @@ phpunit.xml
 /js/base.js
 /templates/components/
 /oe_theme/
-*.tar.gz
```
ademarco: The "create release" command generates tar balls, hence the line it, we should not remove it as it is useful when testing releases.

composer.json
```diff
@@ -19,6 +18,7 @@
         "drupal/config_devel": "~1.2",
         "drupal/console": "^1",
         "drupal/drupal-extension": "^4.0",
+        "drupal/styleguide": "~1.0-alpha3",
```
ademarco: We should keep this alphabetically ordered.

voidtek: The alphabetically ordered is respected.

ademarco: facepalm.

README.md
```diff
@@ -7,12 +7,19 @@ Drupal 8 theme based on the [Europa Component Library][1] (ECL).
...
 
 - [Installation](#installation)
+- [Requirements](#requirements)
 - [Development](#development)
   - [Project setup](#project-setup)
   - [Using Docker Compose](#using-docker-compose)
   - [Disable Drupal 8 caching](#disable-drupal-8-caching)
   - [Working with ECL components](#working-with-ecl-components)
-- [Demo module](#demo-module)
+- [Demo module](https://github.com/openeuropa/oe_theme/blob/master/modules/oe_theme_demo/README.md)
```
ademarco: Shouldn't we link the local file here? Like:

```
[Demo module](./modules/oe_theme_demo/README.md)
```

Then it will always point to the right file regardless on the branch.

PR #118, opened 2018-08-08
OPENEUROPA-772: Update compose.json to have openeuropa/drupal-core-require-dev.

voidtek:


composer.json
```diff
@@ -6,7 +6,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "drupal/core": "~8.6@alpha",
+        "drupal/core": "8.6.0-alpha1",
```
ademarco: This is great as a first test, obviously we need to put it back to  `~8.6@alpha` as soon as `openeuropa/drupal-core-require-dev` has a 8.6.0-beta2 release.

PR #124, opened 2018-09-03
OPENEUROPA-780: Make sure component dependencies are as relaxed as possible.

aritomelo:
Dependencies:
     
     - Update some version to use tilde, so it uses the minimal minor version and also for a consistency use decimal number as the version.

     - Update some packages to use minimal minor version.

     - Added composer boundary to test lowest and highest.

     - Added packagist badge.


composer.json
```diff
@@ -8,26 +8,26 @@
     "require": {
         "drupal/core": "~8.6.0@alpha",
```
pfrenssen: This version restriction seems very limiting since it will not allow to go to 8.7.0. Since we are only days away from the final release of Drupal 8.6, can we maybe change this to `~8.6`? This should normally work since our minimum stability is set to `dev`.

composer.json
```diff
@@ -6,28 +6,28 @@
...
+        "drupal/core": "~8.6",
         "drupal/ui_patterns": "~1.0-rc1",
-        "openeuropa/ecl-twig-loader": "^0.1",
+        "openeuropa/ecl-twig-loader": "~0.1",
         "php": "^7.1"
     },
     "require-dev": {
-        "composer/installers": "^1.2",
+        "composer/installers": "~1.0",
         "cweagans/composer-patches": "~1.0",
```
voidtek: We should change 
`"cweagans/composer-patches": "^1.6"`

pfrenssen:
I had a green build after an afternoon of experimentation. I have now rebased the branch to remove the experimental commits, keeping only the changes that were actually necessary to get a working build.

PR #133, opened 2018-10-09
 OPENEUROPA-1255: Create UI pattern for dropdown component

nagyad:


oe_theme.theme
```diff
@@ -413,6 +413,30 @@ function oe_theme_preprocess_pattern_accordion(&$variables) {
   }
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
```
upchuk: Comment here is wrong: `template_preprocess_pattern_dropdown()`

js/dropdown.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @file
+ * ECL dropdown behavior.
+ */
+(function (ECL, Drupal) {
+    Drupal.behaviors.eclDropdown = {
+        attach: function attach() {
+            ECL.initExpandables('button.ecl-button');
```
upchuk: From @ademarco: This targets all buttons, it shouldn't.

oe_theme.theme
nagyad: Can you take a look on the others as well?

PR #137, opened 2018-10-12
OPENEUROPA-1253: Consolidate the existing pattern APIs.

upchuk:


oe_theme.theme
```diff
@@ -475,6 +490,38 @@ function oe_theme_preprocess_pattern_social_media_links_horizontal(&$variables)
...
+  ];
+
+  $skip = (array) $variables['skip_date_parts'];
+  foreach ($date_parts as $part => $value) {
+    if (in_array($part, $skip)) {
+      unset($date_parts[$part]);
+    }
+  }
+
+  $variables += $date_parts;
```
drupol: I know we discussed yesterday about that, but do you think such construction has its place in there?
Because the code block from 516 to 520 is an array filter... so let's use the proper PHP function for this, what do you think?

```php
$variables += \array_filter(
  $date_parts, 
  function ($part) use ($skip) {
    return !\in_array($part, $skip);
  },
  ARRAY_FILTER_USE_KEY
);

```

ademarco: Those are actually variants, we can now deal with that with theme suggestions.

PR #139, opened 2018-10-17
OPENEUROPA-1251: Add button pattern.

imanoleguskiza:


ademarco:
@imanoleguskiza lower dependency build is failing because we forgot to increase UI Pattern version, it should now be set to `~8.x-1.0-rc2` as variants are not available in earlier versions.

nagyad:
Please, rebase to OPENEUROPA-1254 since the OPENEUROPA-1251 is the epic and not the button implementation ticket.

PR #140, opened 2018-10-18
Openeuropa 1257: Implement "Featured Item" pattern with variants.

sergepavle:


composer.json
```diff
@@ -23,7 +23,7 @@
         "nikic/php-parser": "~3.0",
         "openeuropa/code-review": "^1.0.0-alpha4",
         "openeuropa/behat-transformation-context" : "~0.1",
-        "openeuropa/drupal-core-require-dev": "^8.6",
+        "openeuropa/drupal-core-require-dev": "8.6.x-dev",
```
ademarco: Shouldn't this follow what we have in require? @voidtek what do you think?

templates/patterns/featured-item/featured_item.ui_patterns.yml
```diff
@@ -0,0 +1,50 @@
+featured_item:
+  label: "Featured item"
+  description: "The featured item component uses a different visualisation to attract user attention. It's always the first item on the \"Latest section\"."
+  variants:
+    '':
```
nagyad: Is this the correct way, to leave empty for default? I'm asking because I don't know.

sergepavle: I don't see other more correct solution. Yml is still valid. I did this strictly by documentation.
But probably `~` should work as well and actually have more nicer view.
Next reviewer I hope will help with this dilemma.

.drone.yml
```diff
@@ -4,7 +4,8 @@ workspace:
 
 services:
   web:
-    image: fpfis/httpd-php-dev:7.1
+    # @todo: move back to 7.1 after Docker image is fixed upstream.
```
imanoleguskiza: What is the reason for the change?

templates/patterns/featured-item/featured_item.ui_patterns.yml
imanoleguskiza: I also had a similar issue on my own PR and I don't have a clear answer either. I would try with ~ just because it seems cleaner but I'll leave it to @sergepavle

.drone.yml
```diff
@@ -4,7 +4,8 @@ workspace:
 
 services:
   web:
-    image: fpfis/httpd-php-dev:7.1
+    # @todo: move back to 7.1 after Docker image is fixed upstream. Issue related to missed apache mod_rewrite module.
```
dxvargas: @sergepavle please add a reference to some ticket/issue where we can track when it's solved.

templates/patterns/featured_item/featured_item.ui_patterns.yml
```diff
@@ -0,0 +1,50 @@
+featured_item:
+  label: "Featured item"
+  description: "The featured item component uses a different visualisation to attract user attention. It's always the first item on the \"Latest section\"."
+  variants:
+    ~:
```
ademarco: Better to avoid empty variant names, I should actually make this not to pass validation in UI Patterns as it would break theme suggestions badly. Let's go for `default`.

tests/fixtures/rendering.yml
```diff
@@ -684,3 +684,60 @@
       'a.ecl-list-item__link[href="http://example.com"]': 1
       'h3.ecl-list-item__title.ecl-heading.ecl-heading--h3': 0
       'div.ecl-list-item__detail.ecl-paragraph': 0
+- array:
+    '#type': pattern
+    '#id': featured_item
+    '#variant': ""
```
ademarco: Same here, let's use `default`.

composer.json
```diff
@@ -23,7 +23,7 @@
         "nikic/php-parser": "~3.0",
         "openeuropa/code-review": "^1.0.0-alpha4",
         "openeuropa/behat-transformation-context" : "~0.1",
-        "openeuropa/drupal-core-require-dev": "^8.6",
+        "openeuropa/drupal-core-require-dev": "~8.6@dev",
```
ademarco: Shouldn't this follow what's specified in `require`? `"drupal/core": "^8.6",`

templates/patterns/featured_item/featured_item.ui_patterns.yml
```diff
@@ -0,0 +1,50 @@
...
+      label: "Extra classes"
+      description: "Specific extra classes."
+      preview: ""
+    extra_attributes:
+      type: "array"
+      label: "Extra attributes"
+      description: "Specific extra attributes."
+      preview:
+        - name: "data-extra-attr"
+          value: "extra-attr-value"
```
ademarco: I'd not allow site builders to specify extra classes, even if the component does allow that. Here we are defining our own API that can be more restrictive and it doesn't have to necessarily follow what's in the ECL. Shall we remove `extra_classes` and `extra_attributes`?

templates/patterns/featured_item/pattern-featured-item.html.twig
```diff
@@ -0,0 +1,15 @@
...
+ */
+#}
+{% include '@ecl/ec-component-featured-item' with {
+  'variant': variant,
+  'image': image,
+  'metas': metas,
+  'title': title,
+  'link': link,
+  'extra_classes': extra_classes,
+  'extra_attributes': extra_attributes,
```
ademarco: See comment above, remove `extra_classes` and `extra_attributes`.

sergepavle: Removed.

templates/patterns/featured_item/featured_item.ui_patterns.yml
sergepavle: I removed this from preview

composer.json
sergepavle: Corrected!

templates/patterns/featured_item/featured_item.ui_patterns.yml
sergepavle: Done.

.drone.yml
drupol: This has been added and reverted later, so, discard this please.

.drone.yml
drupol: This has been added and reverted later, so, discard this please.

tests/fixtures/rendering.yml
drupol: Done.

PR #141, opened 2018-10-19
OPENEUROPA-1236: Add breadcrumb items to page header block.

imanoleguskiza:


nagyad:
Tests? :)

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -99,4 +124,33 @@ public function setTitle($title): self {
...
+   * Returns a list of breadcrumb items ready to be rendered.
+   *
+   * @param string $title
+   *   The title of the current page.
+   *
+   * @return array
+   *   A list of breadcrumb items.
+   */
+  protected function getBreadcrumbSegments($title) {
+
```
upchuk: Remove empty line please

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -28,13 +30,27 @@ class PageHeaderBlock extends BlockBase implements ContainerFactoryPluginInterfa
 
   use StringTranslationTrait;
 
+  /**
+   * The breadcrumb manager.
+   *
+   * @var \Drupal\Core\Breadcrumb\BreadcrumbBuilderInterface
+   */
+  protected $breadcrumbManager;
```
upchuk: Let's call this `$breadcrumbBuilder`

tests/Behat/MinkContext.php
```diff
@@ -100,4 +101,50 @@ protected function getLanguageSwitcherOverlay(): NodeElement {
...
+   *
+   * @param \Behat\Gherkin\Node\TableNode $table
+   *   The pages data.
+   *
+   * @throws \Behat\Mink\Exception\ElementNotFoundException
+   * @throws \Behat\Mink\Exception\ExpectationException
+   *
+   * @Then the breadcrumb should contain the following item(s):
+   */
+  public function checkBreadcrumbItems(TableNode $table): void {
```
upchuk: Let's move all the public methods above the protected ones.

tests/features/page-header.feature
```diff
@@ -13,13 +13,25 @@ Feature: Page header block component.
     When I am an anonymous user
     And I go to the "Robots are everywhere" demo page
     Then I should see the heading "Robots are everywhere" in the "page header"
+    And the breadcrumb should contain the following items:
+      | element | text                  |
+      | a       | Home                  |
+      | span    | Robots are everywhere |
```
upchuk: Can we make these elements more business language oriented? Something like "link" and "active" or something like this?

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -77,14 +101,15 @@ public static function create(ContainerInterface $container, array $configuratio
    */
   public function build(): array {
     $metadata = $this->getContext('page_header')->getContextData()->getValue();
-
+    $title = $metadata['title'] ?? $this->title;
```
upchuk: Why did you move this to a new line?

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -99,4 +124,33 @@ public function setTitle($title): self {
...
+  /**
+   * Returns a list of breadcrumb items ready to be rendered.
+   *
+   * @param string $title
+   *   The title of the current page.
+   *
+   * @return array
+   *   A list of breadcrumb items.
+   */
+  protected function getBreadcrumbSegments($title) {
```
upchuk: You are missing the return value type hint. I see why $title cannot be type hinted.

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -99,4 +124,33 @@ public function setTitle($title): self {
...
+    $breadcrumb_segments = [];
+    /** @var \Drupal\Core\Link $link */
+    foreach ($breadcrumb->getLinks() as $link) {
+      $breadcrumb_segments[] = [
+        'href' => $link->getUrl(),
+        'label' => $link->getText(),
+      ];
+    }
+    // Add the title to the segments only if it's not empty.
+    if (!empty($title)) {
```
upchuk: Do we need to check if empty? Won't there always be a title here?

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
upchuk: @imanoleguskiza ?

imanoleguskiza: I need the $title in two places: the build array itself and to generate the breadcrumb array

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -51,13 +67,19 @@ class PageHeaderBlock extends BlockBase implements ContainerFactoryPluginInterfa
...
    *   The plugin implementation definition.
+   * @param \Drupal\Core\Breadcrumb\BreadcrumbBuilderInterface $breadcrumb_builder
+   *   The breadcrumb builder service.
    * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
    *   The factory for configuration objects.
+   * @param \Drupal\Core\Routing\RouteMatchInterface $current_route_match
+   *   The current route match.
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, ConfigFactoryInterface $config_factory) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, BreadcrumbBuilderInterface $breadcrumb_builder, ConfigFactoryInterface $config_factory, RouteMatchInterface $current_route_match) {
```
brummbar: New services should be added at the end.

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -99,4 +124,32 @@ public function setTitle($title): self {
...
+   * Returns a list of breadcrumb items ready to be rendered.
+   *
+   * @param string $title
+   *   The title of the current page.
+   *
+   * @return array
+   *   A list of breadcrumb items.
+   */
+  protected function getBreadcrumbSegments($title): array {
+    $breadcrumb = $this->breadcrumbBuilder->build($this->currentRouteMatch);
```
brummbar: The `\Drupal\Core\Breadcrumb\Breadcrumb` class has cacheability metadata. We need to bubble up that data. See `::toRenderable()` in said class.

tests/Behat/MinkContext.php
```diff
@@ -16,6 +17,23 @@ class MinkContext extends DrupalExtensionMinkContext {
...
+   */
+  private $breadcrumbElements = [];
+
+  /**
+   * TransformationContext constructor.
+   *
+   * @param array $breadcrumb_elements
+   *   Breadcrumb elements mapping.
+   */
+  public function __construct(array $breadcrumb_elements = []) {
```
brummbar: There's no need to have these data coming from the context. Let's remove all of that and have two steps:
`@Then the breadcrumb (links|trail) should be :links`
`@Then the breadcrumb active element should be :text`
Below I'll make an example for the scenario.

tests/Behat/MinkContext.php
```diff
@@ -85,6 +103,38 @@ public function assertActiveLanguageSwitcherLink($label): void {
...
+   *
+   * @throws \Behat\Mink\Exception\ElementNotFoundException
+   * @throws \Behat\Mink\Exception\ExpectationException
+   *
+   * @Then the breadcrumb should contain the following item(s):
+   */
+  public function checkBreadcrumbItems(TableNode $table): void {
+    $selector = 'ol.ecl-breadcrumb__segments-wrapper li.ecl-breadcrumb__segment';
+    $breadcrumb = $this->getBreadcrumb();
+    $this->assertSession()->elementsCount('css', $selector, count($table->getHash()), $breadcrumb);
```
brummbar: In the future we will have more breadcrumb elements than two. Let's predict this.

tests/Behat/MinkContext.php
```diff
@@ -85,6 +103,38 @@ public function assertActiveLanguageSwitcherLink($label): void {
...
+    $breadcrumb_elements = $breadcrumb->findAll('css', $selector);
+
+    foreach ($table->getHash() as $key => $hash) {
+      /** @var \Behat\Mink\Element\NodeElement $breadcrumb_element */
+      $breadcrumb_element = $breadcrumb_elements[$key];
+      $element = $this->transformElement($hash['element']);
+      $value = $breadcrumb_element->find('css', $element)->getText();
+      Assert::assertEquals($value, $hash['text']);
```
brummbar: Always put the expected value as first parameter.

tests/Behat/MinkContext.php
```diff
@@ -100,4 +150,32 @@ protected function getLanguageSwitcherOverlay(): NodeElement {
...
+   * Get breadcrumb element.
+   *
+   * @return \Behat\Mink\Element\NodeElement
+   *   The breadcrumb element.
+   *
+   * @throws \Behat\Mink\Exception\ElementNotFoundException
+   */
+  protected function getBreadcrumb(): NodeElement {
+    $selector = 'div.ecl-page-header nav.ecl-breadcrumb';
+    $this->assertSession()->elementExists('css', $selector);
```
brummbar: This method returns the node element already, you can avoid the code duplication with the next return statement.

tests/features/page-header.feature
```diff
@@ -13,13 +13,25 @@ Feature: Page header block component.
     When I am an anonymous user
     And I go to the "Robots are everywhere" demo page
     Then I should see the heading "Robots are everywhere" in the "page header"
+    And the breadcrumb should contain the following items:
```
brummbar: And the breadcrumb trail should be "Home"
And the breadcrumb active element should be "Robots are everywhere"

brummbar: In case of multiple links, it would be something like:
`And the breadcrumb trail should be "Home, Term 1, Term 1-1"`

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -99,4 +123,34 @@ public function setTitle($title): self {
...
   }
 
+  /**
+   * Constructs a new PageHeaderBlock instance.
+   *
+   * @param array $build
+   *   Header build array.
+   *
+   * @return array
+   *   Header build array with the a breadcrumb (if any is applicable)
```
brummbar: The phpdoc here is wrong. You can also simplify the param and return docs with `array $build A render array` and `@return array the processed render array` or something like that.

tests/Behat/MinkContext.php
```diff
@@ -16,6 +17,23 @@ class MinkContext extends DrupalExtensionMinkContext {
 
   use UtilityTrait;
 
+  /**
+   * Mapping between human readable element labels and CSS selectors.
+   *
+   * @var array
+   */
+  private $breadcrumbElements = [];
```
brummbar: This is not used anymore so we can delete it.

tests/Behat/MinkContext.php
brummbar: Still to be adressed.

tests/Behat/MinkContext.php
```diff
@@ -85,6 +103,50 @@ public function assertActiveLanguageSwitcherLink($label): void {
...
+   * @param string $trail
+   *   Comma separated breadcrumb trail.
+   *
+   * @throws \Behat\Mink\Exception\ElementNotFoundException
+   * @throws \Behat\Mink\Exception\ExpectationException
+   *
+   * @Then the breadcrumb trail should be :items
+   */
+  public function checkBreadcrumbTrail(string $trail): void {
+    $trail_elements = explode(',', $trail);
```
brummbar: Use `', '` (extra space) to separate multiple elements.

tests/Behat/MinkContext.php
```diff
@@ -85,6 +103,50 @@ public function assertActiveLanguageSwitcherLink($label): void {
...
+    $selector = 'ol.ecl-breadcrumb__segments-wrapper li.ecl-breadcrumb__segment a';
+    $breadcrumb = $this->getBreadcrumb();
+    $this->assertSession()->elementsCount('css', $selector, count($trail_elements), $breadcrumb);
+    $breadcrumb_elements = $breadcrumb->findAll('css', $selector);
+
+    foreach ($trail_elements as $key => $trail_element) {
```
brummbar: We can make here an assertion that will help showing the difference between the expected and actual breadcrumb trail.
If you do something like this (see tests/Kernel/ContentLanguageSwitcherTest.php:140):
```
$actual = array_map('trim', array_column(iterator_to_array($breadcrumb_elements), 'nodeValue'));
Assert::assertEquals($trail_elements, $actual);
```

tests/Behat/MinkContext.php
```diff
@@ -100,4 +162,32 @@ protected function getLanguageSwitcherOverlay(): NodeElement {
...
+  /**
+   * Transform element label into an CSS selector, if any.
+   *
+   * @param string $label
+   *   Element label.
+   *
+   * @return string
+   *   CSS selector.
+   */
+  protected function transformElement($label): string {
```
brummbar: This can be now removed.

tests/features/page-header.feature
```diff
@@ -13,13 +13,19 @@ Feature: Page header block component.
     When I am an anonymous user
     And I go to the "Robots are everywhere" demo page
     Then I should see the heading "Robots are everywhere" in the "page header"
+    And the breadcrumb trail should be "Home"
```
brummbar: 👍

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
upchuk: OK!

PR #144, opened 2018-11-06
OPENEUROPA-1261: Implement "Files" pattern with variants

nagyad:
We found several technical problems with @ademarco and we came up with this implementation. Basically we created a file pattern field type with a value object in order to handle file occurrences in general.


oe_theme.theme
```diff
@@ -602,3 +604,134 @@ function oe_theme_preprocess_paragraph__oe_content_row__variant_inpage(&$variabl
...
+    // MultilingualHelper::getEuropeanUnionLanguageList() in configuration.
+    $languages = \Drupal::service('oe_multilingual.helper')->getLanguageNameList();
+    if (isset($languages[$language_code])) {
+      $language_name = $languages[$language_code];
+    }
+  }
+
+  // @todo: Complete and/or centralise extension to icon conversion.
+  // Get icons from given file extension.
+  switch ($file->getExtension()) {
```
ademarco: We should create a follow up for this.

src/ValueObject/FileValueObject.php
```diff
@@ -0,0 +1,255 @@
...
+      $file->setLanguageCode($values['language_code']);
+    }
+
+    return $file;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function fromAny($value): ValueObjectInterface {
```
ademarco: We have been discussing with other team members and we all kind of felt uneasy at a certain point about this. Should we remove this generic constructor? For the moment let's just use the from array one in the preprocess when in a preview context. Make sure also documentation is updated. We can leave the tests for `fromFileEntity` as we will probably use that to style media entities soon enough.

src/ValueObject/ValueObjectBase.php
```diff
@@ -0,0 +1,43 @@
...
+   * {@inheritdoc}
+   */
+  public function offsetGet($offset) {
+    return $this->toArray()[$offset];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function offsetSet($offset, $value) {
```
upchuk: Why are the setters needed?

upchuk: A, just saw the ArrayAccess. OK

docs/developer-documentation.md
```diff
@@ -1,32 +1,49 @@
...
+The OpenEuropa Theme exposes ECL components using [UI Patterns][1] plugins, allowing them to be used seamlessly as drop-in
 templates for panels, field groups, views, paragraphs, nodes, etc.
 
 Patterns are located in [`./templates/patterns`](../templates/patterns), each sub-directory contains a pattern along with
-its variants. Pattern definitions are stored in a YAML file (`[PATTERN-NAME].ui_patterns.yml`) along with their Twig templates.
+its variants. Pattern definitions are stored in as YAML files (for ex. `[PATTERN-NAME].ui_patterns.yml`) along with their
+Twig templates (for ex. `pattern-[PATTERN-NAME].html.twig`).
 
-Each pattern definition exposes a list of fields that can be used to pass data to be rendered using the pattern template.
+Each pattern definition exposes a list of fields that can be used to pass data that will will be rendered using the
```
upchuk: `will be rendered`.

docs/developer-documentation.md
```diff
@@ -1,32 +1,49 @@
...
-Value objects are available at [`./src/ValueObject`](../src/ValueObject) and implement the `\Drupal\oe_theme\ValueObject\ValueObjectInterface`.
+Value objects are available at [`./src/ValueObject`](../src/ValueObject) and implement `\Drupal\oe_theme\ValueObject\ValueObjectInterface`.
 
-## Using value objects
+Value objects can be constructed only by using one of their factory methods. By default all value objects expose the
+`ValueObjectInterface::fromArray()` factory which builds and returns a value object from a given array.
 
-Value objects can be constructed only by using one of their factory methods. By default all value objects will expose the
-following factory:
+When patterns are rendered programmatically value objects can be passed directly to related pattern fields, as shown below:
```
upchuk: comma after "programatically"

docs/developer-documentation.md
```diff
@@ -1,32 +1,49 @@
...
+  '#fields' => [
+    'foo' => FooValueObject::fromArray(['bar' => 'Bar']),
+  ]
+];
+
+\Drupal::service('renderer')->render($elements);
+
+```
+
+Value objects will typically be constructed in pattern template preprocess functions, for example, when rendering a pattern
```
upchuk: Actually, they will typically be constructed in other preprocess functions, not those of patterns. Most notably the pattern preview because there we can only pass the values as array.

ademarco: That is not entirely true, core components will be using generic view mode preprocess function but site builders can get advantage of the modules that are shipped with UI Patterns, which can offer pattern theme suggestions based on rendering contexts, such as a view mode, a view row display etc.

ademarco: But I'll rephrase it as it is misleading the way it is now.

PR #147, opened 2018-11-20
OPENEUROPA-1346: Use Drupal core services to access native language names

pfrenssen:


pfrenssen:
Recently a bunch of deprecation warnings have been added to Twig to warn people about upcoming changes in Twig 2.0. The tests were failing because of these warnings. This has already been fixed in Drupal 8.7. Since we are still on Drupal 8.6 I have included this patch so the tests can pass.

While working on this I also added missing test coverage and documentation for some other filters in the same Twig extension.

modules/oe_theme_helper/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+{{ langcode|to_language }}
+```
+
+Output: `Spanish`.
+
+### to_native_language
+
+The `to_language` filter will convert a language code into the full language
```
upchuk: The `to_native_language` should be here.

composer.json
```diff
@@ -52,6 +52,11 @@
     "extra": {
         "composer-exit-on-patch-failure": true,
         "enable-patching": true,
+        "patches": {
+            "drupal/core": {
+                "Deprecation warnings in Twig code cause test failures. https://www.drupal.org/project/drupal/issues/2600154": "https://www.drupal.org/files/issues/2018-08-13/2600154-3-60.patch"
```
ademarco: This patch should be included in openeuropa/drupal-core-require-dev and not directly on the theme. I've created the following issue for that: https://webgate.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-1387

PR #148, opened 2018-11-23
OPENEUROPA-1344: Complete conversion of file extension to ECL icons.

nagyad:
Improved the mapping between the ECL file icon class names and file extensions.


modules/oe_theme_helper/src/TwigExtension/Filters.php
```diff
@@ -59,24 +59,52 @@ public function toNativeLanguageName($language_code): string {
...
+   *   File icon class name.
    */
-  public function toFileIcon($extension): string {
-    // @todo: Complete and/or centralise extension to icon conversion.
-    // Get icons from given file extension.
-    switch ($extension) {
-      case 'ppt':
-        return 'presentation';
+  public function toFileIcon($extension) {
+    $extansion_mapping = [
```
drupol: Typo $extension

modules/oe_theme_helper/src/TwigExtension/Filters.php
```diff
@@ -59,24 +59,52 @@ public function toNativeLanguageName($language_code): string {
...
+        'm4v',
+        'webm',
+      ],
+    ];
 
-      default:
-        return 'file';
+    foreach ($extansion_mapping as $file_type => $extensions) {
+      if (in_array($extension, $extensions)) {
+        return $file_type;
```
drupol: :+1:

modules/oe_theme_helper/src/TwigExtension/Filters.php
```diff
@@ -59,24 +59,52 @@ public function toNativeLanguageName($language_code): string {
...
-   *   File icon.
+   *   File icon class name.
    */
-  public function toFileIcon($extension): string {
-    // @todo: Complete and/or centralise extension to icon conversion.
-    // Get icons from given file extension.
-    switch ($extension) {
-      case 'ppt':
-        return 'presentation';
+  public function toFileIcon($extension) {
```
brummbar: Here we are missing `string` type for extension and `: string` return type was removed.

brummbar: The `$extension` might come in uppercase (sometimes images have `.JPG` instead of `.jpg`). To improve this I would lowercase the `$extension` here so we are sure to find the proper class.

tests/fixtures/rendering.yml
```diff
@@ -250,31 +250,42 @@
     '#id': file_link
     '#fields':
       files:
-        - title: "Example document 1"
-          mime: "application/pdf"
-          url: "http://example.com/document1.pdf"
+        - title: "Example document"
```
brummbar: Once you add the lowercase part above, add a test with an uppercase extension too.

tests/fixtures/rendering.yml
```diff
@@ -250,31 +250,42 @@
...
+      'a[href="http://example.com/document.doc"]': 1
+      'a[href="http://example.com/presentation.ppt"]': 1
+      'a[href="http://example.com/video.mpeg"]': 1
+      'a[href="http://example.com/spreadsheet.xls"]': 1
+      'span.ecl-icon--file': 1
+      'span.ecl-icon--presentation': 1
+      'span.ecl-icon--video': 1
+      'span.ecl-icon--spreadsheet': 1
+    equals:
+      'a[href="http://example.com/document.doc"] span.ecl-file__title': "Example document"
```
brummbar: If you use this selector:
`a[href="http://example.com/document.doc"] span.ecl-icon--file + span.ecl-file__title`
you can assert also that the correct icon is placed in the correct place.

brummbar: Update the selector also for the other lines.

modules/oe_theme_helper/src/TwigExtension/Filters.php
drupol: Nice remarks !
I would then also suggest to take in account .jpeg and .JPEG

drupol: Oops, forget what I just said!

PR #150, opened 2018-12-03
OPENEUROPA-1378: Implement DateValueObject for Date pattern fields

drupol:


templates/patterns/date_block/date_block.ui_patterns.yml
```diff
@@ -0,0 +1,42 @@
...
+      type: "string"
+      label: "Month"
+      description: "(optional) Numeric or textual representation of a month."
+      preview: "9"
+    year:
+      type: "integer"
+      label: "Year"
+      description: "(optional) A full numeric representation of a year, 4 digits."
+      preview: "2018"
+    timestamp:
```
imanoleguskiza: Why did you add this parameter and added all the extra functionality? ECL doesn't mention it and I don't see it as a requirement in the ticket

src/ValueObject/DateValueObject.php
```diff
@@ -0,0 +1,91 @@
...
+   * @param string $day
+   *   The date day.
+   * @param string $month
+   *   The date month.
+   * @param string $year
+   *   The date year.
+   * @param string $variant
+   *   The pattern variant.
+   */
+  private function __construct(string $day, string $month, string $year, string $variant = 'default') {
```
ademarco: I'd keep the current approach with setters and getters for now, we have a follow up issue to consider the alternative proposed by @drupol , until that's not discussed and agreed we should not change direction. Refactorying would be relatively quick anyway and can be assigned to a person different than @drupol so to spread more the knowledge about the new approach, if accepted.

tests/Unit/ValueObject/DateValueObjectTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Test date value object.
+ */
+class DateValueObjectTest extends UnitTestCase {
+
+  /**
+   * Test constructing a date value object from an array.
+   */
+  public function testFromArray() {
```
imanoleguskiza: If we are testing the fromArray method I guess testing the other methods (fromTimestamp, from DateTimePlus) is also a good idea

tests/fixtures/rendering.yml
```diff
@@ -1,3 +1,59 @@
+- array:
```
imanoleguskiza: We only have rendering tests that use the timestamp, in theory it should be possible to pass the individual variables and so I think we should test it as well.

templates/patterns/date_block/date_block.ui_patterns.yml
```diff
@@ -0,0 +1,42 @@
...
+      type: "integer"
+      label: "Timestamp"
+      description: "(optional) A full numeric representation of a year, 4 digits. When this is set, it will override other parameters."
+      preview: "1543583805"
```
ademarco: We should have here only one field called "date" and of "type" DateValueObject, then in the preprocess we build the value object.

templates/patterns/date_block/date_block.ui_patterns.yml
drupol: Fixed.

tests/Unit/ValueObject/DateValueObjectTest.php
drupol: Ok

templates/patterns/date_block/date_block.ui_patterns.yml
drupol: Ok

src/ValueObject/DateValueObject.php
drupol: See #151

tests/fixtures/rendering.yml
drupol: Done.

tests/Unit/ValueObject/DateValueObjectTest.php
drupol: I removed the methods from the DateValueObject because they are no more relevant for now.

src/ValueObject/DateValueObject.php
drupol: @ademarco I'm using variables in the constructor, nothing else for now... I guess this comment is outdated then.
I removed the "withers" as well.

tests/fixtures/rendering.yml
```diff
@@ -321,27 +321,6 @@
     equals:
       'h3.ecl-list-item__title': "Item title"
       'div.ecl-list-item__detail': "Item text"
-- array:
```
imanoleguskiza: Why was this test removed?

tests/Unit/ValueObject/DateValueObjectTest.php
```diff
@@ -0,0 +1,35 @@
...
```
imanoleguskiza: Do we need to keep this test if we already have the DateBlockPatternRenderingTest.php? If so, we still need a test for the fromTimestamp method here

tests/fixtures/rendering.yml
drupol: As the pattern is using a `DateValueObject`, we need to deal with an object, not only an array.
This is why this test was removed from there and moved into `DateBlockPatternRenderingTest.php`

tests/Kernel/Patterns/DateBlockPatternRenderingTest.php
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /**
+   * Data provider for testFromTimestamp.
+   *
+   * @return array
+   *   An array of test data arrays with assertions.
+   */
+  public function fromTimestampDataProvider(): array {
+    return [
```
voidtek: Why we are not using an external yml file, as `renderingDataProvider`?

src/ValueObject/DateValueObject.php
```diff
@@ -0,0 +1,152 @@
...
+   * @param string $year
+   *   The date year.
+   * @param string|int|null $weekDay
+   *   The day of the week.
+   */
+  private function __construct(string $day, string $month, string $year, string $weekDay = NULL) {
+    $this->day = $day;
+    $this->month = $month;
+    $this->year = $year;
+    $this->weekDay = $weekDay;
```
voidtek: This line is useless, duplicate of  
```
$this->weekDay = empty($weekDay) ?
      $date->format('l') :
      $weekDay;
```

src/ValueObject/DateValueObject.php
```diff
@@ -0,0 +1,152 @@
...
+
+    $date = new DateTimePlus(implode('-', [$year, $month, $day]));
+
+    $this->weekDay = empty($weekDay) ?
+      $date->format('l') :
+      $weekDay;
+    $this->monthName = $date->format('F');
```
voidtek: I am not comfortable to have this `monthName` created only by the `DateTimePlus` and not an override if it's `NULL`

src/ValueObject/DateValueObject.php
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function fromArray(array $parameters = []): ValueObjectInterface {
+    return new static(
+      $parameters['day'],
+      $parameters['month'],
+      $parameters['year']
```
voidtek: the parameter `weekDay` is missing.

templates/patterns/date_block/date_block.ui_patterns.yml
```diff
@@ -0,0 +1,25 @@
...
+      description: "Past event"
+  fields:
+    date:
+      type: "DateValueObject"
+      label: "Date"
+      description: "A date value object."
+      preview:
+        day: "24"
+        month: "09"
+        year: "1981"
```
voidtek: we should add the parameter `weekday: "Thu."`

ademarco:
After discussing it with @voidtek we realized that it would be simpler if we just construct the date using timestamps, if we do so we can easily accommodate date periods, which is actually in line with ECL specifications.

tests/Kernel/fixtures/patterns/date_block_pattern_rendering.yml
```diff
@@ -0,0 +1,50 @@
...
+      div.ecl-date-block.ecl-date-block--default: 1
+      div.ecl-date-block.ecl-date-block--cancelled: 0
+      div.ecl-date-block.ecl-date-block--past: 0
+    equals:
+      span.ecl-date-block__week-day: Thu
+      span.ecl-date-block__day: 24
+      span.ecl-date-block__month: Sep
+      span.ecl-date-block__year: 1981
+
+# Test "cancelled" variant.
```
voidtek: We forget to check the variant "ecl-date-block ecl-date-block--ongoing"

src/ValueObject/DateValueObject.php
```diff
@@ -0,0 +1,140 @@
...
+      $this->end = DateTimePlus::createFromTimestamp($end, $timezone);
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function fromArray(array $parameters = []): ValueObjectInterface {
+    return new static(
+      $parameters['start'],
```
voidtek: We should check if $parameters['start'] exists.

src/ValueObject/DateValueObject.php
```diff
@@ -0,0 +1,140 @@
...
+
+    if ($end !== NULL) {
+      $this->end = DateTimePlus::createFromTimestamp($end, $timezone);
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function fromArray(array $parameters = []): ValueObjectInterface {
```
voidtek: `ValueObjectInterface` should be `DateValueObjectInterface`

src/ValueObject/DateValueObject.php
```diff
@@ -0,0 +1,140 @@
...
+   *   Format to be used to print the interval.
+   * @param string $extra
+   *   Run extra check to make sure we should not print an interval.
+   *   This is useful, for example, in case of same days on different months
+   *   which should still be printed as an interval.
+   *
+   * @return string
+   *   Formatted interval.
+   */
+  protected function getDateInterval(string $format, string $extra): string {
```
voidtek: We should improve this function by putting variables, to be consistent.

src/ValueObject/DateValueObjectInterface.php
```diff
@@ -0,0 +1,67 @@
...
+   *   Start date as UNIX timestamp.
+   * @param int $end
+   *   End date as UNIX timestamp.
+   * @param string|null $timezone
+   *   Timezone string, e.g. "Europe/Brussels".
+   *
+   * @return \Drupal\oe_theme\ValueObject\DateValueObjectInterface
+   *   A new DateValueObject.
+   */
+  public static function fromTimestamp(int $start, int $end = NULL, string $timezone = NULL): DateValueObjectInterface;
```
voidtek: The function `fromArray ` is missing.

src/ValueObject/DateValueObject.php
drupol: Fixed in upcoming commit.

src/ValueObject/DateValueObject.php
drupol: No, you cannot override what's in the Interface.

tests/Kernel/fixtures/patterns/date_block_pattern_rendering.yml
drupol: Fixed.

tests/Kernel/fixtures/patterns/date_block_pattern_rendering.yml
```diff
@@ -0,0 +1,91 @@
...
+      div.ecl-date-block.ecl-date-block--past: 1
+      div.ecl-date-block.ecl-date-block--ongoing: 0
+    equals:
+      span.ecl-date-block__week-day: Thu
+      span.ecl-date-block__day: 24
+      span.ecl-date-block__month: Sep
+      span.ecl-date-block__year: 1981
+
+# Test "unknown" variant.
+- variant: unexistant
```
voidtek: This variant is out of scope and I don't see the purpose.
This comment is not a blocker.

drupol: I think this should be part of the test, to check if when rendering a variant that is unknown, it doesn't set any specific existing classes.

ademarco: I think it's good to test what happen when the variant is either empty or not a valid one, we should then fallback to the default variant. We can add this logic directly in the pattern template or in the global pattern preprocess.

drupol: Fixed!

tests/Kernel/Patterns/DateBlockPatternRenderingTest.php
drupol: Same reason as previously explained.

src/ValueObject/DateValueObject.php
drupol: Fixed.

src/ValueObject/DateValueObject.php
drupol: Fixed.

src/ValueObject/DateValueObject.php
drupol: Fixed.

templates/patterns/date_block/date_block.ui_patterns.yml
drupol: Fixed.

oe_theme.theme
```diff
@@ -644,6 +641,26 @@ function oe_theme_preprocess_paragraph__oe_content_row__variant_inpage(array &$v
   $variables['links'] = $links;
 }
 
+/**
+ * Implements hook_preprocess_pattern().
+ */
+function oe_theme_preprocess_pattern_date_block(array &$variables): void {
+  // This make sure that the variant is not empty and has a "default" value.
+  // @todo: How about doing this in all components ?
+  // @todo: We should optimize how to compute the list of variants.
```
ademarco: Comment should rather be:
```
// This make sure that the variant value will always fallback to "default" if an empty or not valid value is passed.
```

I'd also remove the todo here since it's not something to do about the current block of code, I'd rather open a follow up to do this on all components.

drupol: Updating now.

docs/developer-documentation.md
```diff
@@ -85,6 +85,24 @@ Check the [UI Patterns documentation][2] for more information about how to use p
...
 
+#### `DateValueObject`
+
+Used in the following patterns:
+
+- [`date_block`](../templates/patterns/date_block/date_block.ui_patterns.yml)
+
+Provides the following factory methods:
+
+- `DateValueObject::fromArray(array $values = [])`: accepts an array with the following properties:
```
imanoleguskiza: The PR is fine, but do we really need this static method? Its essentially the same as the fromTimestamp method but a bit worse since fromTimestamp tells you what parameters to pass.

drupol: We need this, yes. It's defined in the the interface... and we cannot go over it.

PR #152, opened 2018-12-04
OPENEUROPA-1381: Use Url objects on ui_patterns.

imanoleguskiza:
### Description

Use Drupal core \Drupal\Core\Url objects wherever necessary in custom ui_patterns.

### Change log

- Changed: Url fields on ui_patterns now use \Drupal\Core\Url objects


drupol:
The branch was out of date, I rebased it.

PR #153, opened 2018-12-10
OPENEUROPA-1474: Fix fromFileEntity method to provide file URL by URI.

nagyad:
Fixing a bug in the FileValueObject.

### DIFF
--- a/tests/Unit/ValueObject/FileValueObjectTest.php
+++ b/tests/Unit/ValueObject/FileValueObjectTest.php
@@ -4,10 +4,8 @@ declare(strict_types = 1);
 
 namespace Drupal\Tests\oe_theme\Unit\Patterns;
 
-use Drupal\Core\Language\LanguageInterface;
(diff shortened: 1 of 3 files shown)


tests/Unit/ValueObject/FileValueObjectTest.php
```diff
@@ -42,46 +42,4 @@ public function testFromArray() {
...
-    $file = FileValueObject::fromFileEntity($file_entity);
-
-    $this->assertEquals('123', $file->getSize());
-    $this->assertEquals('pdf', $file->getMime());
-    $this->assertEquals('http://example.com/test.pdf', $file->getUrl());
-    $this->assertEquals('Test.pdf', $file->getName());
-    $this->assertEquals('pdf', $file->getExtension());
-    $this->assertEquals('fr', $file->getLanguageCode());
-  }
-
```
ademarco: If we remove this test in favor of a kernel test we should make sure that the kernel test covers all cases that were covered here, such as the language code.

PR #156, opened 2019-01-15
OPENEUROPA-1522: Make sure we don't use Drupal paths in Behat scenarios.

sergepavle:
## OPENEUROPA-1522


composer.json
```diff
@@ -20,8 +20,8 @@
         "drupal/styleguide": "~1.0-alpha3",
         "drush/drush": "~9.0@stable",
         "nikic/php-parser": "~3.0",
-        "openeuropa/code-review": "^1.0.0-alpha4",
         "openeuropa/behat-transformation-context" : "~0.1",
+        "openeuropa/code-review": "^1.0.0-alpha4",
```
drupol: :heart:

behat.yml.dist
```diff
@@ -24,9 +24,11 @@ default:
               priorities dropdown menu: "#nav-menu-expandable-group-3 .ecl-navigation-menu__links"
               search box: ".demo-block-header-search"
               sites switcher: ".demo-block-site-switcher"
+              link block list: ".ecl-link-block__list"
```
ademarco: This selector represent the actual dropdown content, so the proper name would be:
```
              dropdown content: ".ecl-link-block__list"
```
This will make it understandable  by business.

PR #169, opened 2019-01-30
Issue #168: InvalidArgumentException the language code fr does not exist or is not enabled.

### LINKED ISSUE #168: InvalidArgumentException: The language code fr does not exist or is not enabled.
### Description
Hello,

I noticed the following error when visiting /components:
```
The website encountered an unexpected error. Please try again later.
InvalidArgumentException: The language code fr does not exist or is not enabled. in Drupal\oe_theme_helper\TwigExtension\Filters->toNativeLanguageName() (line 73 of themes/contrib/oe_theme/modules/oe_theme_helper/src/TwigExtension/Filters.php).
```

I'm currently using https://github.com/openeuropa/drupal-site-template
But I'm following the latest oe_theme v0.9.0 and I don't have oe_multilingual enabled.
So I don't have 'fr' or any enabled languages at the moment other than 'en'.

### Steps to reproduce
Install oe_theme on a website which doesn't have any languages OR manually remove 'fr' from the final install of oe_theme. And clear cache.

Then visit /components -> InvalidArgumentException.

**Expected behavior:** No error...

**Actual behavior:** Error...

**Reproduces how often:** 100% of the time in the above conditions.

### Versions
oe_theme v0.9.0 and latest master

### Additional information
There are several references to specific languages in file ui_patterns.
```
templates/patterns/file/file.ui_patterns.yml > fr
templates/patterns/file_translation/file_translation.ui_patterns.yml > fr, it, hu
```

I suppose the goal is no to have a hard dependency on oe_multilingual.
I suggest to relax a bit `\Drupal\oe_theme_helper\TwigExtension\Filters::toNativeLanguageName()` by not throwing an exception and adopt the same behavior as `\Drupal\Core\Language\LanguageManager::getLanguageName()`.

Regards,
Hervé

vever001:
Here's a first proposal to resolve #168.

Let me know,
Hervé

### DIFF
--- a/modules/oe_theme_helper/src/TwigExtension/Filters.php
+++ b/modules/oe_theme_helper/src/TwigExtension/Filters.php
@@ -63,15 +63,15 @@ class Filters extends \Twig_Extension {
    * @param string $language_code
    *   The language code as defined by the W3C language tags document.
    *
-   * @return string
+   * @return string|\Drupal\Core\StringTranslation\TranslatableMarkup
    *   The native language name.
    */
-  public function toNativeLanguageName($language_code) {
+  public function toNativeLanguageName(string $language_code) {
     if (empty($language_code)) {
       return $this->t('Unknown');
     }
 
-    if ($language_code == LanguageInterface::LANGCODE_NOT_SPECIFIED) {
+    if ($language_code === LanguageInterface::LANGCODE_NOT_SPECIFIED) {
       return $this->t('None');
     }
 
--- a/modules/oe_theme_helper/src/TwigExtension/Filters.php
+++ b/modules/oe_theme_helper/src/TwigExtension/Filters.php
@@ -4,7 +4,9 @@ declare(strict_types = 1);
 
 namespace Drupal\oe_theme_helper\TwigExtension;
 
+use Drupal\Core\Language\LanguageInterface;
 use Drupal\Core\Language\LanguageManagerInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
 
 /**
  * Collection of extra Twig filters.
@@ -13,6 +15,8 @@ use Drupal\Core\Language\LanguageManagerInterface;
  * coming straight from Twig templates.
  */
 class Filters extends \Twig_Extension {
+
+  use StringTranslationTrait;
 
   /**
    * The language manager.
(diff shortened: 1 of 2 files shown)


modules/oe_theme_helper/src/TwigExtension/Filters.php
```diff
@@ -61,16 +65,22 @@ public function toLanguageName($language_code): string {
    *
    * @return string
    *   The native language name.
-   *
-   * @throws \InvalidArgumentException
-   *   Thrown when the passed in language code does not exist.
    */
-  public function toNativeLanguageName($language_code): string {
+  public function toNativeLanguageName($language_code) {
```
drupol: Do you think it would b possible to add a type cast to the argument `$language_code` and to the function return as well ?

modules/oe_theme_helper/src/TwigExtension/Filters.php
```diff
@@ -61,16 +65,22 @@ public function toLanguageName($language_code): string {
...
+    if (empty($language_code)) {
+      return $this->t('Unknown');
+    }
+
+    if ($language_code == LanguageInterface::LANGCODE_NOT_SPECIFIED) {
```
drupol: Use strict comparison.

modules/oe_theme_helper/src/TwigExtension/Filters.php
```diff
@@ -61,16 +65,22 @@ public function toLanguageName($language_code): string {
...
+      return $this->t('None');
```
drupol: Don't forget to declare the return type of the function.

modules/oe_theme_helper/src/TwigExtension/Filters.php
vever001: The function can return a string or a TranslatableMarkup so I should pbly update the return doc to:
`@return string|\Drupal\Core\StringTranslation\TranslatableMarkup`
But I don't think I can set the return type since it's multiple.

For the argument do you mean something like this?:
`public function toNativeLanguageName(string $language_code) {`
I think that should be ok.

drupol: Thanks Hervé.

I think it's always better to have functions that returns on single type of value, but sometimes it's not possible.

For the argument, yes, it's fine, thanks mate.

modules/oe_theme_helper/src/TwigExtension/Filters.php
vever001: Done

modules/oe_theme_helper/src/TwigExtension/Filters.php
vever001: Updated to `@return string|\Drupal\Core\StringTranslation\TranslatableMarkup`

modules/oe_theme_helper/src/TwigExtension/Filters.php
vever001: I made the changes, let me know.

ademarco:
@vever001 we have actually fixed this and it's now available on version 0.11.0, this is the relevant PR #178. We had to slightly change the approach but it should cover your case.

PR #173, opened 2019-02-11
OPENEUROPA-1603: Provide a simple search box that will submit searched terms to the external Europa search service.

sergepavle:
### Description

Provide a simple search box that will submit searched terms to the external Europa search service

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "openeuropa/drupal-core-require-dev": "8.6.x-dev",
         "openeuropa/oe_multilingual": "~0.2.3",
         "openeuropa/oe_paragraphs": "~0.3",
+        "openeuropa/oe_search": "dev-OPENEUROPA-1603-search-block",
         "openeuropa/task-runner": "~1.0-beta4",
         "phpunit/phpunit": "~5.5||~6.0",
         "symfony/dom-crawler": "~3.4"
@@ -47,6 +48,10 @@
         "drupal": {
             "type": "composer",
             "url": "https://packages.drupal.org/8"
+        },
+        "openeuropa": {
+            "type": "git",
+            "url": "https://github.com/openeuropa/oe_search.git"
         }
     },
     "extra": {
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -714,3 +714,58 @@ function oe_theme_preprocess_pattern_list_item__variant_date__preview(array &$va
 function oe_theme_preprocess_pattern_date_block__preview(array &$variables): void {
   $variables['date'] = DateValueObject::fromArray($variables['date']);
 }
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function oe_theme_theme_suggestions_form_alter(array &$suggestions, array $variables): void {
+  $suggestions[] = $variables['theme_hook_original'] . '__' . $variables['element']['#form_id'];
+}
+
+/**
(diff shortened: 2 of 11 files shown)


composer.json
```diff
@@ -47,6 +48,10 @@
         "drupal": {
             "type": "composer",
             "url": "https://packages.drupal.org/8"
+        },
+        "openeuropa": {
```
drupol: We should not add the name to the repo.

```yaml
 "repositories": [
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        },
       {        
           "type": "git",
           "url": "https://github.com/openeuropa/oe_search.git"
       }
]

```

oe_theme.theme
```diff
@@ -714,3 +714,58 @@ function oe_theme_preprocess_pattern_list_item__variant_date__preview(array &$va
...
+/**
+ * Helper function to additionally preprocess textfield, search elements.
+ *
+ * Moves the label element to the input template and creates some helper
+ * variables to be used with the ECL templates.
+ *
+ * @param array $element
+ *   Set of element data.
+ */
+function _oe_theme_preprocess_input_text(array &$element): void {
```
drupol: Helpers should be in another file maybe?

sergepavle: I will create follow up ticket. Because this "helper" consistent to other analogical helper function inside oe_theme.theme file. I believe scope of this new ticket should be refining approach how we could decouple all future helper function (including files to oe_theme.theme (old scool approach), use classes or even using new twig filters).

drupol: I think this should be done asap in order to keep a proper order.
the file oe_theme.theme should only contain the hooks, all other methods and functions should be in another file.

tests/features/search-form.feature
```diff
@@ -0,0 +1,12 @@
+@api
+Feature: Search form block
```
imanoleguskiza: I don't think we need this test here. The oe_search component already makes sure the functionality works.

tests/Behat/MinkContext.php
```diff
@@ -160,4 +160,21 @@ protected function getBreadcrumb(): NodeElement {
...
+   * Assert redirect to expected url.
+   *
+   * @param string $uri
+   *   Expected redirect url.
+   *
+   * @Then I should be redirected to :uri
+   *
+   * @throws \Exception
+   */
+  public function assertRedirect(string $uri): void {
```
imanoleguskiza: See below.

tests/features/search-form.feature
upchuk: @imanoleguskiza Yes we do. The component tests the functionality but this tests that the homepage has it. This block is the final block that needs to be placed there by the theme on all pages. @sergepavle Potentially we can test that it's available also on other pages no?

imanoleguskiza: Then the last 3 steps are not required and neither is the custom context. Agree testing other pages would be nice.

upchuk: @imanoleguskiza Well, it doesn't hurt to have it does it?

sergepavle: I have removed 3 steps and added check presents search block on other pages.
Also I have added Kernel tests for testing rendering block with ECL components (like for other elements of a pages like breadcrumbs, language switcher etc.).

PR #178, opened 2019-02-21
OPENEUROPA-1577: InvalidArgumentException the language code fr does not exist or is not enabled.

sergepavle:
### Description

The scope of this ticket is to follow up on this bug report: InvalidArgumentException the language code fr does not exist or is not enabled.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_helper/src/TwigExtension/Filters.php
```diff
@@ -70,7 +71,52 @@ public function toNativeLanguageName($language_code): string {
...
       return $languages[$language_code]->getName();
     }
-    throw new \InvalidArgumentException("The language code $language_code does not exist or is not enabled.");
+    // The fallback implemented in case we don't have enabled language.
+    $predefined = array_merge(self::getEuropeanUnionLanguageList(), LanguageManager::getStandardLanguageList());
+    if (!empty($predefined[$language_code][1])) {
+      return $predefined[$language_code][1];
+    }
+
+    throw new \InvalidArgumentException("The language code $language_code does not exist.");
```
drupol:
```suggestion
    throw new \InvalidArgumentException('The language code ' . $language_code . ' does not exist.');
```

tests/Kernel/Patterns/ComponentsPageTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Test a components page rendering.
+   *
+   * @throws \Exception
+   */
+  public function testComponentsPageRendering(): void {
+    $controller = new PatternsLibraryController($this->container->get('plugin.manager.ui_patterns'));
+    $renderable_array = $controller->overview();
+    $this->container->get('renderer')->renderRoot($renderable_array);
```
drupol: Do you think it would be possible to assert that the page has been rendered without issue ?
It would be nicer to have a more self-explanatory test because at first sight, it seems that this method (_which is supposed to have assertions_) doesn't test anything.

sergepavle: Yes. Your doubts is correct. I have found that this approach didn't covered all cases. I reworked testing approach just by changing \Drupal\Tests\oe_theme\Kernel\RenderingTest class.

PR #183, opened 2019-02-28
 OPENEUROPA-1675: Provide fallback title for inpage navigation variant.

nagyad:


tests/Kernel/Paragraphs/ContentRowTest.php
```diff
@@ -190,6 +190,23 @@ public function testSideMenuTranslation(): void {
     $left_column = $crawler->filter('.ecl-row .ecl-col-md-3.ecl-u-z-navigation');
     $this->assertContains('French page navigation', $left_column->html());
     $this->assertContains('French rich text title', $left_column->html());
+
+    // Create the main content row paragraph with empty title.
```
imanoleguskiza: Its a bit weird to add this test inside the paragraph translation test, but I would like to have the next reviewers input before asking for changes.

nagyad: We are testing if the fallback title gets translated as well.

nagyad:
Rebased.

tests/Kernel/Paragraphs/ContentRowTest.php
imanoleguskiza: Fair enough :+1:

sergepavle: Yes. I agree. It is make sense to move this code in other place.

nagyad:
Did rebase.

PR #185, opened 2019-03-05
OPENEUROPA-1693: Create UI pattern for footer

sergepavle:
### Description

Create UI Pattern in theme (PR in theme) -> ECL footer
Add rendering Kernel test

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -776,11 +776,26 @@ function _oe_theme_preprocess_input_text(array &$element): void {
  * Implements hook_preprocess_HOOK().
  */
 function oe_theme_preprocess_pattern_footer(array &$variables): void {
-  $social_media_links = &$variables['corporate_footer']['social_media']['items'] ?? [];
   // Workaround related to UI Pattern module feature
-  // which recognize array with "type" key element as renderable.
-  foreach ($social_media_links as $key => $link) {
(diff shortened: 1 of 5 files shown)


oe_theme.theme
```diff
@@ -771,3 +771,16 @@ function _oe_theme_preprocess_input_text(array &$element): void {
...
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_pattern_footer(array &$variables): void {
+  $social_media_links = &$variables['corporate_footer']['social_media']['items'] ?? [];
+  // Workaround related to UI Pattern module feature
+  // which recognize array with "type" key element as renderable.
+  foreach ($social_media_links as $key => $link) {
```
voidtek: @sergepavle : Why you don't use the reference on the foreach?
```
foreach ($social_media_links as &$link) {
    $link['type'] = $link['link_type'];
    unset($link['link_type']);
}
```

ademarco: @voidtek I've checked and it seems we don't use links value objects anywhere. Anyway we decided not to implement this as a pattern since this is not a reusable theme component.

ademarco:
@sergepavle you can close this PR.

PR #201, opened 2019-04-16
OPENEUROPA-1774: Apply typography styling inside rich text fields

upchuk:
* Body field
* Accordion item long field
* Rich text long field


templates/field/field--paragraph--field-oe-text-long--oe-rich-text.html.twig
```diff
@@ -6,6 +6,6 @@
  * @see ./core/themes/stable/templates/field/field.html.twig
  */
 #}
-<div class="ecl-paragraph ecl-paragraph--m">
+<div class="ecl-editor">
```
ademarco: This is correct, I've double check with INNO team. Couldn't we have only one template that applies to all `field_oe_text_long`? like:
```
templates/field/field--paragraph--field-oe-text-long.html.twig
```

PR #215, opened 2019-04-25
OPENEUROPA-1798: Page header content is not reflecting changes made before node is published

hernani:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRouteBase.php
```diff
@@ -0,0 +1,129 @@
...
+   * Returns the node retrieved from the current route.
+   *
+   * @return \Drupal\node\NodeInterface|null
+   *   The node entity, or NULL if not found.
+   */
+  protected function getNode(): ?NodeInterface {
+    $supported = [
+      'entity.node.canonical',
+      'entity.node.latest',
+      'entity.node.revision',
```
upchuk: It's missing the route: `entity.node.latest_version`. This is what users see as part of the editorial workflow. And I wonder if `entity.node.latest` actually exists anywhere?

brummbar: Yep I think `entity.node.latest` was a mistake of mine.

PR #217, opened 2019-04-25
OPENEUROPA-1807: featured_image must be after content_moderation_control

dxvargas:
### Description

Change weight for `featured_image` to appear after `content_moderation_control`.

### Change log

- Changed: Weight for `featured_image` decreased from -100 to -19.

### DIFF
--- a/modules/oe_theme_helper/oe_theme_helper.module
+++ b/modules/oe_theme_helper/oe_theme_helper.module
@@ -1,0 +1,24 @@
+<?php
+
+/**
+ * @file
+ * The OE Theme Helper module.
+ */
+
+declare(strict_types = 1);
+
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Implements hook_entity_view_alter().
+ */
+function oe_theme_entity_view_alter(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display) {
+  if (!isset($build['content_moderation_control'])) {
+    return;
+  }
+
+  // Ensure that the content moderation control block will always
+  // show at the top, above any other elements.
+  $build['content_moderation_control']['#weight'] = -500;
+}


dxvargas:
@upchuk weights are always relative. We always need to worry about what we have in other places.
Even if using -1000 someone can put in other place (also in `featured_image`) weights of -9999.
It's up to each one's imagination.

Why not, instead of going until -1000, just go until -20?
Like this, instead of escalating values here then there, we're tuning the values. We don't need to override a third party definition, we just need to adapt ours.

modules/oe_theme_content_news/oe_theme_content_news.module
```diff
@@ -27,7 +27,7 @@ function oe_theme_content_news_preprocess_node__oe_news(&$variables) {
   if ($media instanceof MediaInterface) {
     $builder = \Drupal::entityTypeManager()->getViewBuilder('media');
     $variables['content']['featured_image'] = $builder->view($media, 'oe_theme_main_content');
-    // Show the image always at the top.
+    // Show the image always at the top after content_moderation_control.
```
upchuk: No need for this change.

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -0,0 +1,25 @@
+<?php
+
+/**
+ * @file
+ * Module file additional functionality for theming.
```
upchuk:
```
The OE Theme Helper module.
```

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -0,0 +1,25 @@
...
+ * Implements hook_entity_view_alter().
+ *
+ * Change content_moderation_control's weight.
+ */
+function oe_theme_entity_view_alter(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display) {
+  if (!isset($build['content_moderation_control'])) {
+    return;
+  }
+
+  // Set weight minor to be always on top even before News' featured_image.
```
upchuk:
```
Ensure that the content moderation control block will always show at the top, above any other elements.
```

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -0,0 +1,25 @@
...
+
+declare(strict_types = 1);
+
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Implements hook_entity_view_alter().
+ *
+ * Change content_moderation_control's weight.
```
upchuk: You can remove this comment.

PR #223, opened 2019-05-10
OPENEUROPA-1865: Make sure that all default values of ECL is translatable.

sergepavle:
### Description

Given in our ambition to have our strings translated, the scope of this ticket is to look at all the patterns and templates that use ECL components and ensure that all the default strings inherited from ECL are translated (ran through t())


nagyad:
@sergepavle templates/form/form--search-form.html.twig:23 is missing translation. Please make sure that you go through the full templates/ folder.

templates/patterns/list_item/pattern-list-item--variant-thumbnail-secondary.html.twig
```diff
@@ -10,5 +10,6 @@
   'metas': meta,
   'href': url,
   'detail': detail,
-  'secondary_image': image
+  'secondary_image': image,
+  'screen_reader': 'List item't
```
nagyad: 'screen_reader': 'List item'|t

PR #229, opened 2019-05-15
OPENEUROPA-1864: Update dependencies.

sergepavle:
### Description

Upgrade dependencies on composer.


dxvargas (dismissed):

nagyad (changes requested):
In require-dev dependencies we are allowed to put dev-master since these set of dependencies are for development. Therefore I would not change these to release tags.

imanoleguskiza:
Outdated PR

PR #230, opened 2019-05-20
OPENEUROPA-1897: Removed drone permission fix.

imanoleguskiza:
### Description

Removed drone permission fix

### Change log

- Added:
- Changed: Removed drone permission fix
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -71,12 +71,6 @@ pipeline:
     group: test
     image: fpfis/httpd-php-full:7.1
     commands:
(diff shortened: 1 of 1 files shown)


voidtek:
In this PR, we should change all image: `fpfis/httpd-php-full:7.1` on .drone file to image: `fpfis/httpd-php-ci:7.1`

voidtek:
For the step `before-release`, we should use `ci` too (or `full`).
This change should be tested.

PR #232, opened 2019-05-28
OPENEUROPA-1927: Theming of Policy content type.

sergepavle:
### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_content_policy/oe_theme_content_policy.info.yml
```diff
@@ -0,0 +1,7 @@
+name: OpenEuropa Theme Policy Content
+type: module
+description: Companion module for the OE Content Policy module
+package: OpenEuropa
+core: 8.x
+dependencies:
+  - oe_content_policy:oe_content_policy
```
brummbar: We missed this in the other similar companion modules, but let's take the chance to fix it here: the correct format is
`project:module`, so this should be `oe_content:oe_content_policy`.

PR #233, opened 2019-06-06
OPENEUROPA-1948: Theming of Document media type.

sergepavle:
### Description

See: https://webgate.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-1944
### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


composer.json
```diff
@@ -30,6 +29,7 @@
         "openeuropa/drupal-core-require-dev": "^8.7",
         "openeuropa/oe_content": "dev-master",
         "openeuropa/oe_corporate_blocks": "~1.0",
+        "openeuropa/oe_media": "dev-OPENEUROPA-1946 as 1.0.0-beta3",
```
upchuk: Let's use dev-master like we did with oe_content (once OPENEUROPA-194 is merged). It's dev dependencies anyway.

tests/Kernel/MediaRenderTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+    $this->mediaStorage = $this->container->get('entity_type.manager')->getStorage('media');
+    $this->mediaViewBuilder = $this->container->get('entity_type.manager')->getViewBuilder('media');
+  }
+
+  /**
+   * Tests that the Document media type is rendered with the correct ECL markup.
+   */
+  public function testDocumentMedia(): void {
+    $file = file_save_data(file_get_contents(drupal_get_path('module', 'oe_media') . '/tests/fixtures/sample.pdf'), 'public://test.pdf');
```
upchuk: Shouldn't we test this with private file system?

sergepavle: Unfortunatly it is not easy. Can't use private file system inside KernelTests.
I tried but didn't find solution.

PR #238, opened 2019-06-14
OPENEUROPA-1977: Update @ecl/buttons for ecl2.

voidtek:
### Description

Update @ecl/buttons for ecl2.

### Change log

- Changed: Update @ecl/buttons for ecl2.


templates/patterns/button/button.ui_patterns.yml
```diff
@@ -2,9 +2,6 @@ button:
   label: "Button"
   description: "Buttons indicate actions the users can take."
   variants:
-    default:
-      label: "Default"
-      description: "The default button is used for important links on the page."
     primary:
```
ademarco: We cannot remove variants etc. since they end up in configuration here and there. UI patterns are essentially our API, the exercise is really trying to fit the new ECL component into the "old" pattern, without changing anything of the pattern definition. In case we will need to change it then we might need an upgrade path.

voidtek: ok :)

templates/patterns/button/button.ui_patterns.yml
```diff
@@ -23,30 +23,35 @@ button:
...
+      type: "text"
+      label: "Label"
+      description: "Label of the button."
+      preview: "Button"
     type_attribute:
       type: "text"
       label: "Type attribute"
       description: "The buttons type attribute"
       preview: "submit"
-    href:
```
ademarco: Just saw that in ECL buttons are only using the HTML element `button`. I've checked the documentation and, indeed, it seems that buttons are not supposed to link anything anymore. 

This needs to be escalated to the INNO team as, clearly, we will still have just links that are supposed to look as buttons but are simply links, like the "call to action" ones.

templates/patterns/button/button.ui_patterns.yml
```diff
@@ -23,30 +23,35 @@ button:
...
+      type: "array"
       label: "Icon"
-      description: "An icon to use with the button."
-      preview: "caret-right"
-    label:
+      description: "An icon."
+      preview:
+        - path: ''
+        - type: 'ui'
+        - name: 'corner-arrow'
```
ademarco: In the new ECL 2 icons are, indeed, hashes but for us it's a string, we will need to keep this a string in the pattern definition and turn it into an array "behind the scenes", in the Twig template, otherwise we will not be able to upgrade seamlessly anymore.

The UI Patterns are our API, we can only add things, not change it (unless really really necessary, which is not this case).

templates/patterns/button/button.ui_patterns.yml
```diff
@@ -23,30 +23,35 @@ button:
...
+    icon_position:
       type: "text"
-      label: "Label"
-      description: "Label of the button."
-      preview: "Button"
-    is_block:
```
ademarco: Same here, we cannot remove the `is_block` property here, we need to make the `is_block` behave correctly using the ECL 2, so within the Twig template, as we will do with the icon above.

templates/patterns/button/button.ui_patterns.yml
ademarco: The ECL 2 itself has an example of such call to action usage of a link, [check here](https://v2--europa-component-library.netlify.com/playground/ec/?path=/story/templates-pages--campaign):

image

That is currently a `<button>`, but there is no form or anything around it, so it's a dead-end:

image

imanoleguskiza:
The PR does what its needed, but from my side I would mark all the parameters and variants that are not supported as "deprecated". Since it is not a blocker I'll move it forward

templates/patterns/button/pattern-button.html.twig
```diff
@@ -4,11 +4,35 @@
...
-  'label': label,
-  'is_block': is_block
-} only %}
+{% if variant == 'default' %}
+  {% include '@ecl/link' with {
+    'link': {
+      'type': type_attribute,
+      'label': label,
+      'path': href,
+      'icon_position': icon_position
```
ademarco: Default it to `after`, if possible.

templates/patterns/button/button.ui_patterns.yml
```diff
@@ -38,7 +41,12 @@ button:
       type: "text"
       label: "Icon"
       description: "An icon to use with the button."
-      preview: "caret-right"
+      preview: "check"
+    icon_position:
+      type: "text"
+      label: "Icon position"
+      description: "Position of link icon (can be 'before' or 'after') if icon is available"
+      preview: 'before'
```
ademarco: Since this is a new property, let's default it to "after" if empty, in `pattern-button.html.twig`.

templates/patterns/button/pattern-button.html.twig
```diff
@@ -4,11 +4,35 @@
...
+    },
+    'icon': {
+      'path': '/dist/images/icons/sprites/icons.svg',
+      'type': 'ui',
+      'name': icon,
+      'size': 's',
+    },
+    'icon_position': icon_position,
```
ademarco: Default it to `after`, if possible.

templates/patterns/button/pattern-button.html.twig
```diff
@@ -4,11 +4,35 @@
...
+    'label': label,
+    'variant': variant|replace({'_':'-'}),
+    'type': type_attribute,
+    'icon': {
+      'path': '/dist/images/icons/sprites/icons.svg',
+      'type': 'ui',
+      'name': icon,
+      'size': 's',
+    },
+    'icon_position': icon_position,
```
ademarco: Default it to `after`, if possible.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -1,783 +1,94 @@
-- array:
```
ademarco: Instead of creating a `_ecl1` file let's keep one file and comment the tests that should not run. People will then un-comment relevant tests.

templates/patterns/button/pattern-button.html.twig
voidtek: In the twig template, the icon_position have already 'after' as default.
`{% set _icon_position = icon_position|default('after') %}`

But we could put here too.

templates/patterns/button/button.ui_patterns.yml
voidtek: ok.

templates/patterns/button/pattern-button.html.twig
voidtek: yes.

templates/patterns/button/pattern-button.html.twig
voidtek: with pleasure.

tests/Kernel/fixtures/rendering.yml
voidtek: ok, nice idea.

templates/patterns/button/pattern-button.html.twig
```diff
@@ -4,11 +4,35 @@
...
+{% if variant == 'default' %}
+  {% include '@ecl/link' with {
+    'link': {
+      'type': type_attribute,
+      'label': label,
+      'path': href,
+      'icon_position': icon_position|default('after')
+    },
+    'icon': {
+      'path': '/dist/images/icons/sprites/icons.svg',
```
ademarco: Let's move this into a global variable that will be always available in all templates, we can achieve that by implementing `function hook_preprocess(&$variables, $hook)`, something like:

```php
function oe_theme_preprocess(&$variables, $hook) {
  global $base_url;
  $theme = \Drupal::theme()->getActiveTheme();
  $variables['ecl_icon_path'] = $base_url.'/'. $theme->getPath() . '/dist/images/icons/sprites/icons.svg';
}
```
and then we simply do this everywhere:
```
       'path': ecl_icon_path,
```
Not that this path `'/dist/images/icons/sprites/icons.svg'` works by chance since the server root points at the root of the project in the theme, but it would not work on a site.

voidtek: done.

PR #239, opened 2019-06-18
OPENEUROPA-1973: ECL2: Update page-header and footer to ecl2.

voidtek:
### Description

Update these following component for ECL2:
- page-header
- footer


voidtek:
image

voidtek:
image

templates/navigation/breadcrumb.html.twig
```diff
@@ -10,8 +10,21 @@
...
+      'path' : breadcrumb_item.href
+    } %}
+    {% set _breadcrumb_links = _breadcrumb_links|merge([_links]) %}
+  {% endfor %}
+
+  {% include '@ecl/breadcrumb' with {
+    'links' : _breadcrumb_links,
+    'navigation_text' : '',
+    'ellipsis_label': '',
+    'icon_file_path' : '/dist/images/icons/sprites/icons.svg'
```
imanoleguskiza: Shouldn't we use the variables you set in the preprocess function?

templates/navigation/oe-corporate-blocks-footer.html.twig
```diff
@@ -10,7 +10,111 @@
  * @see templates_preprocess_oe_corporate_blocks_footer()
  */
 #}
-{% include '@ecl/ec-component-footer' with {
-  'corporate_footer': corporate_footer|default([]),
-  'custom_footer': custom_footer|default([]),
+{% set _back_to_top = {
```
imanoleguskiza: Question about this implementation: We have a lot of logic in the twig file, shouldn't we do this in the preprocess file?

templates/navigation/breadcrumb.html.twig
voidtek: I don't think, because this value is static and very specific of the theme.

voidtek: Ok I will change that.

templates/navigation/oe-corporate-blocks-footer.html.twig
voidtek: done.

templates/navigation/breadcrumb.html.twig
```diff
@@ -10,8 +10,21 @@
  * @see ./core/themes/stable/templates/navigation/breadcrumb.html.twig
  */
 #}
+
 {% if segments %}
```
imanoleguskiza: Because you got rid of the segments variable in the preprocess function, this code is never called. You also already format the links in the preprocess so there is no need for you to re-do it in the template.

voidtek: After investigation, we could remove this template.

templates/patterns/page_header/page_header.ui_patterns.yml
```diff
@@ -5,20 +5,20 @@ page_header:
     identity:
       type: "text"
       label: "Identity"
-      description: "Name of the site"
+      description: "Name of the site (deprecated)"
```
ademarco: Is this deprecated? We should double check it with @kugta

templates/patterns/page_header/page_header.ui_patterns.yml
```diff
@@ -47,3 +47,16 @@ page_header:
...
+      description: "A list of infos of the current page."
+      preview:
+        - text: "Monday 8 February"
+          icon_type: "general"
+          icon_name: "calendar"
+          icon_path: "/dist/images/icons/sprites/icons.svg"
+        - text: "Munich, Germany"
+          icon_type: "general"
+          icon_name: "location"
+          icon_path: "/dist/images/icons/sprites/icons.svg"
```
ademarco: We should hide all icon implementation details (such as type, name and path) in the pattern template since components using this pattern should not know how to build an icon.

I suggest using a generic `type` attribute instead and then, depending from the `type` we will map it to the correct icon, so something like:

```
    infos:
      type: "array"
      label: "Infos"
      description: "A list of infos of the current page."
      preview:
        - text: "Monday 8 February"
          type: "date"
        - text: "Munich, Germany"
          type: "location"
```

templates/patterns/page_header/pattern-page-header.html.twig
```diff
@@ -4,11 +4,47 @@
...
+  {% set _breadcrumb_links = _breadcrumb_links|merge([_links]) %}
+{% endfor %}
+{% set _breadcrumb = {
+  'links' : _breadcrumb_links,
+  'navigation_text' : '',
+  'ellipsis_label': '',
+  'icon_file_path' : ecl_icon_path
+} %}
+
+{% set _infos = {} %}
```
ademarco: See comment above, we should derive the info below from the info type.

oe_theme.theme
```diff
@@ -33,19 +33,24 @@ function oe_theme_preprocess_breadcrumb(array &$variables): void {
...
   if (!empty($title)) {
-    $variables['segments'][] = [
+    $variables['links'][] = [
       'label' => $title,
     ];
   }
+
+  $variables['navigation_text'] = '';
+  $variables['ellipsis_label'] = '';
+  $variables['icon_file_path'] = $variables['ecl_icon_path'];
```
ademarco: Why not just use `ecl_icon_path`?

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
```diff
@@ -110,6 +110,7 @@ public function build(): array {
       '#title' => $title,
       '#introduction' => $metadata['introduction'] ?? '',
       '#metas' => $metadata['metas'] ?? [],
+      '#infos' => $metadata['infos'] ?? [],
```
ademarco: The `infos` part was not originally supported by `@PageHeaderMetadata`, or rather: it's not handled but technically we can support it.

I suggest adding a documentation file here `./docs` where we explain how the metadata plugin works and what are the supported elements that implementations can fill in, such as `intriduction`, `metas` and `infos`.

templates/patterns/page_header/page_header.ui_patterns.yml
voidtek: Is not deprecated (I will change that) and one ticket is already created to follow this part.

templates/patterns/page_header/page_header.ui_patterns.yml
voidtek: Good catch. I will change too.

oe_theme.theme
voidtek: This preprocess goes directly to @ecl/component-breadcumb.
We don't have another twig template to map the values.

templates/patterns/page_header/pattern-page-header.html.twig
voidtek: We don't need to change the name of this variable.
see https://ecl-twig.netlify.com/ec/?path=/story/components-page-header--title
```
breadcrumb: {  
    links: [    
      {  
        label: 'Link 1',  
        path: '/example'  
      },  
    ],  
    icon_file_path: '/path-to-the-icons-file',  
    navigation_text: 'You are here:',  }, 
```

voidtek: Here, it's not an icon component, only a param of the breadcrumb.

ademarco: Sure, sorry I meant something else! :D Just put the comment on the wrong line.

modules/oe_theme_helper/src/Plugin/Block/PageHeaderBlock.php
voidtek: I created an issue https://github.com/openeuropa/oe_theme/issues/250

PR #240, opened 2019-06-19
OPENEUROPA-1974: ECL 2.0: Multilingual components.

sergepavle:
### Description

Add the following components:

- Splash page language switcher
- Pop-up language switcher

Fix language on logo;
Add the new way to init js component.


voidtek:
This rendering test is missing to be tested:  
```- array:
    '#type': pattern
    '#id': button
    '#variant': splash_page
    '#fields':
      label: "Button label"
  assertions:
    equals:
      'button.ecl-button--splash-page': "Button label"```

**Comment by voidtek** [comment]
We should add the behat test 'Scenario: The language switcher dialog can be accessed' or create a follow-up ticket.

File: tests/Kernel/LanguageSwitcherTest.php [hunk 1]
```diff
@@ -55,6 +52,8 @@ public function testLanguageSwitcherLinkListRendering(): void {
    * @param string $lang_prefix
    *   The language prefix.
    *
+   * @todo Adjust this function after implementation of Language switcher in page header.
```
voidtek: Can you provide the ticket number?

voidtek (changes requested):

tests/Kernel/LanguageSwitcherTest.php
```diff
@@ -39,11 +39,8 @@ public function testLanguageSwitcherLinkListRendering(): void {
       // Get the language prefix.
       $lang_prefix = $lang_config->get('url.prefixes.' . $id);
 
-      $actual = $crawler->filter(".ecl-dialog a.ecl-language-list__button[lang={$lang_prefix}]")->text();
-      $this->assertEquals($name, trim($actual));
-
-      $actual = $crawler->filter(".ecl-dialog a.ecl-language-list__button[hreflang={$id}]")->text();
```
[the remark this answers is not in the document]
sergepavle: This part from scope of other ticket related to site header and should be updated when this block (language switcher) gonna be adjusted to site header.

templates/compositions/ec-component-navigation/navigation-list.html.twig
```diff
@@ -0,0 +1,53 @@
+{% spaceless %}
```
imanoleguskiza: Why is this template part of this PR? It doesn;t seem to involve any of the multilingual components at all? At the same time, the navigation list composition described in the ECL is used to navigate a page, not to add shortcuts to different functionalities so I'm not sure its the best composition to use.

imanoleguskiza (changes requested):

templates/compositions/ec-component-navigation/navigation-list.html.twig
voidtek: I added this template to unblock us when we want to use the link 'edit' of a node (used in Behat tests) or others links. I didn't see any ticket concerning this requirement.
It's not a design component, it's a pure drupal menu. I tried to put it very simple, only to unblock us. But I can remove this template and wait for ecl-twig to create one, as you wish.

ademarco (changes requested):
I've tested this locally but I get the following error, with and without aggregation:

image

templates/compositions/ec-component-navigation/navigation-list.html.twig
voidtek: Removed and PR created https://github.com/openeuropa/oe_theme/pull/319

oe_theme.theme
```diff
@@ -599,11 +611,16 @@ function oe_theme_preprocess_links__oe_multilingual_content_language_block(array
       ->setAbsolute(TRUE)
       ->toString();
 
+    // Get the language prefix.
+    $lang_prefix = \Drupal::configFactory()
+      ->get('language.negotiation')
+      ->get('url.prefixes.' . $language_code);
```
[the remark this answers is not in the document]
voidtek: Done.

oe_theme.theme
```diff
@@ -599,11 +617,16 @@ function oe_theme_preprocess_links__oe_multilingual_content_language_block(array
       ->setAbsolute(TRUE)
       ->toString();
 
+    // Get the language prefix.
+    $lang_prefix = \Drupal::configFactory()
```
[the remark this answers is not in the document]
voidtek: Changed to use the helper and the twig filter.

oe_theme.theme
```diff
@@ -997,8 +1015,16 @@ function oe_theme_preprocess_block(&$variables) {
...
-  $lang_id = \Drupal::languageManager()->getCurrentLanguage()->getId();
-  $variables['site_logo'] = $variables['ecl_logo_path'] . '/logo--' . $lang_id . '.svg';
+  $current_language_id = \Drupal::languageManager()
+    ->getCurrentLanguage()
+    ->getId();
+
+  // Get the language prefix.
+  $lang_prefix = \Drupal::configFactory()
+    ->get('language.negotiation')
+    ->get('url.prefixes.' . $current_language_id);
```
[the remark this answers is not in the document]
voidtek: done.

tests/Kernel/LanguageSwitcherTest.php
```diff
@@ -21,31 +19,27 @@ public function testLanguageSwitcherLinkListRendering(): void {
...
-      $this->assertEquals($name, trim($actual));
+      $lang_id = $language->getId();
+      $lang_prefix = $lang_config->get('url.prefixes.' . $lang_id);
+
+      $actual = $crawler->filter(".ecl-language-list--overlay a.ecl-language-list__link[lang={$lang_prefix}]")->text();
+      $this->assertEquals(
+        $languages[$lang_id]->getName(),
+        preg_replace('/[\x00-\x1F\x7F\xA0]/u', '', $actual)
```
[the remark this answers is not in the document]
voidtek: This code is to remove unprintable characters, included by the ::before of ::after icons. I will add a comment.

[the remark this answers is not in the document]
ademarco: :+1:

oe_theme.theme
```diff
@@ -557,18 +559,24 @@ function oe_theme_preprocess_links__language_block(array &$variables): void {
       ->setAbsolute(TRUE)
       ->toString();
 
-    // Get the language prefix.
-    $lang_prefix = \Drupal::configFactory()
-      ->get('language.negotiation')
-      ->get('url.prefixes.' . $language_code);
+    $eu_lang_list = Filters::getEuropeanUnionLanguageList();
+    $language_prefix = !empty($eu_lang_list[$language_code][2]) ? $eu_lang_list[$language_code][2] : $language_code;
```
ademarco: Maybe we can take out the static method `getEuropeanUnionLanguageList` from the filter and put it in its own class, so that we can add simple helper methods to get lang code etc. and hide all the implementation details in there. @voidtek is it ok if I do that?

ademarco (dismissed):

modules/oe_theme_helper/src/TwigExtension/Filters.php
```diff
@@ -73,14 +75,35 @@ public function toNativeLanguageName($language_code): string {
...
+   * @throws \InvalidArgumentException
+   *   Thrown when the passed in language code does not exist.
+   */
+  public function toInternalLanguageId($language_code): string {
+    // The fallback implemented in case we don't have enabled language.
+    if (EuropeanUnionLanguages::hasLanguage($language_code)) {
+      return EuropeanUnionLanguages::getInternalLanguageCode($language_code);
+    }
+
+    throw new \InvalidArgumentException('The language code ' . $language_code . ' does not exist.');
```
imanoleguskiza: This exception is misleading. The language code is not in the European Union language list but it could have been enabled in the site (lets say chinese for example). I think we should check for the internal code language in the language list but return the language code if it is not found. (I know this is an scenario that will probably never occur and atm we only use this method to pick the appropriate logo, but we can rule out it could be used somewhere else)

modules/oe_theme_helper/src/EuropeanUnionLanguages.php
```diff
@@ -0,0 +1,118 @@
...
+   * @return array
+   *   An array with language codes as keys, and English and native language
+   *   names as values.
+   */
+  public static function getLanguageList(): array {
+    return self::$languages;
+  }
+
+  /**
+   * Get a language ID given its code.
```
imanoleguskiza: It should probably be "Assert whether the language information is available."

imanoleguskiza (changes requested):

modules/oe_theme_helper/src/TwigExtension/Filters.php
ademarco: That's a good point, I'll change that.

imanoleguskiza (approved):

voidtek (approved):

PR #241, opened 2019-06-20
OPENEUROPA-2014: Theming of publication content type.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_theme_content_publication/oe_theme_content_publication.libraries.yml
+++ b/modules/oe_theme_content_publication/oe_theme_content_publication.libraries.yml
@@ -1,4 +1,4 @@
-publication_css:
+oe_publication:
   version: VERSION
   css:
     theme:
--- a/modules/oe_theme_content_publication/oe_theme_content_publication.libraries.yml
+++ b/modules/oe_theme_content_publication/oe_theme_content_publication.libraries.yml
@@ -1,0 +1,5 @@
+publication_css:
+  version: VERSION
+  css:
+    theme:
+      css/publication.css: {}
--- a/modules/oe_theme_content_publication/oe_theme_content_publication.module
+++ b/modules/oe_theme_content_publication/oe_theme_content_publication.module
@@ -1,0 +1,19 @@
+<?php
+
+/**
+ * @file
+ * Module file used for theming the Publication content type.
+ */
+
+declare(strict_types = 1);
+
+use Drupal\Component\Utility\NestedArray;
+
+/**
+ * Implements template_preprocess_node() for the Publication node type.
+ */
+function oe_theme_content_news_preprocess_node__oe_publication__full(&$variables) {
+  if (NestedArray::keyExists($variables, ['content', 'oe_documents'])) {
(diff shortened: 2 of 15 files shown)


tests/Kernel/ContentRenderTest.php
```diff
@@ -60,24 +67,41 @@ protected function setUp() {
...
       'oe_content_page',
       'oe_content_policy',
+      'oe_content_publication',
       'oe_theme_content_news',
       'oe_theme_content_page',
       'oe_theme_content_policy',
+      'oe_theme_content_publication',
     ]);
 
+    Role::load(RoleInterface::ANONYMOUS_ID)
```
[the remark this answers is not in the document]
sergepavle: I had problem with output of File field during rendering Publication content type. Reason was in checking permissions to document media type.

modules/oe_theme_content_publication/oe_theme_content_publication.libraries.yml
```diff
@@ -0,0 +1,5 @@
+publication_css:
```
upchuk: Let's not call the library `publication_css` just because for now it only handles css. Let's name it: `oe_publication` like the content type.

modules/oe_theme_content_publication/oe_theme_content_publication.module
```diff
@@ -0,0 +1,19 @@
...
+
+declare(strict_types = 1);
+
+use Drupal\Component\Utility\NestedArray;
+
+/**
+ * Implements template_preprocess_node() for the Publication node type.
+ */
+function oe_theme_content_news_preprocess_node__oe_publication__full(&$variables) {
+  if (NestedArray::keyExists($variables, ['content', 'oe_documents'])) {
```
upchuk: Since:

1. We are renaming the library to be more generic and
2. It's configured that the full view mode of the publication will show the documents,

This `if` conditional is not needed. You can just attach the library to the top level `$variables`.

voidtek:
We could add the following step on drone to understand better the issue 
```
  debug:
    image: fpfis/httpd-php-ci:7.1
    commands:
      - ./vendor/bin/drush ws --count 500
    when:
      status: failure
```

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/EntityCanonicalRoutePage.php
```diff
@@ -18,8 +18,9 @@
  *   id = "entity_canonical_route",
  *   label = @Translation("Default entity metadata extractor")
  * )
- *
- * @deprecated in 1.x, will be removed in 2.0.
+ * phpcs:disable Drupal.Commenting.Deprecated
```
voidtek: As discussed, the following solution is recommended.
https://github.com/openeuropa/oe_theme/blob/2.x/modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/EntityCanonicalRoutePage.php#L22

So you can rebase the PR and remove this change.

PR #248, opened 2019-06-25
OPENEUROPA-1981: ECL2: Update the component pattern file to ECL2.

voidtek:
### Description

Update the component pattern file.

### Change log

- Added:
- Changed: Pattern file with and without translation.

### DIFF
--- a/templates/patterns/file_image/pattern-file-image.html.twig
+++ b/templates/patterns/file_image/pattern-file-image.html.twig
@@ -4,9 +4,8 @@
  * File image component.
  */
 #}
-{% include '@ecl/ec-component-file' with {
-  'variant': 'image',
-  'src': src,
+{% include '@ecl/media-container' with {
+  'image': src,
   'alt': alt,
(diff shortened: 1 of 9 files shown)


voidtek:
The File component pattern:
image

The File with translations component pattern:
image

The file image component pattern:
image

voidtek:
The file video component pattern:

image

templates/patterns/file_image/pattern-file-image.html.twig
```diff
@@ -4,9 +4,8 @@
  * File image component.
  */
 #}
-{% include '@ecl/ec-component-file' with {
-  'variant': 'image',
-  'src': src,
+{% include '@ecl/media-container' with {
```
sergepavle: In twig template I see that this component support tracks. But in fact in documentation of ecl2 information is not available. Did you have already discussion with ecl/ecl-twig teams regarding not update documentation?
If this track element required - probably it should be added inside pattern.

PR #254, opened 2019-06-28
OPENEUROPA-2025: Use custom image effect to prevent retina image reducction.

imanoleguskiza:
### Description

When the images we use are too small, the same image is used for both regular and retina displays. This means that on retina displays the image gets displayed at half size. 
Fix the issue by upscaling the image if it is smaller than the desired  scale.

### Change log

- Added: Custom retina scale effect
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_helper/config/schema/oe_theme_helper.schema.yml
```diff
@@ -0,0 +1,10 @@
+image.effect.retina_image_scale:
+  type: image_size
```
upchuk: Since you are extending from the scale image effect, you can just call it type `image.effect.image_scale` and inherit its properties. And add your own `multiplier` property to the definition.

modules/oe_theme_helper/templates/retina-image-scale-summary.html.twig
```diff
@@ -0,0 +1,37 @@
...
+ *   - upscale: If images larger than their original size can scale.
+ * - effect: The effect information, including:
+ *   - id: The effect identifier.
+ *   - label: The effect name.
+ *   - description: The effect description.
+ *
+ * @ingroup themeable
+ */
+#}
+{% if data.width and data.height -%}
```
upchuk: Why is this needed? You copy pasted this entire thing and made no changes whatsoever..

tests/Kernel/BlockTest.php
```diff
@@ -30,6 +30,7 @@ class BlockTest extends EntityKernelTestBase {
     'image',
     'breakpoint',
     'responsive_image',
+    'oe_theme_helper',
```
upchuk: Why?

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,120 @@
...
+
+namespace Drupal\oe_theme_helper\Plugin\ImageEffect;
+
+use Drupal\Component\Utility\Image;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Image\ImageInterface;
+use Drupal\image\Plugin\ImageEffect\ScaleImageEffect;
+
+/**
+ * Scales an image resource.
```
upchuk: Can you please please stop copy pasting comments and things if you don't update them to match what you are doing?

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,120 @@
...
+class RetinaScaleImageEffect extends ScaleImageEffect {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function applyEffect(ImageInterface $image) {
+    $target_with = $this->configuration['width'];
+    $target_height = $this->configuration['height'];
+    $upscale = $this->configuration['upscale'];
+    if (!$upscale) {
```
upchuk: You can inline these. It's very difficult to read. Had to move them to one line to understand what is going on...

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,120 @@
...
+      if (
+      (!empty($target_with) && $this->configuration['width'] > $image->getWidth()) ||
```
upchuk: If you put `$this->configuration['width']` into a variable, why you checking it here like this again?

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,120 @@
...
+      (!empty($target_height) && $this->configuration['height'] > $image->getHeight())
+      ) {
+        $target_with = $image->getWidth() * $this->configuration['multiplier'];
+        $target_height = $image->getHeight() * $this->configuration['multiplier'];
+        $upscale = TRUE;
```
upchuk: I don't think we should override `$upscale` unless there is also a multiplier higher than 1 right?

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,120 @@
...
+      ]);
+      return FALSE;
+    }
+    return TRUE;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function transformDimensions(array &$dimensions, $uri) {
```
upchuk: See comments from `applyEffect ` for here as well

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,120 @@
...
+      Image::scaleDimensions($dimensions, $target_with, $target_height, $upscale);
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getSummary() {
+    $summary = [
+      '#theme' => 'retina_image_scale_summary',
```
upchuk: See other comment.

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form = parent::buildConfigurationForm($form, $form_state);
+    $form['multiplier'] = [
+      '#type' => 'number',
+      '#title' => t('Multiplier'),
+      '#default_value' => $this->configuration['multiplier'],
```
upchuk: Let's put a description to say what this does and that it overrides the upscale etc.

modules/oe_theme_helper/config/schema/oe_theme_helper.schema.yml
```diff
@@ -0,0 +1,10 @@
...
+  label: 'Retina Image scale'
```
upchuk: Also let's not call it "Retina" cause it has nothing to do with retina per se. It's a multiplier based. If I configure a multiplier of 10, it has nothing to do with retina. So I'd call the plugin `MultiplierImageScale`. Though it's also only upscales if the image is smaller than the configuration. Any ideas for naming?

modules/oe_theme_helper/config/schema/oe_theme_helper.schema.yml
imanoleguskiza: I am extending the scale image effect but I'm also creating a new plugin with a separate id. Correct me if I'm wrong but if I use the "image.effect.image_scale" type then the multiplier property would apply to configuration using the "image_scale" id.

modules/oe_theme_helper/templates/retina-image-scale-summary.html.twig
imanoleguskiza: The twig template is needed in because the image style preview requires it. Since the changes we made to the base scale image style are a very edge case I didn't think we would need to change anything on the summary template either.

upchuk: Then why not use the original?!

modules/oe_theme_helper/config/schema/oe_theme_helper.schema.yml
upchuk: config schema "types" extend one another. So `image.effect.image_scale` extends `image_size` and adds its own config fields. You can extend from `image.effect.image_scale` and add yours (the multipler0

tests/Kernel/BlockTest.php
imanoleguskiza: When installing the theme we install all the image styles. Most of the schema for this image styles comes from the image module (already there) but the new retine_image_scale is defined on the oe_theme_helper, hence the need to enable it as well.

modules/oe_theme_helper/config/schema/oe_theme_helper.schema.yml
imanoleguskiza: We could change the configuration form so that the users are forced to select certain multipliers. Currently only 2x and 3x retina screens are available (after a quick google search, let me know if there are more options) so we could restrict the field to those two options. Would this make sense?

upchuk: Yes, true, makes sense what you say.

modules/oe_theme_helper/templates/retina-image-scale-summary.html.twig
imanoleguskiza: Fair enough, I'll re-use the original template.

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
imanoleguskiza: I updated the comments on the plugin to more clearly reflect what we are doing.

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
imanoleguskiza: Now that we are restricting the multiplier values we can keep this as it is.

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
imanoleguskiza: Added a description.

modules/oe_theme_helper/config/schema/oe_theme_helper.schema.yml
```diff
@@ -0,0 +1,7 @@
+image.effect.retina_image_scale:
+  type: image.effect.image_scale
+  label: 'Retina Image scale'
+  mapping:
+    multiplier:
+      type: integer
+      label: 'Multiplier'
```
dxvargas: Needs a new line

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,132 @@
...
+    if ($dimensions['width'] && $dimensions['height']) {
+      $target_with = $this->configuration['width'];
+      $target_height = $this->configuration['height'];
+      // If we are not upscaling the image, check to see if it's smaller
+      // than the defined dimensions.
+      if (!$upscale = $this->configuration['upscale']) {
+        if (
+          (!empty($target_with) && $this->configuration['width'] > $dimensions['width']) ||
+          (!empty($target_height) && $this->configuration['height'] > $dimensions['height'])
+        ) {
```
dxvargas: In these conditions it would be better to use:
`$target_with` instead of `$this->configuration['width']`
`$target_height` instead of `$this->configuration['height']`

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
upchuk: @imanoleguskiza You didn't do this.

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form = parent::buildConfigurationForm($form, $form_state);
+    $form['multiplier'] = [
+      '#type' => 'select',
+      '#title' => t('Multiplier'),
```
upchuk: `$this->t()` everywhere, not `t()`.

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+use Drupal\FunctionalTests\Image\ToolkitTestBase;
+
+/**
+ * Tests that the Retina Scale effect upscales images appropriately.
+ *
+ * @group image
+ */
+class RetinaScaleEffectTest extends ToolkitTestBase {
+  /**
```
upchuk: Empty line and `inheritdoc`

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+declare(strict_types = 1);
```
upchuk: Missing return types on all methods.

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Asserts the effect processing of an image effect plugin.
+   *
+   * @param string $effect_name
+   *   The name of the image effect to test.
+   * @param array $data
+   *   The data to pass to the image effect.
+   *
+   * @return bool
```
upchuk: Not a bool

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,88 @@
...
+namespace Drupal\Tests\oe_theme_helper\Functional;
+
+use Drupal\FunctionalTests\Image\ToolkitTestBase;
+
+/**
+ * Tests that the Retina Scale effect upscales images appropriately.
+ *
+ * @group image
+ */
+class RetinaScaleEffectTest extends ToolkitTestBase {
```
upchuk: You are only testing the multiplier. What if the image is is bigger than the config. Or if the upscale is checked?

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
imanoleguskiza: Done.

modules/oe_theme_helper/config/schema/oe_theme_helper.schema.yml
imanoleguskiza: Done.

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
imanoleguskiza: Done.

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
imanoleguskiza: Done

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
imanoleguskiza: Done

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
imanoleguskiza: Tests added

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+    $this->manager = $this->container->get('plugin.manager.image.effect');
+  }
+
+  /**
+   * Test the retina_image_scale_effect() function.
```
upchuk: What function is this?

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function testRetinaScaleEffect(): void {
+    $this->assertImageEffect('retina_image_scale', [
+      // Set the desired width to be much higher than the image width.
```
upchuk: `. By default, it uses a multiplier of 2.`

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+      'width' => $this->image->getWidth() * 10,
+    ]);
+    $this->assertToolkitOperationsCalled(['scale']);
+
+    // Check the parameters.
```
upchuk: Can remove this comment.

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+    $calls = $this->imageTestGetAllCalls();
+    $this->assertEqual($calls['scale'][0][0], $this->image->getWidth() * 2, 'Width is double the original size.');
+  }
+
+  /**
+   * Test the image_scale_effect() function using a multiplier of 3.
```
upchuk: What function?

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function testTripleMultiplierRetinaScaleEffect(): void {
+    $this->assertImageEffect('retina_image_scale', [
+      // Set the desired width to be much higher than the image width.
+      'width' => $this->image->getWidth() * 10,
+      // Set a multiplier of three for the test.
```
upchuk: No need for this comment.

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+      'multiplier' => 3,
+    ]);
+    $this->assertToolkitOperationsCalled(['scale']);
+
+    // Check the parameters.
```
upchuk: Remove comment

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+    $calls = $this->imageTestGetAllCalls();
+    $this->assertEqual($calls['scale'][0][0], $this->image->getWidth() * 3, 'Width is triple the original size.');
+  }
+
+  /**
+   * Test the image_scale_effect() function using a big enough image.
```
upchuk: What function?

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function testScaleEffect(): void {
+    $this->assertImageEffect('image_scale', [
+      // Set the image width to be smaller than the image width.
```
upchuk: Comment sounds weird. Maybe keep it in line with the same comment in the previous methods?

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+    // Check the parameters.
+    $calls = $this->imageTestGetAllCalls();
+    $this->assertEqual($calls['scale'][0][0], $this->image->getWidth() * 3, 'Width is triple the original size.');
+  }
+
+  /**
+   * Test the image_scale_effect() function using a big enough image.
+   */
+  public function testScaleEffect(): void {
+    $this->assertImageEffect('image_scale', [
```
upchuk: Are we testing the `image_scale` effect or the `retina_image_scale` effect?

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+    $this->assertToolkitOperationsCalled(['scale']);
+
+    // Check the parameters.
+    $calls = $this->imageTestGetAllCalls();
+    $this->assertEqual($calls['scale'][0][0], 10, 'Width was passed correctly');
+    $this->assertEqual($calls['scale'][0][1], 10, 'Height was based off aspect ratio and passed correctly');
+  }
+
+  /**
+   * Test the image_scale_effect() function using upscaling.
```
upchuk: Again, copy paste from core I see...

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function testScaleEffectWithUpscaling(): void {
+    $this->assertImageEffect('image_scale', [
```
upchuk: What are we testing here?

modules/oe_theme_helper/src/Plugin/ImageEffect/RetinaScaleImageEffect.php
```diff
@@ -0,0 +1,128 @@
...
+    $form['multiplier'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Multiplier'),
+      '#options' => [
+        2 => '2x',
+        3 => '3x',
+      ],
+      '#default_value' => $this->configuration['multiplier'],
+      '#required' => TRUE,
+      '#description' => $this->t("The image will be upscaled according to this multiplier if it is smaller than the dimensions defined in the 'Width' and 'Height' properties."),
```
upchuk: Single quotes.

Also we should mention that it only gets *upscaled* if the `upscale` is not unchecked no?

modules/oe_theme_helper/tests/src/Functional/RetinaScaleEffectTest.php
```diff
@@ -0,0 +1,115 @@
...
+   * The image effect manager.
+   *
+   * @var \Drupal\image\ImageEffectManager
+   */
+  protected $manager;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
```
upchuk: `setUp` is not ours so we shouldn't really put the void.

PR #256, opened 2019-07-01
OPENEUROPA-1984: ECL2: Add the contextual navigation component.

voidtek:
### Description

Add the contextual navigation component.

### Change log

- Changed: Update contextual navigation to ECL2.


voidtek:
I rebase the PR with @ecl2.7, but the component doesn't seem to be visually finished but all tests are green.

image

PR #265, opened 2019-07-10
 OPENEUROPA-2034: ECL2: Change Accordion to Accordion2 component.

voidtek:
### Description

Change Accordion to Accordion2 component.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -238,7 +238,6 @@ function oe_theme_preprocess_paragraph__oe_accordion(array &$variables): void {
     $paragraph = $field_item->entity;
     $variables['items'][] = [
       'title' => $builder->viewField($paragraph->get('field_oe_text')),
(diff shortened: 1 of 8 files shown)


oe_theme.theme
```diff
@@ -661,9 +660,6 @@ function oe_theme_preprocess_pattern_accordion(array &$variables): void {
         'label' => $item['title'],
         'icon' => [
```
brummbar: We should drop completely the icon parameter here.

voidtek: We should keep the icon for each item, to specify the icon path for the ecl.

brummbar: Yeah true it's coming from our code... Ignore :)

PR #266, opened 2019-07-10
OPENEUROPA-2046: Implementation of Event dispatching and event subscriber.

sergepavle:
### Description

The PageHeaderMetadata plugins are used in the theme to do essentially return the metadata to be used in the page header. However, it's role can be split into two:

Retrieve the data object which contains what needs to render (entity usually)
Build the metadata array based on that
We should separate the two concerns so that the data retrieval can be delegated also to external clients. For example, the current node should also be retrievable from custom routes and this should not have to be hardcoded in the NodeViewRoutesBase base class.

To start with:

We should dispatch an event from the NodeViewRoutesBase::getNode() to get the node
Implement a subscriber in the same module that does what the getNode() method does now.
Kernel or Functional test (whichever makes sense) that tests the dispatching logic.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
```diff
@@ -51,19 +37,13 @@ abstract class NodeViewRoutesBase extends PageHeaderMetadataPluginBase implement
...
-   *   The current route match.
-   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
-   *   The entity type manager.
-   * @param \Drupal\Core\Entity\EntityRepositoryInterface $entity_repository
-   *   The entity repository.
+   * @param \Symfony\Component\EventDispatcher\EventDispatcherInterface $event_dispatcher
+   *   The event dispatcher.
    */
-  public function __construct(array $configuration, string $plugin_id, $plugin_definition, RouteMatchInterface $current_route_match, EntityTypeManagerInterface $entity_type_manager, EntityRepositoryInterface $entity_repository) {
+  public function __construct(array $configuration, string $plugin_id, $plugin_definition, EventDispatcherInterface $event_dispatcher) {
```
brummbar: Unfortunately this break backwards compatibility. There might be websites using this plugin as base class. We need to leave in place the already injected services. In scope of this ticket, let's create a PR for 2.x where we remove both the deprecated plugin and the unneeded services.

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
```diff
@@ -104,25 +82,11 @@ public function getMetadata(): array {
...
-      return NULL;
-    }
-
-    if ($this->currentRouteMatch->getRouteName() === 'entity.node.revision') {
-      $node_revision = $this->currentRouteMatch->getParameter('node_revision');
-      $node = $this->entityTypeManager->getStorage('node')->loadRevision($node_revision);
 
-      return $node ? $this->entityRepository->getTranslationFromContext($node) : NULL;
-    }
+    $metadata_source_retrieve_event = new MetadataSourceRetrieveEvent();
```
brummbar:
```suggestion
    $event = new MetadataSourceRetrieveEvent();
```

modules/oe_theme_helper/src/MetadataEvents.php
```diff
@@ -0,0 +1,21 @@
...
+final class MetadataEvents {
+
+  /**
+   * Name of the event fired during collecting info for metadata building.
+   *
+   * @Event
+   *
+   * @var string
+   */
+  const COLLECT_ENTITY = 'oe_theme_helper.metadata_collecting_entity_data';
```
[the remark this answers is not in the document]
sergepavle: Current implementation of current request done in more agnostic way. So this event just collect information from all event subscribers regarding entities which could be used for building metadata.
So for now we use only nodes for building metadata but don't have assurance that we will not use other entity types in future.

modules/oe_theme_helper/src/EventSubscriber/MetadataEventSubscriber.php
```diff
@@ -0,0 +1,107 @@
...
+    $this->entityRepository = $entity_repository;
+  }
+
+  /**
+   * Metadata source retrieving event handler.
+   *
+   * @param \Drupal\oe_theme_helper\Event\MetadataSourceRetrieveEvent $event
+   *   Metadata source retrieving event.
+   */
+  public function onMetadataSourceRetrieving(MetadataSourceRetrieveEvent $event): void {
```
upchuk: getNode

modules/oe_theme_helper/src/MetadataEvents.php
upchuk:
```
const NODE = 'metadata_events.node';
```

modules/oe_theme_helper/src/Event/NodeMetadataEvent.php
```diff
@@ -0,0 +1,45 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Event;
+
+use Drupal\Core\Cache\RefinableCacheableDependencyTrait;
+use Drupal\node\NodeInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * Node retriever class for events related to metadata.
```
upchuk: Event used for retrieving the node entity for the page header metadata.

modules/oe_theme_helper/src/Event/NodeMetadataEvent.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+class NodeMetadataEvent extends Event {
+
+  use RefinableCacheableDependencyTrait;
+
+  /**
+   * The node object for passing to metadata building.
```
upchuk: The node entity.

modules/oe_theme_helper/src/Event/NodeMetadataEvent.php
```diff
@@ -0,0 +1,45 @@
...
+   *
+   * @var \Drupal\node\NodeInterface|null
+   */
+  protected $node = NULL;
+
+  /**
+   * Get Node object for metadata.
```
upchuk: Returns the node entity.

modules/oe_theme_helper/src/Event/NodeMetadataEvent.php
```diff
@@ -0,0 +1,45 @@
...
+   *
+   * @return \Drupal\node\NodeInterface|null
+   *   The node entity if applicable, null otherwise.
+   */
+  public function getNode(): ?NodeInterface {
+    return $this->node;
+  }
+
+  /**
+   * Set the node entity for metadata source.
```
upchuk: Sets the node entity.

modules/oe_theme_helper/src/Event/NodeMetadataEvent.php
```diff
@@ -0,0 +1,45 @@
...
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The node entity for metadata.
```
upchuk: The node entity.

modules/oe_theme_helper/src/EventSubscriber/NodeMetadataEventSubscriber.php
```diff
@@ -0,0 +1,107 @@
...
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\node\NodeInterface;
+use Drupal\oe_theme_helper\Event\NodeMetadataEvent;
+use Drupal\oe_theme_helper\MetadataEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Subscribes to the event fired on try to retrieve node for metadata.
+ */
+class NodeMetadataEventSubscriber implements EventSubscriberInterface {
```
upchuk: Actually, thinking about it, let's call this one `DefaultNodeMetadataEventSubscriber` because it is the default implementation. Meaning that it returns the node for the default use cases that the theme actually ships with.

modules/oe_theme_helper/src/EventSubscriber/NodeMetadataEventSubscriber.php
```diff
@@ -0,0 +1,107 @@
...
+use Drupal\Core\Entity\EntityRepositoryInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\node\NodeInterface;
+use Drupal\oe_theme_helper\Event\NodeMetadataEvent;
+use Drupal\oe_theme_helper\MetadataEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Subscribes to the event fired on try to retrieve node for metadata.
```
upchuk:
```
Provides the node entity on the default node-related routes such as canonical and revision.
@see \Drupal\oe_theme_helper\Plugin\PageHeaderMetadata\NodeViewRoutesBase
```

modules/oe_theme_helper/src/EventSubscriber/NodeMetadataEventSubscriber.php
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * The entity repository.
+   *
+   * @var \Drupal\Core\Entity\EntityRepositoryInterface
+   */
+  protected $entityRepository;
+
+  /**
+   * Creates a new MetadataEventSubscriber object.
```
upchuk: `Creates a new DefaultNodeMetadataEventSubscriber  object.`

modules/oe_theme_helper/src/EventSubscriber/NodeMetadataEventSubscriber.php
```diff
@@ -0,0 +1,107 @@
...
+   *   The entity repository.
+   */
+  public function __construct(RouteMatchInterface $current_route_match, EntityTypeManagerInterface $entity_type_manager, EntityRepositoryInterface $entity_repository) {
+    $this->currentRouteMatch = $current_route_match;
+    $this->entityTypeManager = $entity_type_manager;
+    $this->entityRepository = $entity_repository;
+  }
+
+  /**
+   * Node retrieving event handler for metadata.
```
upchuk: Returns the node from the current route for common node routes.

modules/oe_theme_helper/src/EventSubscriber/NodeMetadataEventSubscriber.php
```diff
@@ -0,0 +1,107 @@
...
+   *
+   * @param \Drupal\oe_theme_helper\Event\NodeMetadataEvent $event
+   *   Node retrieving event for metadata.
```
upchuk: The event.

modules/oe_theme_helper/src/EventSubscriber/NodeMetadataEventSubscriber.php
```diff
@@ -0,0 +1,107 @@
...
+        ->addCacheContexts(['route']);
+    }
+
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents(): array {
+    // Subscribing to listening to the Metadata event.
```
upchuk: No need for this comment.

modules/oe_theme_helper/src/MetadataEvents.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper;
+
+/**
+ * Contains all events related to the metadata plugin.
+ */
+final class MetadataEvents {
```
upchuk: PageHeaderMetadataEvents

modules/oe_theme_helper/src/MetadataEvents.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper;
+
+/**
+ * Contains all events related to the metadata plugin.
```
upchuk:
```
Contains all event names related to the page header metadata plugins.
```

modules/oe_theme_helper/src/MetadataEvents.php
```diff
@@ -0,0 +1,21 @@
...
+final class MetadataEvents {
+
+  /**
+   * Name of the event fired during collecting info for metadata building.
+   *
+   * @Event
+   *
+   * @var string
+   */
+  const NODE = 'metadata_events.node';
```
upchuk: Let's call this actually with a bit more namespace: `page_header_metadata_events.node`

modules/oe_theme_helper/src/MetadataEvents.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Drupal\oe_theme_helper;
+
+/**
+ * Contains all events related to the metadata plugin.
+ */
+final class MetadataEvents {
+
+  /**
+   * Name of the event fired during collecting info for metadata building.
```
upchuk:
```
Name of the event used for retrieving the node for the page header metadata plugins.
```

modules/oe_theme_helper/src/MetadataEvents.php
```diff
@@ -0,0 +1,21 @@
...
+   *
+   * @Event
```
upchuk: `@Event`?

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
```diff
@@ -104,25 +122,11 @@ public function getMetadata(): array {
...
 
-      return $node ? $this->entityRepository->getTranslationFromContext($node) : NULL;
-    }
+    $event = new NodeMetadataEvent();
+    $this->eventDispatcher->dispatch(MetadataEvents::NODE, $event);
 
-    // If a node object is present in the route, use that one.
-    $node = $this->currentRouteMatch->getParameter('node');
+    $node = $event->getNode();
     return $node instanceof NodeInterface ? $node : NULL;
```
upchuk: `getNode()` returns already a node or NULL so you dont need this conditional

PR #268, opened 2019-07-11
OPENEUROPA-2060: Implement timeline pattern.

nagyad:


templates/patterns/timeline/timeline.ui_patterns.yml
```diff
@@ -0,0 +1,41 @@
+timeline:
+  label: "Timeline"
+  description: "A timeline is a way of displaying a list of events in chronological order."
+  fields:
+    limit:
+      type: "numeric"
+      label: "Limit"
+      description: "Number of items to show"
+      preview: 3
+      escape: false
```
sergepavle: Can't find for what this element
https://ui-patterns.readthedocs.io/en/8.x-1.x/content/patterns-definition.html
Is this have any value?

nagyad: Yes, otherwise it will be escaped and it will be used as string not int.

templates/patterns/timeline/pattern-timeline.html.twig
```diff
@@ -0,0 +1,22 @@
...
+          "name" (string) (default: ''),
+          "value" (string) (default: '')
+        },
+        ...
+      ]
+#}
+
+{% include '@ecl/ec-component-timeline' with {
+  'limit': limit,
+  'button_label': button_label,
```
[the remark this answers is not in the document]
nagyad: Normally not since the parent pattern has it. But I have to do it because the parent definition misses the t().

templates/patterns/timeline/pattern-timeline.html.twig
```diff
@@ -0,0 +1,22 @@
+{#
+  Parameters:
```
voidtek: To be consistent and correct this description should be different.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -964,3 +964,32 @@
       'div.ecl-dialog[aria-hidden="false"] h3#ecl-dialog-title': "Dialog title"
       'div.ecl-dialog[aria-hidden="false"] p#ecl-dialog-description': "Dialog description"
       'div.ecl-dialog[aria-hidden="false"] div.ecl-dialog__body': "<p>Dialog body</p>"
+- array:
+    '#type': pattern
+    '#id': timeline
+    '#fields':
+      limit: 1
```
voidtek: We should have a second test with `limit: 0`

tests/Kernel/fixtures/rendering.yml
```diff
@@ -964,3 +964,32 @@
...
+      button_label: "Timeline button"
```
voidtek: We should test when we do not specify a button label. 
According to set default value here https://github.com/openeuropa/oe_theme/pull/268/files#diff-24773e644d42d1c7a2ec9a7643d80560R20

templates/patterns/timeline/timeline.ui_patterns.yml
```diff
@@ -0,0 +1,41 @@
+timeline:
```
voidtek: Please use the same data provided here https://v1--europa-component-library.netlify.com/ec/components/detail/ec-component-timeline to set the preview.

voidtek:
Thanks to don't use my suggest commits and to force push during the code review.

nagyad:
What is suggest commits? I created a separate commit for all review comments, I just force pushed a typo as last that had no reason to have separate commits. I don't see your problem @voidtek I implemented all your requests.

PR #269, opened 2019-07-15
OPENEUROPA-1993: ECL2 Update gallery.

imanoleguskiza:
### Description

Update gallery component to ECL2

### Change log

- Added:
- Changed: Update gallery component to ECL2
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/js/gallery.js
+++ b/js/gallery.js
@@ -1,0 +1,15 @@
+/**
+ * @file
+ * ECL gallery behavior.
+ */
+(function (ECL, Drupal) {
+  Drupal.behaviors.eclGallery = {
+    attach: function attach() {
+      var galleryElements = document.querySelectorAll("[data-ecl-gallery]");
+      galleryElements.forEach((galleryElement) => {
+        var gallery = new ECL.Gallery(galleryElement);
+        gallery.init();
+      });
+    }
+  };
+})(ECL, Drupal);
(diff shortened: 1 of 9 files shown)


js/gallery.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @file
+ * ECL gallery behavior.
+ */
+(function (ECL, Drupal) {
+  Drupal.behaviors.eclGallery = {
+    attach: function attach() {
+      var galleryElements = document.querySelectorAll("[data-ecl-gallery]");
```
voidtek: To be consistent
```suggestion
      var elements = document.querySelectorAll('[data-ecl-accordion]');
      elements.forEach(function(element) {
        (new ECL.Accordion(element)).init();
      });
```

templates/patterns/gallery/gallery.ui_patterns.yml
```diff
@@ -7,34 +7,119 @@ gallery:
       label: "Items"
       description: "List of GalleryItemValueObject objects."
       preview:
-        - thumbnail:
```
voidtek: We should keep the same structure of the old pattern.

voidtek (changes requested):

templates/patterns/gallery/gallery.ui_patterns.yml
imanoleguskiza: We are keeping the same structure though, we still have a list of GalleryItemValueObjects. The objects themselves have changed, which I think its ok?

imanoleguskiza:
Outdated PR

PR #274, opened 2019-07-18
OPENEUROPA-2061: Style timeline formatter with ECL and provide test.

nagyad:
First, this https://github.com/openeuropa/oe_content/pull/88 has to be merged.

### DIFF
--- a/templates/field/timeline.html.twig
+++ b/templates/field/timeline.html.twig
@@ -2,8 +2,6 @@
 /**
  * @file
  * Theme override for the timeline field template.
- *
- * @see ./contrib/oe_content/oe_content_timeline/templates/timeline.html.twig
  */
 #}
 {{ pattern('timeline', {
--- a/templates/field/timeline.html.twig
+++ b/templates/field/timeline.html.twig
@@ -1,0 +1,13 @@
+{#
+/**
+ * @file
+ * Theme override for the timeline field template.
+ *
+ * @see ./contrib/oe_content/oe_content_timeline/templates/timeline.html.twig
+ */
+#}
+{{ pattern('timeline', {
+  'limit': limit,
+  'button_label': show_more,
+  'items': items,
+}) }}
--- a/tests/Kernel/TimelineTest.php
+++ b/tests/Kernel/TimelineTest.php
@@ -1,0 +1,218 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme\Kernel;
+
+use Drupal\Core\Entity\Entity\EntityViewDisplay;
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\filter\Entity\FilterFormat;
+use Drupal\node\Entity\Node;
+use Drupal\node\Entity\NodeType;
+use Drupal\Tests\rdf_entity\Traits\RdfDatabaseConnectionTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Tests the timeline field type definition.
+ */
+class TimelineTest extends AbstractKernelTestBase {
+
+  use RdfDatabaseConnectionTrait;
+
+  /**
+   * The node storage.
+   *
+   * @var \Drupal\node\NodeStorageInterface
+   */
+  protected $nodeStorage;
+
+  /**
+   * The node view builder.
+   *
+   * @var \Drupal\Core\Entity\EntityViewBuilderInterface
+   */
+  protected $nodeViewBuilder;
+
+  /**
+   * A field storage to use in this test class.
+   *
+   * @var \Drupal\field\Entity\FieldStorageConfig
+   */
+  protected $fieldStorage;
+
+  /**
+   * The field used in this test class.
+   *
+   * @var \Drupal\field\Entity\FieldConfig
+   */
+  protected $field;
+
+  /**
+   * The display options to use in the formatter.
+   *
+   * @var array
+   */
+  protected $displayOptions = [
+    'type' => 'timeline_formatter',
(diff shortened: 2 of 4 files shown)


tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,218 @@
...
+   * @var \Drupal\Core\Entity\EntityViewBuilderInterface
+   */
+  protected $nodeViewBuilder;
+
+  /**
+   * A field storage to use in this test class.
+   *
+   * @var \Drupal\field\Entity\FieldStorageConfig
+   */
+  protected $fieldStorage;
```
imanoleguskiza: Is there any reason why we store the fieldStorage in a parameter?

tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,218 @@
...
+
+  /**
+   * The field used in this test class.
+   *
+   * @var \Drupal\field\Entity\FieldConfig
+   */
+  protected $field;
```
imanoleguskiza: Same as the fieldStorage, why do we need this parameter in the class?

tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,218 @@
...
+        ],
+      ],
+    ];
+
+    // Create node.
+    $node = Node::create($values);
+    $node->save();
+
+    $entity_type_manager = \Drupal::entityTypeManager()->getStorage('node');
+    $entity_type_manager->resetCache();
```
imanoleguskiza: Could you add a coment perhaps explaining why we need to reload the node and reset the cache?

tests/Kernel/TimelineTest.php
nagyad: Yes, because we need to reuse it. Otherwise it would not be needed indeed.

tests/Kernel/TimelineTest.php
nagyad: Same reason, I need to reuse the object in constructing the display.

tests/Kernel/TimelineTest.php
nagyad: You are right, it is not required.

templates/field/timeline.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Theme override for the timeline field template.
+ *
+ * @see ./contrib/oe_content/oe_content_timeline/templates/timeline.html.twig
```
upchuk: I'd remove this `see` thing. The path is also wrong and it assumes our own project structure.

tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,199 @@
...
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\filter\Entity\FilterFormat;
+use Drupal\node\Entity\Node;
+use Drupal\node\Entity\NodeType;
+use Drupal\Tests\rdf_entity\Traits\RdfDatabaseConnectionTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Tests the timeline field type definition.
```
upchuk: Tests the rendering of the timeline field.

tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,199 @@
...
+    'settings' => [
+      'limit' => 2,
+      'show_more' => 'Button label',
+    ],
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = [
```
upchuk: protected. The parent is protected so we need to keep it as such

tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,199 @@
...
+      'name' => 'My text format',
+      'filters' => [
+        'filter_autop' => [
+          'module' => 'filter',
+          'status' => TRUE,
+        ],
+      ],
+    ])->save();
+
+    // Add text field to entity, to sort by.
```
upchuk: Obsolete comment.

tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,199 @@
...
+        ],
+      ],
+    ];
+
+    // Create node.
+    $node = Node::create($values);
+    $node->save();
+
+    // Verify the timeline is correctly rendered by ECL.
+    $display = EntityViewDisplay::collectRenderDisplay($node, 'default');
```
upchuk: Let's put a comment here explaining what we expect the timeline to be rendered like (given the formatter configuration).

tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,199 @@
...
+    $timeline_body = $crawler->filter('.ecl-timeline__body');
+    $this->assertCount(3, $timeline_body);
+
+    $hidden_timeline_item = $crawler->filter('.ecl-timeline__item--over-limit');
+    $this->assertCount(1, $hidden_timeline_item);
+
+    $show_more_button = $crawler->filter('.ecl-timeline__button');
+    $this->assertCount(1, $show_more_button);
+
+    // Change the limit to show all item without show more button.
```
upchuk: Change the limit to show all items without the "show more" button.

tests/Kernel/TimelineTest.php
```diff
@@ -0,0 +1,199 @@
...
+    $this->displayOptions['settings']['limit'] = 0;
+    $display->setComponent('field_timeline', $this->displayOptions)->save();
+
+    $build = $display->build($node);
+    $output = $this->renderRoot($build);
+    $this->verbose($output);
```
upchuk: Not needed.

PR #277, opened 2019-07-23
OPENEUROPA-2086: ECL 2.0: Timeline.

sergepavle:
### Description

Port the timeline pattern component to ECL2.

### DIFF
--- a/templates/patterns/timeline/timeline.ui_patterns.yml
+++ b/templates/patterns/timeline/timeline.ui_patterns.yml
@@ -13,17 +13,15 @@ timeline:
       label: "Timeline index for starting displaying"
       description: "Item index after which to resume displaying timeline items. Default is null."
       escape: false
-      preview: 0
+      preview: 3
     button_label:
       type: "text"
       label: "Button label"
       description: "Show timeline button label."
-      preview: "Show all timeline items"
     button_label_hidden:
       type: "text"
       label: "Button label hidden"
       description: "Hidden timeline button label."
-      preview: "Hidden all timeline items"
     items:
       type: "array"
       label: "Timeline items"
@@ -55,4 +53,3 @@ timeline:
           body: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer nec tellus congue, pretium nibh eu, mattis ligula. Integer tempus varius enim ut scelerisque. Curabitur rutrum vestibulum elit ac pulvinar."
   libraries:
     - oe_theme/timelines
-
--- a/templates/patterns/timeline/timeline.ui_patterns.yml
+++ b/templates/patterns/timeline/timeline.ui_patterns.yml
@@ -1,0 +1,58 @@
+timeline:
+  label: "Timeline"
+  description: "A timeline is a way of displaying a list of events in chronological order."
+  fields:
+    toggle_collapsed:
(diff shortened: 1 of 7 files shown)


tests/Kernel/fixtures/rendering.yml
```diff
@@ -1047,3 +1047,94 @@
 #      'div.ecl-dialog[aria-hidden="false"] h3#ecl-dialog-title': "Dialog title"
 #      'div.ecl-dialog[aria-hidden="false"] p#ecl-dialog-description': "Dialog description"
 #      'div.ecl-dialog[aria-hidden="false"] div.ecl-dialog__body': "<p>Dialog body</p>"
+- array:
+    '#type': pattern
+    '#id': timeline
+    '#fields':
+      toggle_collapsed: 'Show timelines'
+      toggle_expended: 'Hide timelines'
```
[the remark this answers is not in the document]
sergepavle: I had problem with using this selector
      `'li.ecl-timeline2__item button.ecl-timeline2__toggle[data-ecl-label-expanded="Hide timelines"]': 1`
But by some reasons library `dom-crawler`

templates/patterns/timeline/timeline.ui_patterns.yml
```diff
@@ -0,0 +1,58 @@
+timeline:
```
voidtek: This file should be the same as https://github.com/openeuropa/oe_theme/blob/master/templates/patterns/timeline/timeline.ui_patterns.yml

sergepavle: Ok we could cut new functionality of ECL2 where we could specify from what element show again again lines.
But what we have to do with `button label` (show again hidden items). Because in ecl2 we have to show something?
It this case it should be configurable anymore?
Also question about structure of `items` elements.

voidtek: I know that it's painful, but we should take into consideration the legacy of this timeline pattern.
We don't cut any functionality of ECL2, we should adapt our code.
So for this case, we could add an extra field:
```
    button_label_hidden:
      type: "text"
      label: "Button label hidden"
      description: "Hidden timeline button label."
      preview: "Hidden all timeline items"
```

For item it's the same, the pattern should be more generic and use the preprocess and the template to transform for ecl2.

sergepavle: Ok. Thank you for input. I have taken in consideration and finished rework.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -1047,3 +1047,97 @@
 #      'div.ecl-dialog[aria-hidden="false"] h3#ecl-dialog-title': "Dialog title"
 #      'div.ecl-dialog[aria-hidden="false"] p#ecl-dialog-description': "Dialog description"
 #      'div.ecl-dialog[aria-hidden="false"] div.ecl-dialog__body': "<p>Dialog body</p>"
+- array:
+    '#type': pattern
+    '#id': timeline
+    '#fields':
+      button_label: 'Show timelines'
+      button_label_hidden: 'Hide timelines'
```
voidtek: I tried to remove these two fields to test the internal logic and it's working.

PR #278, opened 2019-07-24
OPENEUROPA-2047: Add class for timeline to limit width.

nagyad:
image

### DIFF
--- a/templates/field/timeline.html.twig
+++ b/templates/field/timeline.html.twig
@@ -4,8 +4,11 @@
  * Theme override for the timeline field template.
  */
 #}
+
+<div class="ecl-editor">
 {{ pattern('timeline', {
   'limit': limit,
   'button_label': show_more,
   'items': items,
 }) }}
+</div>


ademarco (approved):
We are wrapping everything around `ecl-editor` class as both title and long text field should be affected by that. This is not ideal but it will do the job, also considering that the ECL2 upgrade is coming soon.

voidtek:
The way to handle it is to add a `<p class='ecl-paragraph'>` for every item.

as we can see in the demo: 
```
<div class="ecl-timeline__body">
        <p class="ecl-paragraph">
          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer nec tellus congue, pretium nibh eu, mattis ligula. Integer tempus varius enim ut scelerisque. Curabitur rutrum vestibulum elit ac pulvinar.
        </p>
      </div>
```

ademarco (changes requested):
After talking with @brummbar we should not do this in the pattern but in the field template.

nagyad:
@voidtek As Anto wrote, we need all elements to be wrapped, title, and body as well.

ademarco (approved):

ademarco (changes requested):
Let's put this on the outside field template instead, it should be like `field--....html.twig`.

ademarco (approved):
After careful consideration we are back to square one. Approved.

brummbar (approved):

PR #282, opened 2019-07-31
OPENEUROPA-2135: Add publication date to Publication page header metadata

imanoleguskiza:
### Description

The Publication page header needs to show the publication date.

### Change log

- Added: Publication date to Publication page header metadata.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_content_publication/src/Plugin/PageHeaderMetadata/PublicationContentType.php
```diff
@@ -54,6 +111,11 @@ public function getMetadata(): array {
       ],
     ];
 
+    $timestamp = $entity->get('oe_publication_date')->date->getTimestamp();
```
upchuk: If there is no node summary, above we return the metadata and we don't get a chance to set the publication date. Please see NewsContentType as an example of essentially exact same thing.

PR #283, opened 2019-07-31
ECL 2.0: List item patterns

voidtek:
### Description

- Add list-item templates
- Refactor port ecl1 to ecl2
- Fix template consistency


templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
```diff
@@ -2,12 +2,30 @@
...
  */
 #}
-{% include '@ecl/ec-component-list-item' with {
-  'variant': 'navigation',
-  'title': title,
-  'href': url,
-  'detail': detail,
-  'screen_reader': 'List item'|t
+
+{% set _infos = {} %}
```
imanoleguskiza: The infos are not used in the navigation variant.

templates/patterns/list_item/list_item.ui_patterns.yml
```diff
@@ -59,6 +59,14 @@ list_item:
...
-        src: "http://via.placeholder.com/400x400"
+        src: "https://placeimg.com/1200/900/arch"
         alt: "Alternative text for image"
-
+    infos:
+      type: "array"
+      label: "Infos"
+      description: "A list of infos of the page."
+      preview:
+        - text: "Monday 8 February"
```
imanoleguskiza: We already have a date field in the pattern. I'm wondering if we should try and add the value on the date field to the infos whenever infos is used. 
Also having both meta and infos is a bit confusing. I know it comes from ECL but as it stands there is no way to know what this fields mean without looking at the code.

templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
```diff
@@ -2,12 +2,30 @@
...
+    icon : {
+      type : 'general',
+      name : infos_item.icon,
+      path : ecl_icon_path
+    }
+  } %}
+  {% set _infos = _infos|merge([_info]) %}
+{% endfor %}
+
+{% include '@ecl/card' with {
```
imanoleguskiza: The ticket mentions using the Navigation list composition, it would be nice if you could put a comment in the ticket or here explaining why you chose to use the card template instead.

templates/patterns/list_item/pattern-list-item--variant-thumbnail-primary.html.twig
```diff
@@ -2,14 +2,44 @@
...
+
+{% set _meta = '' %}
+{% for meta_item in meta %}
+  {% if _meta != '' %}
+    {% set _meta = _meta ~ ' | ' %}
+  {% endif %}
+  {% set _meta = _meta ~ meta_item %}
+{% endfor %}
+
+{% set _infos = {} %}
```
imanoleguskiza: This transformation happens more than once. Why not do it in the preprocess function so it applies to all variants regardless?

templates/patterns/list_item/pattern-list-item--variant-thumbnail-primary.html.twig
```diff
@@ -2,14 +2,44 @@
...
-  'title': title,
-  'metas': meta,
-  'href': url,
-  'detail': detail,
-  'primary_image': image,
-  'screen_reader': 'List item'|t
+
+{% set _image_position = image_position|default('before') %}
+
+{% set _meta = '' %}
```
imanoleguskiza: THe _meta variable is never used.

templates/components-custom/ec-component-card/card-thumbnail.html.twig
```diff
@@ -0,0 +1,148 @@
+{% spaceless %}
```
imanoleguskiza: Cann you add a comment explaining the need for this template? Also add a @see tag pointing to the original template it was based on.

templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
voidtek: good catch.

templates/patterns/list_item/list_item.ui_patterns.yml
voidtek: I understand I will change the preview to be consistent with ECL2.
https://ec.europa.eu/component-library/ec/templates/compositions/content-item/code/

templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
voidtek: You are alright.

templates/patterns/list_item/pattern-list-item--variant-thumbnail-primary.html.twig
voidtek: Yes.

templates/components-custom/ec-component-card/card-thumbnail.html.twig
voidtek: True.

templates/patterns/list_item/pattern-list-item--variant-thumbnail-primary.html.twig
voidtek: It's true that this transformation happens more than once, but these patterns start to use different ways (Highline, navigation, thumbnail, others) so it's more flexible to leave it there.
And to be consistent with other pattern components, I prefer to leave this transformation in the template as pattern_page_header, pattern_file, others.
But I am open to change my opinion.

templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
voidtek: Commented on the ticket.

templates/components-custom/ec-component-card/card-thumbnail.html.twig
voidtek: Pushed.

templates/components-custom/ec-component-card/card-thumbnail.html.twig
```diff
@@ -0,0 +1,150 @@
+{#
```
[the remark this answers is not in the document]
imanoleguskiza: The problem here is that we are not really adding functionality to the card component, we are simply reusing parts of it. The INNO team already stablished that they are not going to use the list item component for ECL2 so the implementation we do will be entirely up to us. That being said, we could simply not call this template card-thumbnail to avoid confusion and call it navigation-thumbnail and host it in the list-item folder or something like that.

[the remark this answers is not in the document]
ademarco: Yeah we shouldn't use patching to add entire components, in this case we can maintain them on our side. We could maybe check if ECL Twig wants to maintain them instead, but for now let's proceed with by creating a new template.

modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
```diff
@@ -88,16 +88,16 @@ public function testNodeRoutes(): void {
...
     // Verify that the block is also shown in the latest version route with the
     // correct draft revision loaded.
     $this->drupalGet($node->toUrl('latest-version'));
-    $this->assertSession()->elementTextContains('css', '.ecl-page-header__intro', $draft_revision_body);
+    $this->assertSession()->elementTextContains('css', '.ecl-page-header__description', $draft_revision_body);
 
     // Verify also for the node single revision route.
     $this->drupalGet($first_revision_url);
-    $this->assertSession()->elementTextContains('css', '.ecl-page-header__intro', $published_revision_body);
+    $this->assertSession()->elementTextContains('css', '.ecl-page-header__description', $published_revision_body);
```
ademarco: Out of scope but we can keep them, since 2.x is such a moving target.

templates/components-custom/ec-component-card/card-thumbnail.html.twig
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * @file
+   * Theme implementation for a ECL thumbnail list item component.
+   *
+   * @see @ecl/card.html.twig
```
ademarco: We should use the "Content item" composition instead of the card component: https://ec.europa.eu/component-library/ec/templates/compositions/content-item/code/

The composition was not available at the time of writing, it looks more like it is today on ECL 1.x:

image

templates/missing-component.html.twig
```diff
@@ -5,3 +5,7 @@
  */
 #}
 {# @todo: Remove this template once the migration to ECL 2.0 is done. #}
+<div class="ecl-u-bg-grey-25 ecl-u-pa-s">
+  <p class="ecl-u-type-color-blue ecl-u-type-paragraph-s ecl-u-type-bold">Missing component '{{ theme_hook_original }} - {{ variant }}' template.</p>
+</div>
```
ademarco: Can we remove this please? It clutters the theme, we have this information in the error message anyway. If you really want to keep it for development reasons you can make it configurable with a [Drupal settings](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Site%21Settings.php/function/Settings%3A%3Aget/8.2.x) toggle and disable it by default. Up to you.

templates/patterns/list_item/list_item.ui_patterns.yml
```diff
@@ -59,6 +59,14 @@ list_item:
...
-
+    infos:
+      type: "array"
+      label: "Infos"
+      description: "A list of infos of the page."
+      preview:
+        - text: "Munich, Germany"
+          icon: "location"
+        - text: "Live stream available"
+          icon: "livestreaming"
```
ademarco: On the "Content item" composition these are called "Secondary meta", ideally we could create patterns for these, so to reuse as much as we can. Would it be too bad to split this out in a follow up issue? So we can give it the right amount of attention?

templates/patterns/list_item/pattern-list-item--variant-highlight.html.twig
```diff
@@ -4,10 +4,12 @@
...
-  'primary_image': image,
-  'screen_reader': 'List item'|t
+{% include '@ecl/card' with {
+  card : {
+    title: {
+      label: title,
+      path: url
+    },
+    image: image.src
+  }
```
ademarco: This is ok, we will use the card component for the highlight variant, and not the "Content item" composition.

templates/patterns/list_item/pattern-list-item.html.twig
```diff
@@ -2,12 +2,31 @@
...
+  {% set _info = {
+    label : infos_item.text,
+    icon : {
+      type : 'general',
+      name : infos_item.icon,
+      path : ecl_icon_path
+    }
+  } %}
+  {% set _infos = _infos|merge([_info]) %}
+{% endfor %}
```
ademarco: As said above, I'd remove the `infos` all together in this PR and move it to a follow up, maybe we will get ECL Twig to provide this by the time we need them.

templates/patterns/page_header/pattern-page-header.html.twig
```diff
@@ -44,7 +44,7 @@
 {% if identity is not empty %}
   <div class="ecl-u-bg-blue ecl-u-pv-2xs">
     <div class="ecl-container ">
-      <h2 class="ecl-u-color-white ecl-u-type-heading-2">{{ identity }}</h2>
+      <h2 class="ecl-u-type-color-white ecl-u-type-heading-2">{{ identity }}</h2>
```
ademarco: Out of scope but ok to include it.

templates/ui_patterns_library/patterns-meta-information.html.twig
```diff
@@ -5,30 +5,28 @@
...
+      <tr class="ecl-table__row">
+        <td data-ecl-table-header="{{ "Field"|t }}" class="ecl-table__cell"><code>{{ field.name }}</code></td>
+        <td data-ecl-table-header="{{ "Label"|t }}" class="ecl-table__cell">{{ field.label }}</td>
+        <td data-ecl-table-header="{{ "Type"|t }}" class="ecl-table__cell"><code>{{ field.type }}</code></td>
+        <td data-ecl-table-header="{{ "Description"|t }}" class="ecl-table__cell">{{ field.description }}</td>
       </tr>
     {% endfor %}
     </tbody>
   </table>
-
```
ademarco: Out of scope but ok to include it.

templates/ui_patterns_library/patterns-variant-meta-information.html.twig
```diff
@@ -7,18 +7,18 @@
...
-    <tbody>
-    <tr>
-      <td>{{ variant.label }}</td>
-      <td><code>{{ variant.name }}</code></td>
-      <td>{{ variant.description }}</td>
+    <tbody class="ecl-table__body">
+    <tr class="ecl-table__row">
+      <td data-ecl-table-header="{{ "Variant"|t }}" class="ecl-table__cell">{{ variant.label }}</td>
+      <td data-ecl-table-header="{{ "Name"|t }}" class="ecl-table__cell"><code>{{ variant.name }}</code></td>
+      <td data-ecl-table-header="{{ "Description"|t }}" class="ecl-table__cell">{{ variant.description }}</td>
```
ademarco: Out of scope but ok to include it.

tests/Kernel/MediaRenderTest.php
```diff
@@ -76,27 +76,26 @@ public function testDocumentMedia(): void {
...
-    $file_info_language = $file_row->filter('.ecl-file__info span.ecl-file__language');
+    $file_info_language = $file_row->filter('.ecl-file__info div.ecl-file__language');
     $this->assertContains('English', $file_info_language->text());
 
-    $file_info_properties = $file_row->filter('.ecl-file__info span.ecl-file__properties');
+    $file_info_properties = $file_row->filter('.ecl-file__info div.ecl-file__meta');
     $this->assertContains('KB - PDF)', $file_info_properties->text());
 
-    $file_download_link = $file_row->filter('.ecl-file__download');
+    $file_download_link = $file_row->filter('a.ecl-file__download');
```
ademarco: This is out of scope, we can leave it but we need to make sure it does not impact any other issue.

templates/components-custom/ec-component-card/card-thumbnail.html.twig
voidtek: Changed.
new template `compositions/ec-component-content-item/content-item.html.twig`

modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
voidtek: :+1: thanks
The purpose is to add a maximum of PHPUnit tests.

templates/missing-component.html.twig
voidtek: Removed.

templates/patterns/list_item/pattern-list-item--variant-highlight.html.twig
voidtek: :+1:

templates/patterns/page_header/pattern-page-header.html.twig
voidtek: After upgrading packages, the class `ecl-u-color-white` doesn't exist and replaced by `ecl-u-type-color-white` .
If we had visual tests, this upgrade will break the test. so one part is on the scope of the ticket.
;)

templates/ui_patterns_library/patterns-meta-information.html.twig
voidtek: :+1:

templates/ui_patterns_library/patterns-variant-meta-information.html.twig
voidtek: :+1:

tests/Kernel/MediaRenderTest.php
voidtek: This media is already implemented with the new ECL2, so normally no impact.
I prefer to remove it because it's not consistent with the others.

templates/patterns/list_item/pattern-list-item.html.twig
```diff
@@ -2,12 +2,17 @@
...
-  'screen_reader': 'List item'|t
+{% include '@ecl/card' with {
+  card : {
+    description: detail,
+    title: {
+      label: title,
+      path: url
+    },
+    meta: meta
+  }
```
ademarco: This should also be done with content item composition, we use the card only on the highlight variant.

templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
```diff
@@ -4,10 +4,12 @@
...
-  'detail': detail,
-  'screen_reader': 'List item'|t
+{% include '@ecl/card' with {
+  card : {
+    title: {
+      label: title,
+      path: url
+    },
+    description: detail
+  }
```
ademarco: This should also be done with content item composition, we use the card only on the highlight variant.

templates/patterns/list_item/pattern-list-item.html.twig
voidtek: ok

templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
voidtek: ok

.drone.yml
```diff
@@ -68,21 +68,7 @@ pipeline:
       # @todo: Re-run line below after port to ECL 2.0 is completed.
       # - ./vendor/bin/phpunit
       # @todo: Manually add tests to run until port to ECL 2.0 is completed.
```
ademarco: I think this comment is outdated now, since we use the `ecl1` tag. It should probably read:
```
       # @todo: remove the `@group ecl1` on PHPUnit tests to have them executed. We will remove this after porting is completed.
```

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,123 @@
...
+        }
+    - "extra_classes" (string) (default: '')
+    - "extra_attributes" (array) (default: []): format: [
+        {
+          "name" (string) (default: ''),
+          "value" (string) (default: '')
+        },
+        ...
+      ]
+#}
```
ademarco: Is this description still relevant? It seems to be related to the `card` component.

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,123 @@
...
+    <div class="ecl-content-item__title ecl-u-type-prolonged-m ecl-u-type-bold ecl-u-mt-xs">
+      {%- if _content_item.title.path %}
+        {% include '../ec-component-link/link.html.twig' with {
+          link: _content_item.title
+        } only %}
+      {% else %}
+        {{- _content_item.title.label }}
+      {%- endif -%}
+    </div>
+    <div class="ecl-content-item__description ecl-u-type-paragraph ecl-u-type-color-grey-100 ecl-u-mt-xs">{{ _content_item.description }}</div>
```
ademarco: This is a `p` tag [on the playground](https://ec.europa.eu/component-library/playground/ec/?path=/story/templates-compositions--content-items).

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,123 @@
...
+        {% for info in _content_item.infos %}
+          <li class="ecl-content-item__info-item ecl-u-d-flex ecl-u-align-items-center ecl-u-type-color-grey-75 ecl-unordered-list__item">
+            {% include '../ec-component-icon/icon.html.twig' with {
+              icon: info.icon|merge({ size: 'm' })
+            } only %}
+            <span class="ecl-u-type-s ecl-u-ml-s">{{ info.label }}</span>
+          </li>
+        {% endfor %}
+      </ul>
+    {% endif %}
```
ademarco: We don't support `infos` yet right? Let's remove this part for the moment, and any other reference to that.

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,123 @@
...
+      <div class="ecl-content-item__meta ecl-u-type-s ecl-u-type-color-grey-75">{{ _content_item.meta|join(" | ") }}</div>
+    {% endif %}
+    <div class="ecl-content-item__title ecl-u-type-prolonged-m ecl-u-type-bold ecl-u-mt-xs">
+      {%- if _content_item.title.path %}
+        {% include '../ec-component-link/link.html.twig' with {
+          link: _content_item.title
+        } only %}
+      {% else %}
+        {{- _content_item.title.label }}
+      {%- endif -%}
```
ademarco: The HTML on [the playground](https://ec.europa.eu/component-library/playground/ec/?path=/story/templates-compositions--content-items) does not use the link component right? But a simple `a` tag. Visuals seems to be different, like this implementation has the title underlined while on the preview it looks like it's not:

image

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,123 @@
...
+      </ul>
+    {% endif %}
+  </div>
+  {% if _content_item.image and _content_item.image_position == 'after' %}
+    <div role="img" aria-label="{{ _content_item.image.alt }}" class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-ml-xl ecl-u-d-lg-none"
+         style="background-image:url(&quot;{{ _content_item.image.src }}&quot;);background-size:contain;height:5rem;width:7.5rem"></div>
+    <div role="img" aria-label="{{ _content_item.image.alt }}"
+         class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-ml-xl ecl-u-d-none ecl-u-d-lg-block"
+         style="background-image:url(&quot;{{ _content_item.image.src }}&quot;);background-size:contain;height:8.75rem;width:13.125rem">
+    </div>
```
ademarco: The image part is used twice, here and on the top. We can use [Twig's `macros`](https://twig.symfony.com/doc/2.x/tags/macro.html) to define this part only once and use it in two different places.

.drone.yml
voidtek: good catch

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: This description for the object content-item is required to be clear.
All the description is updated.

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: We have an issue if we want to use a `p` because we are including a <div class='ecl-editor>...</div>' and it breaks the html structure.
For example, if you check the kernel test `ParagraphsTest::testListItem` the result is `<p ...><div class='ecl-editor'><p>...</p></div></p>`.
If we want to use `p` we should change the field description on the paragraph.

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: I confirm that the pattern doesn't support the `infos` yet. I prefer to keep the same template (but resolving bugs) that ECL.
But I will remove this reference.

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: The part image is not really the same, some classes are different.
Updated.

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: Good catch. The title is a link with the type 'standalone'. updated.

templates/compositions/ec-component-content-item/content-item.html.twig
ademarco: Ok but I think the issue is the `<div class='ecl-editor'>` getting into the item description right? How that ends up there?

templates/patterns/list_item/pattern-list-item--variant-thumbnail-primary.html.twig
```diff
@@ -4,12 +4,17 @@
...
 #}
-{% include '@ecl/ec-component-list-item' with {
-  'variant': 'thumbnail',
-  'title': title,
-  'metas': meta,
-  'href': url,
-  'detail': detail,
-  'primary_image': image,
-  'screen_reader': 'List item'|t
+{% set _image_position = image_position|default('before') %}
```
ademarco: No need to set a variable for this one, let's directly assign it to the value below:
```
     image_position: image_position|default('before'),
```
It will increase Twig file readability.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -166,26 +168,18 @@ public function testListItem(): void {
...
 
-    $meta_elements = $crawler->filter('.ecl-meta__item');
-    $this->assertCount(3, $meta_elements);
-    $this->assertEquals('Meta 1', trim($meta_elements->getNode(0)->nodeValue));
-    $this->assertEquals('Meta 2', trim($meta_elements->getNode(1)->nodeValue));
-    $this->assertEquals('Meta 3', trim($meta_elements->getNode(2)->nodeValue));
+    $this->assertEquals('Meta 1 | Meta 2 | Meta 3', trim($crawler->filter('article.ecl-content-item div.ecl-content-item__meta')->text()));
 
     // No images should be rendered in this variant.
-    $this->assertCount(0, $crawler->filter('img.ecl-image'));
```
ademarco: We should keep this assertion about no image being rendered.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -166,26 +168,18 @@ public function testListItem(): void {
...
-    // No date should be rendered neither.
```
ademarco: Keep this comment for the assertion below.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -194,64 +188,45 @@ public function testListItem(): void {
...
-    $this->assertCount(1, $crawler->filter('.ecl-list-item.ecl-list-item--highlight'));
-    $this->assertEquals('Item title', trim($crawler->filter('.ecl-list-item__title')->text()));
-    // The description should not be rendered in this variant.
-    $this->assertCount(0, $crawler->filter('.ecl-list-item__detail'));
-    // Neither the date.
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__week-day'));
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__day'));
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__month'));
-    // Neither the metas.
-    $this->assertCount(0, $crawler->filter('.ecl-meta__item'));
```
ademarco: We need to keep all the negative assertion above, this will make sure those areas are not filled on the card when we use it for an highlight variant:

image

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -194,64 +188,45 @@ public function testListItem(): void {
...
+    // @todo will be fixed on OPENEUROPA-2123.
+    // $this->assertEquals('Druplicon', $image_element->attr('alt'));
     // Change the variant to thumbnail primary.
     $paragraph->get('oe_paragraphs_variant')->setValue('thumbnail_primary');
     $paragraph->save();
 
     $html = $this->renderParagraph($paragraph);
     $crawler = new Crawler($html);
 
-    $this->assertCount(1, $crawler->filter('.ecl-list-item.ecl-list-item--thumbnail'));
```
ademarco: This asserts that we don't show both images, primary and secondary, and it should be kept.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -194,64 +188,45 @@ public function testListItem(): void {
...
+    $this->assertEquals('Druplicon', $image_element->attr('aria-label'));
 
-    $meta_elements = $crawler->filter('.ecl-meta__item');
-    $this->assertCount(3, $meta_elements);
-    $this->assertEquals('Meta 1', trim($meta_elements->getNode(0)->nodeValue));
-    $this->assertEquals('Meta 2', trim($meta_elements->getNode(1)->nodeValue));
-    $this->assertEquals('Meta 3', trim($meta_elements->getNode(2)->nodeValue));
-
-    // The secondary image markup should not be rendered.
-    $this->assertCount(0, $crawler->filter('.ecl-list-item__secondary img.ecl-image'));
```
ademarco: We should keep this assertion when testing the `thumbnail_primary` variant.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -260,63 +235,76 @@ public function testListItem(): void {
     $html = $this->renderParagraph($paragraph);
     $crawler = new Crawler($html);
 
-    $this->assertCount(1, $crawler->filter('.ecl-list-item.ecl-list-item--thumbnail'));
```
ademarco: We should keep this assertion.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -260,63 +235,76 @@ public function testListItem(): void {
...
-
-    $meta_elements = $crawler->filter('.ecl-meta__item');
-    $this->assertCount(3, $meta_elements);
-    $this->assertEquals('Meta 1', trim($meta_elements->getNode(0)->nodeValue));
-    $this->assertEquals('Meta 2', trim($meta_elements->getNode(1)->nodeValue));
-    $this->assertEquals('Meta 3', trim($meta_elements->getNode(2)->nodeValue));
+    $this->assertEquals('Druplicon', $image_element->attr('aria-label'));
 
-    // The primary image markup should not be rendered.
-    $this->assertCount(0, $crawler->filter('.ecl-list-item__primary img.ecl-image'));
```
ademarco: We should keep this assertion.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -260,63 +235,76 @@ public function testListItem(): void {
...
-    // The date field should not be rendered.
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__week-day'));
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__day'));
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__month'));
+    $this->assertEquals('Meta 1 | Meta 2 | Meta 3', trim($crawler->filter('article.ecl-content-item div.ecl-content-item__meta')->text()));
 
-    // Change the variant to date.
-    $paragraph->get('oe_paragraphs_variant')->setValue('date');
+    // Change the variant to thumbnail secondary.
+    $paragraph->get('oe_paragraphs_variant')->setValue('navigation');
```
ademarco: Comment is wrong, it should read:
```
      // Change the variant to navigation.
```

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -260,63 +235,76 @@ public function testListItem(): void {
...
     $paragraph->save();
 
     $html = $this->renderParagraph($paragraph);
     $crawler = new Crawler($html);
 
-    $this->assertCount(1, $crawler->filter('.ecl-list-item.ecl-list-item--date'));
```
ademarco: We should keep this assertion as we apparently want to date to be shown.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -260,63 +235,76 @@ public function testListItem(): void {
...
 
-    $link_element = $crawler->filter('.ecl-list-item__link');
+    $link_element = $crawler->filter('article.ecl-content-item div.ecl-content-item__title a.ecl-link');
     $this->assertCount(1, $link_element);
     $this->assertEquals('http://www.example.com/', $link_element->attr('href'));
 
-    $this->assertEquals('Thu', trim($crawler->filter('.ecl-date-block__week-day')->text()));
-    $this->assertEquals('24', trim($crawler->filter('.ecl-date-block__day')->text()));
-    $this->assertEquals('Sep', trim($crawler->filter('.ecl-date-block__month')->text()));
-    $this->assertEquals('1981', trim($crawler->filter('.ecl-date-block__year')->text()));
```
ademarco: Can we keep asserting this text in the meta?

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -260,63 +235,76 @@ public function testListItem(): void {
...
+    $this->assertCount(1, $crawler->filter('article.ecl-content-item > div'));
+    $this->assertCount(0, $crawler->filter('article.ecl-content-item > div.ecl-u-d-lg-block'));
+    $this->assertCount(0, $crawler->filter('article.ecl-content-item div.ecl-content-item__meta'));
 
-    // No images should be rendered in this variant.
-    $this->assertCount(0, $crawler->filter('img.ecl-image'));
-    // Neither the metas.
-    $this->assertCount(0, $crawler->filter('.ecl-meta__item'));
```
ademarco: We have to keep these assertions.

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,94 @@
...
+{% endif %}
+
+{# Create macro #}
+
+{% macro image(src, alt, class1, class2) %}
+  <div role="img" aria-label="{{ alt }}" class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl {{ class1 }}"
+       style="background-image:url(&quot;{{ src }}&quot;);background-size:contain;height:5rem;width:7.5rem"></div>
+  <div role="img" aria-label="{{ alt }}"
+       class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl ecl-u-d-none {{ class2 }}"
+       style="background-image:url(&quot;{{ src }}&quot;);background-size:contain;height:8.75rem;width:13.125rem">
```
ademarco: The `&quot;` must be removed, it is probably a rendering bug of the ECL UI thingie. According the [W3 specification](https://www.w3.org/TR/CSS21/syndata.html#uri) quotes can be double `"`, single `'`, or not present at all, and more importantly, the URL inside must be escaped for a few special characters, when not using quotes:

> Some characters appearing in an unquoted URI, such as parentheses, white space characters, single quotes (') and double quotes ("), must be escaped with a backslash so that the resulting URI value is a URI token: '\(', '\)'. 

I suggest to use single quotes and to escape parenthesis and double quotes.

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,94 @@
...
+{#
+  /**
+   * @file
+   * Theme implementation for a ECL Content Item.
+   */
+  Parameters:
+    - "content_item" (associative array) (default: predefined structure): format:
+        {
+          description: '', (string) (default: ''): Description of Content Item
+          meta: [], (array) (default: []): Meta's for the Content Item
```
ademarco: Should read:
```
           meta: [], (array) (default: []): Meta information for the Content Item
```

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,94 @@
...
+          title: {}, (associative array) (default: {}): Predefined structure compatible with Link component. If Content Item type is a 'tile', only label property is required.
+          image: '', (string) (default: ''): Url/path to background image (non required if Content Item type is a 'tile')
+          image_position: 'after' (string) (Can be 'before' or 'after'. Is required only if Icon is specified)
```
ademarco: There is no `Icon` property as such, sure this comment is correct?

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,94 @@
...
+   * @file
+   * Theme implementation for a ECL Content Item.
+   */
+  Parameters:
+    - "content_item" (associative array) (default: predefined structure): format:
+        {
+          description: '', (string) (default: ''): Description of Content Item
+          meta: [], (array) (default: []): Meta's for the Content Item
+          title: {}, (associative array) (default: {}): Predefined structure compatible with Link component. If Content Item type is a 'tile', only label property is required.
+          image: '', (string) (default: ''): Url/path to background image (non required if Content Item type is a 'tile')
```
ademarco: Is this `tile` or `title`? Either way I don't see any reference of it in the implementation.

templates/patterns/list_item/pattern-list-item--variant-thumbnail-primary.html.twig
voidtek: True.

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: :+1:

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: Good catch

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: :+1:

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: Legacy that I need remove.

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: :+1:

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: The issue here is that the description field on oe_paragraph is a `field_oe_text_long` and not a `field_oe_text`.
field_oe_text_long includes `<p>` and other tag.
According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p
`Paragraphs are block-level elements, and notably will automatically close if another block-level element is parsed before the closing </p> tag.`

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: I changed the name of the class to be more clear.

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: `ecl-card__description` and `ecl-card__meta` are added.

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: Tested during link of the image.

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: I chose that we test that the image is print only once.

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: Already present on the block for the image. check below.

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: I chose that we test that the image is print only once.

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: The variant navigation doesn't have meta.

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: Already on the code, check on the top. I will add code comments to be more clear.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -166,26 +168,18 @@ public function testListItem(): void {
...
-    $this->assertEquals('Meta 3', trim($meta_elements->getNode(2)->nodeValue));
+    $this->assertEquals('Meta 1 | Meta 2 | Meta 3', trim($crawler->filter('article.ecl-content-item div.ecl-content-item__meta')->text()));
 
     // No images should be rendered in this variant.
-    $this->assertCount(0, $crawler->filter('img.ecl-image'));
-    // No date should be rendered neither.
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__week-day'));
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__day'));
-    $this->assertCount(0, $crawler->filter('.ecl-date-block__month'));
+    $this->assertCount(0, $crawler->filter('article.ecl-content-item > div.ecl-u-d-lg-block'));
```
ademarco: Can we use `'article.ecl-content-item > div[role="img"].ecl-u-d-lg-block'` as it's done below? So it's more clear we are referring to the image part.

tests/Kernel/Paragraphs/ParagraphsTest.php
ademarco: Ok so the pattern behaves is differently than it was on ECL 1. I guess we should track this somewhere, can we open an issue for @kugta to check that all the new patterns do what they are supposed to. We should add this one in the list for them to test.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -260,63 +256,85 @@ public function testListItem(): void {
...
-    $this->assertEquals('Meta 2', trim($meta_elements->getNode(1)->nodeValue));
-    $this->assertEquals('Meta 3', trim($meta_elements->getNode(2)->nodeValue));
+    $this->assertEquals('Meta 1 | Meta 2 | Meta 3', trim($crawler->filter('article.ecl-content-item div.ecl-content-item__meta')->text()));
 
-    // The primary image markup should not be rendered.
-    $this->assertCount(0, $crawler->filter('.ecl-list-item__primary img.ecl-image'));
     // The date field should not be rendered.
     $this->assertCount(0, $crawler->filter('.ecl-date-block__week-day'));
     $this->assertCount(0, $crawler->filter('.ecl-date-block__day'));
     $this->assertCount(0, $crawler->filter('.ecl-date-block__month'));
```
ademarco: Since that markup will never be there these 3 assertions will always be true, I suggest to comment them out and add a todo on top referring to the date variant issue, so we make sure we deal with this particular thing in the scope of that issue.

voidtek: Perfect.

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: I prefer to use `$this->assertCount(0, $crawler->filter('article.ecl-content-item > div[role="img"]'));` because it's more generic and handles 2 cases:
- no image at all
- no image visible

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,94 @@
...
+{% macro image(src, alt, class1, class2) %}
+  <div role="img" aria-label="{{ alt }}" class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl {{ class1 }}"
+       style="background-image:url('{{ src }}');background-size:contain;height:5rem;width:7.5rem"></div>
+  <div role="img" aria-label="{{ alt }}"
+       class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl ecl-u-d-none {{ class2 }}"
+       style="background-image:url('{{ src }}');background-size:contain;height:8.75rem;width:13.125rem">
+  </div>
+{% endmacro %}
+
+{# Print the result #}
```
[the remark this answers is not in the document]
voidtek: This file is a composition and all ecl2 templates have this comment.
To be consistent, with others I suggest to keep it.

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,94 @@
...
+{% endif %}
+
+{# Create macro #}
+
+{% macro image(src, alt, class1, class2) %}
+  <div role="img" aria-label="{{ alt }}" class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl {{ class1 }}"
+       style="background-image:url('{{ src }}');background-size:contain;height:5rem;width:7.5rem"></div>
+  <div role="img" aria-label="{{ alt }}"
+       class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl ecl-u-d-none {{ class2 }}"
+       style="background-image:url('{{ src }}');background-size:contain;height:8.75rem;width:13.125rem">
```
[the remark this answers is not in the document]
voidtek: done

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -0,0 +1,92 @@
...
+{% endif %}
+
+{# Create macro #}
+
+{% macro image(src, alt, class1, class2) %}
+  <div role="img" aria-label="{{ alt }}" class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl {{ class1 }}"
+       style="background-image:url({{ src }});background-size:contain;height:5rem;width:7.5rem"></div>
+  <div role="img" aria-label="{{ alt }}"
+       class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl ecl-u-d-none {{ class2 }}"
+       style="background-image:url({{ src }});background-size:contain;height:8.75rem;width:13.125rem">
```
ademarco: We didn't have to remove the single quotes, we have to make sure that whater character is passed in the URL does not break the HTML. For example, if I use this:

```yaml
    image:
      type: "array"
      label: "Primary image"
      description: "Attributes of the image to render"
      preview:
        src: "https://placeimg.com/1200/900/arch&test=')"
        alt: "Alternative text for image"
```

The the HTML gets rendered like that:

image

Which is clearly broken.

tests/Kernel/Paragraphs/ParagraphsTest.php
ademarco: Ok that would work too.

ademarco: Only, how we test if the image is on the left or on the right? Did we test that before? If we did we should test this also now.

templates/compositions/ec-component-content-item/content-item.html.twig
voidtek: Pushed

tests/Kernel/Paragraphs/ParagraphsTest.php
voidtek: I added tests

templates/compositions/ec-component-content-item/content-item.html.twig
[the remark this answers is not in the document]
voidtek: I removed

templates/compositions/ec-component-content-item/content-item.html.twig
ademarco: So about the issue above we decided to open a follow up as we need to carefully tests all sorts of scenarios with this one. We will add a comment with a link to said follow up.

PR #285, opened 2019-08-01
OPENEUROPA-2145: Add spacing between paragraphs.

sergepavle:
### Description

Add spacing between the paragraphs.

Check ECL for an indication of spacing value that we can use and use that. Otherwise a default of 20px would be good.

This needs to go into oe_theme.

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -245,6 +245,18 @@ function oe_theme_preprocess_form_element(array &$variables): void {
  * Add paragraph formatting to body.
  */
 function oe_theme_preprocess_field(&$variables, $hook): void {
+  // Add space between paragraphs.
+  if (($variables['entity_type'] === 'node' &&
+      $variables['field_name'] === 'paragraphs' &&
+      $variables['element']['#bundle'] === 'landing_page') ||
+    ($variables['entity_type'] === 'block_content' &&
+      $variables['field_name'] === 'field_content' &&
+      $variables['element']['#bundle'] === 'content_block')) {
+    foreach ($variables['items'] as &$item) {
+      $item['attributes']->addClass('ecl-u-mv-s');
+    }
+  }
+
   if ($variables['field_name'] === 'body') {
     $variables['attributes']['class'][] = 'ecl-editor';
   }
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -79,6 +79,11 @@ function oe_theme_theme_suggestions_field_alter(array &$suggestions, array $vari
     // fields, while keeping all the rest markup-free.
     array_unshift($suggestions, 'field__bare');
   }
+
+  // Add suggestions for all 'Entity reference revisions' for each target type.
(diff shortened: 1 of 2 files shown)


oe_theme.theme
```diff
@@ -245,6 +245,18 @@ function oe_theme_preprocess_form_element(array &$variables): void {
...
+      $variables['field_name'] === 'paragraphs' &&
+      $variables['element']['#bundle'] === 'landing_page') ||
+    ($variables['entity_type'] === 'block_content' &&
+      $variables['field_name'] === 'field_content' &&
+      $variables['element']['#bundle'] === 'content_block')) {
+    foreach ($variables['items'] as &$item) {
+      $item['attributes']->addClass('ecl-u-mv-s');
+    }
+  }
+
```
ademarco: We can't simply hardcode bundle names in the theme, this is a generic component.

oe_theme.theme
```diff
@@ -245,6 +245,18 @@ function oe_theme_preprocess_form_element(array &$variables): void {
  * Add paragraph formatting to body.
  */
 function oe_theme_preprocess_field(&$variables, $hook): void {
+  // Add space between paragraphs.
+  if (($variables['entity_type'] === 'node' &&
+      $variables['field_name'] === 'paragraphs' &&
+      $variables['element']['#bundle'] === 'landing_page') ||
```
ademarco: We can't simply hardcode bundle names in the theme, this is a generic component.

ademarco (dismissed):

oe_theme.theme
nagyad: I think this should not be tied to any bundle, spacing has to be correct on any bundle that has paragraph field.

oe_theme.theme
```diff
@@ -245,6 +245,18 @@ function oe_theme_preprocess_form_element(array &$variables): void {
  * Add paragraph formatting to body.
  */
 function oe_theme_preprocess_field(&$variables, $hook): void {
+  // Add space between paragraphs.
```
voidtek: Could have more information concerning the class chose?
for example the link of the ECL2 marge documentation.

oe_theme.theme
```diff
@@ -245,6 +245,19 @@ function oe_theme_preprocess_form_element(array &$variables): void {
  * Add paragraph formatting to body.
  */
 function oe_theme_preprocess_field(&$variables, $hook): void {
+  // Add space between field items of
+  // 'Entity reference revisions' field type (paragraphs for example).
+  // We have to avoid adding of margins for a nested paragraphs.
+  // For more info @see https://v1--europa-component-library.netlify.com/ec/components/detail/ec-utility-margin
```
nagyad: Add margin between paragraph items.

nagyad (changes requested):
image
As you can see on the image the content-row paragraph items are not containing the correct spacing.

nagyad (approved):

imanoleguskiza (approved):

PR #286, opened 2019-08-02
OPENEUROPA-1983: Update video banner to ECL2.

imanoleguskiza:
### Description

Update video banner to ECL 2

### Change log

- Added:
- Changed: Update video banner to ECL 2
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -673,10 +673,9 @@ function oe_theme_preprocess_pattern_accordion(array &$variables): void {
  */
 function oe_theme_preprocess_pattern_banner_video(array &$variables): void {
   if (!empty($variables['video']['sources'])) {
-    // We need to rename the variables that the pattern module uses.
+    // We need to rename the variables that the Patterns module uses.
     foreach ($variables['video']['sources'] as &$source) {
-      $source['src'] = $source['source'];
-      $source['type'] = $source['video_type'];
+      $source['type'] = $source['src_type'];
     }
   }
 }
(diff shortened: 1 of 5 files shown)


oe_theme.theme
```diff
@@ -668,6 +668,19 @@ function oe_theme_preprocess_pattern_accordion(array &$variables): void {
   }
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_pattern_banner_video(array &$variables): void {
```
voidtek: We should add in this function the way to change a single source to html5.
we change the video field of banner_video. we should create an 'upgrade path'.

imanoleguskiza: In this case there is no upgrade path as such. There is no actual implementation of any banner so there is nothing to update but the pattern. If there were components out there using this pattern they would need to deal with the update themselves since we are deealing with a major upgrade.

voidtek: Ok I agree.
thanks

templates/patterns/banners/banner_video.ui_patterns.yml
```diff
@@ -5,14 +5,16 @@ banner_video:
...
       label: "Video"
-      description: "Banner video"
+      description: "Banner video: uses the media container parameters as defined by the ECL"
       preview:
-        src: "https://ec.europa.eu/avservices/play.cfm?ref=I101631"
-        is_iframe: "true"
-        ratio: "16-9"
-        caption: "Neque porro quisquam est qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit"
+        sources:
+          - source: "https://clips.vorwaerts-gmbh.de/big_buck_bunny.mp4"
```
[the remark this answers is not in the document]
voidtek: thanks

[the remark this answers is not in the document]
voidtek: Thanks.

PR #290, opened 2019-08-06
OPENEUROPA-2088: ECL 2.0: Inpage Navigation.

sergepavle:
### Description
https://ec.europa.eu/component-library/ec/components/navigation/inpage-navigation/usage/

### DIFF
--- a/templates/paragraphs/paragraph--oe-content-row--variant-inpage_navigation.html.twig
+++ b/templates/paragraphs/paragraph--oe-content-row--variant-inpage_navigation.html.twig
@@ -9,7 +9,7 @@
 {{ attach_library('oe_theme/inpage_navigation') }}
 <div class="ecl-row">
   <div class="ecl-col-md-3 ecl-u-z-navigation">
-    {% include '@ecl/ec-component-inpage-navigation' with {
+    {% include '@ecl/inpage-navigation' with {
       'title': title,
       'links': links
     } %}
--- a/templates/paragraphs/paragraph--oe-content-row--variant-inpage_navigation.html.twig
+++ b/templates/paragraphs/paragraph--oe-content-row--variant-inpage_navigation.html.twig
@@ -7,14 +7,16 @@
  */
 #}
 {{ attach_library('oe_theme/inpage_navigation') }}
-<div class="ecl-row">
-  <div class="ecl-col-md-3 ecl-u-z-navigation">
-    {% include '@ecl/inpage-navigation' with {
-      'title': title,
-      'links': links
-    } %}
-  </div>
-  <div class="ecl-col-md-9">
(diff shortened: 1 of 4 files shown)


tests/Kernel/Paragraphs/ContentRowTest.php
```diff
@@ -17,6 +17,7 @@ class ContentRowTest extends ParagraphsTestBase {
    * Tests the rendering of the paragraph type.
    *
    * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
+   * @todo Update (restore) after porting to ECL2 other components.
```
voidtek: Can you add the number of the ticket?

sergepavle: Thanks. I have added mentioning of related tickets.

templates/paragraphs/paragraph--oe-content-row--variant-inpage_navigation.html.twig
```diff
@@ -9,7 +9,7 @@
 {{ attach_library('oe_theme/inpage_navigation') }}
 <div class="ecl-row">
```
voidtek: This template doesn't seem the same of https://ecl-twig.netlify.com/ec/?path=/story/components-navigation-inpage-navigation--default

tests/Kernel/Paragraphs/ContentRowTest.php
```diff
@@ -17,6 +17,7 @@ class ContentRowTest extends ParagraphsTestBase {
    * Tests the rendering of the paragraph type.
    *
    * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
+   * @todo Update (restore) after porting to ECL2 other components: Link blocks (OPENEUROPA-2092), List item (OPENEUROPA-1989, OPENEUROPA-1990, OPENEUROPA-2124).
    */
   public function testRendering(): void {
```
voidtek: This test is not added on `drone.yml`

templates/paragraphs/paragraph--oe-content-row--variant-inpage_navigation.html.twig
sergepavle: I have added wrapper (ecl-container).
Difference still exist but it is related to many aspects of integration. 
For instance we adding id for anchor for div wrapper (not for H1 tag as in example).
Also for many paragraphs we don't usage h1.
But in general inpage navigation currently work without big difference from demo (or I miss something).

tests/Kernel/Paragraphs/ContentRowTest.php
sergepavle: Fixed (sorry for that).

voidtek:
thanks for the changes.

PR #292, opened 2019-08-08
OPENEUROPA-1990: ECL 2.0: List item blocks.

voidtek:
### Description

Update list-item-block system.

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -908,33 +908,6 @@ function oe_theme_preprocess_pattern_file_translation__preview(array &$variables
 /**
  * Implements hook_preprocess_HOOK().
  */
-function oe_theme_preprocess_pattern_list_item_block_one_column(array &$variables): void {
-  foreach ((array) $variables['items'] as $key => $value) {
-    $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-12';
-  }
-}
-
-/**
- * Implements hook_preprocess_HOOK().
- */
-function oe_theme_preprocess_pattern_list_item_block_two_columns(array &$variables): void {
-  foreach ((array) $variables['items'] as $key => $value) {
-    $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-6';
-  }
-}
-
-/**
- * Implements hook_preprocess_HOOK().
- */
-function oe_theme_preprocess_pattern_list_item_block_three_columns(array &$variables): void {
-  foreach ((array) $variables['items'] as $key => $value) {
-    $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-4';
-  }
-}
-
-/**
- * Implements hook_preprocess_HOOK().
- */
 function oe_theme_preprocess_pattern_list_item__variant_date__preview(array &$variables): void {
   $variables['date'] = DateValueObject::fromArray($variables['date']);
 }
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -908,6 +908,34 @@ function oe_theme_preprocess_pattern_file_translation__preview(array &$variables
 /**
  * Implements hook_preprocess_HOOK().
  */
+function oe_theme_preprocess_pattern_list_item_block_one_column(array &$variables): void {
+  foreach ((array) $variables['items'] as $key => $value) {
+    $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-12';
+  }
+  var_dump($variables['items']);
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_pattern_list_item_block_two_columns(array &$variables): void {
+  foreach ((array) $variables['items'] as $key => $value) {
+    $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-6';
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_pattern_list_item_block_three_columns(array &$variables): void {
+  foreach ((array) $variables['items'] as $key => $value) {
+    $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-4';
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
 function oe_theme_preprocess_pattern_list_item__variant_date__preview(array &$variables): void {
   $variables['date'] = DateValueObject::fromArray($variables['date']);
 }
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -912,7 +912,6 @@ function oe_theme_preprocess_pattern_list_item_block_one_column(array &$variable
   foreach ((array) $variables['items'] as $key => $value) {
     $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-12';
   }
-  var_dump($variables['items']);
 }
 
 /**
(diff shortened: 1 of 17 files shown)


oe_theme.theme
```diff
@@ -932,6 +932,33 @@ function oe_theme_preprocess_pattern_file_translation__preview(array &$variables
...
+    $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-6';
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_pattern_list_item_block_three_columns(array &$variables): void {
+  foreach ((array) $variables['items'] as $key => $value) {
+    $variables['items'][$key]['#fields']['extra_classes'] = 'ecl-col-4';
```
ademarco: `extra_classes` is not defined as a pattern field so it will be ignored. Let's not do this via but directly in the Twig template.

ademarco (changes requested):

oe_theme.theme
voidtek: I changed 3 times to way to handle that. Now it's working.

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
```diff
@@ -0,0 +1,93 @@
...
+{# Internal properties #}
+
+{% set _content_item_block = {
+  type: 'one',
+  title: '',
+  items: [],
+  button: {}
+} %}
+
+{% if content_item_block is defined %}
```
ademarco: We don't need to do this in twig: ` is defined`, the code will be compiled in the right PHP code, let's remove them all, it clutters the code.

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
```diff
@@ -0,0 +1,93 @@
...
+  {% set _content_item_block = _content_item_block|merge(content_item_block) %}
+{% endif %}
+
+{% set _css_class = 'ecl-content-item-block ecl-container' %}
+{% set _extra_attributes = '' %}
+
+{# Internal logic - Process properties #}
+
+{% if extra_classes is defined and extra_classes is not empty %}
```
ademarco: Only this is enough:
```
 {% if extra_classes %}
```
this Twig code will be compiled in the right PHP code, let's remove them all, it clutters the code.

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
```diff
@@ -0,0 +1,93 @@
...
+  {% set _css_class = _css_class ~ ' ' ~ extra_classes %}
+{% endif %}
+
+{% if extra_attributes is defined and extra_attributes is not empty and extra_attributes is iterable %}
```
ademarco: Same here, we can remove the ` is iterable` too.

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
```diff
@@ -0,0 +1,93 @@
...
+{% endif %}
+
+{% set _extra_classes_item = 'ecl-col-12' %}
+{% if _content_item_block.type == 'two' %}
+  {% set _extra_classes_item = 'ecl-col-6' %}
+{% elseif _content_item_block.type == 'three' %}
+  {% set _extra_classes_item = 'ecl-col-4' %}
+{% endif %}
+
+{# Print the result #}
```
ademarco: I wouldn't put this type of comments, the purpose of a Twig template is to print out things.

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
```diff
@@ -0,0 +1,93 @@
...
+#}
+
+{# Internal properties #}
+
+{% set _content_item_block = {
+  type: 'one',
+  title: '',
+  items: [],
+  button: {}
+} %}
```
ademarco: I would remove this object, let's just use the properties directly, don't worry about defining the defaults globally, we can define them when we use the property with the `|default(...)` Twig filter.

templates/patterns/list_item_blocks/pattern-list-item-block-one-column.html.twig
```diff
@@ -4,31 +4,14 @@
...
+{% include '@oe_theme/compositions/ec-component-content-item-block/content-item-block.html.twig' with {
+  content_item_block : {
+    type: type|default('one'),
+    title: title,
+    items: items,
+    button: {
+      path: button_url,
+      label: button_label
+    }
+  }
```
ademarco: As said above, let's make this simpler by just passing the inner properties, without wrapping them into `content_item_block`. Also we just do `type: 'one',`.

ademarco (changes requested):
Good job! I've left some comments that are about simplify the Twig part a little.

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
voidtek: :+1:

templates/patterns/list_item_blocks/pattern-list-item-block-one-column.html.twig
voidtek: I removed the `content_item_block`, but I don't remove `type: 'one',` because this template is one two and three columns.

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
voidtek: :+1:

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
voidtek: ok :+1:

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
voidtek: Done :+1:

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
voidtek: I agree with you, but this type of comments is to be consistent with the template created by ecl-twig team.

templates/patterns/list_item_blocks/pattern-list-item-block-one-column.html.twig
ademarco: Yeah but we are in the `-one-colum.html.twig` template, we set this value here as `one`, it cannot come from outside.

ademarco (changes requested):

templates/compositions/ec-component-content-item-block/content-item-block.html.twig
voidtek: done.

ademarco:
Tested locally, it works but it looks uneven but that will be fixed on the ECL side.

image

ademarco (dismissed):

templates/patterns/list_item_blocks/pattern-list-item-block-one-column.html.twig
voidtek: I changed the template, according to our conversation.

templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -26,22 +23,15 @@
...
-
 {% set _css_class = 'ecl-content-item ecl-u-d-flex ecl-u-border-bottom ecl-u-border-color-grey-15 ecl-u-pv-m' %}
 {% set _extra_attributes = '' %}
+{% set _type = type|default('default') %}
+{% set _description = description|default('') %}
+{% set _meta = meta|default([]) %}
+{% set _title = title|default({}) %}
+{% set _image = image|default('') %}
+{% set _image_position = image_position|default('before') %}
+{% set _links = links|default([]) %}
```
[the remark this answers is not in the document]
voidtek: :+1:

imanoleguskiza (dismissed):

ademarco (approved):

imanoleguskiza (approved):

Issue #284, opened 2019-08-01
Respect the weight of the Local Tasks


skounis:
When the Local Tasks are preprocessed the new arrays that are created ignore the `weight` property of each tasks. As a results the items in the `$variables['primary_links']` and `$variables['secondary_links']` are having the order they are created with and not the order they should be rendered. 

- https://github.com/openeuropa/oe_theme/blob/master/oe_theme.theme#L154

In Bartik the rendering of the local tasks happens in the `menu-local-tasks.html.twig` [1] file of `classy`: 

Here each array with the links (eg `primary`) is again sent for rendering 
```
 <ul class="tabs primary">{{ primary }}</ul>
```
This triggers the rendering process of drupal/twig and before the `menu-local-task.html.twig`[2] gets the control the array is sorted in the `Renderer.php` [3]

### References
1. https://git.drupalcode.org/project/drupal/blob/8.8.x/core/themes/classy/templates/navigation/menu-local-tasks.html.twig#L16
2. https://git.drupalcode.org/project/drupal/blob/8.8.x/core/themes/classy/templates/navigation/menu-local-task.html.twig 
3. https://git.drupalcode.org/project/drupal/blob/8.8.x/core/lib/Drupal/Core/Render/Renderer.php#L401

## Suggested solutions
We can sort the array in a method in the `oe_theme.theme` file or move this methid in the `oe_theme_helper` modules.

skounis:
I the clone of `oe_theme` we use for `futurium` we solved this with: 

```
/**
 * Compares two objects based on their '#weight' property
 *
 * @param $item1
 * @param $item2
 * @return bool|void
 */
function _blellow_weight_comparator($item1, $item2) {
  if (!isset($item1, $item2)) {
    return 0;
  }
  if (!isset($item1['#weight'], $item2['#weight'])) {
    return 0;
  }
  return $item1['#weight'] > $item2['#weight'];
}

/**
 * Sorts an array of objects without mutating the array.
 *
 * @param $array
 * @param $comparator
 * @return mixed
 */
function _blellow_sort($array, $comparator) {
  usort($array, $comparator);
  return $array;
}
``` 
And in `blellow_preprocess_menu_local_tasks()`
```
  $sorted = _blellow_sort($variables['primary'], '_blellow_weight_comparator');
  foreach ($sorted as $link) {
    /** @var Drupal\Core\Url $url */
    $url = $link['#link']['url'];
    if ($url->access($variables['user'])) {
[...]

PR #294, opened 2019-08-12
feat(theme) Sort local tasks when preprocess them #284

skounis:
### Description

Respect the weight of the Local Tasks #284

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -156,6 +156,7 @@ function oe_theme_preprocess_paragraph__oe_contextual_navigation(array &$variabl
  * Implements hook_preprocess_menu_local_tasks().
  */
 function oe_theme_preprocess_menu_local_tasks(array &$variables): void {
+  uasort ($variables['primary'], '_oe_theme_weight_comparator');
   foreach ($variables['primary'] as $link) {
     /** @var Drupal\Core\Url $url */
     $url = $link['#link']['url'];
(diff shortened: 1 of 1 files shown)


oe_theme.theme
```diff
@@ -156,6 +156,7 @@ function oe_theme_preprocess_paragraph__oe_contextual_navigation(array &$variabl
  * Implements hook_preprocess_menu_local_tasks().
  */
 function oe_theme_preprocess_menu_local_tasks(array &$variables): void {
+  uasort ($variables['primary'], '_oe_theme_weight_comparator');
```
ademarco: We should use `SortArray::sortByWeightElement` instead, check docs [here](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21SortArray.php/function/SortArray%3A%3AsortByWeightElement/8.2.x).

ademarco (changes requested):

oe_theme.theme
skounis: @ademarco thank you for the hint, due to the `#weight` property we finally needed [`SortArray::sortByWeightProperty`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21SortArray.php/function/SortArray%3A%3AsortByWeightProperty/8.2.x)

The changes is made.

PR #300, opened 2019-08-19
OPENEUROPA-2018: Add remaining form elements - Checkbox

voidtek:
### Description
Add remaining form elements - Checkbox

### DIFF
--- a/templates/compositions/ec-component-checkbox/checkbox.html.twig
+++ b/templates/compositions/ec-component-checkbox/checkbox.html.twig
@@ -78,7 +78,7 @@
     value="{{ _value }}"
     {{ _disabled ? "disabled" }}
     {{ _checked ? "checked" }}
-    {{ _helper_id is not empty ? "aria-describedby=\"" ~ _helper_id ~ "\"" -}}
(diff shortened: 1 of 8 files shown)


templates/compositions/ec-component-checkbox/checkbox.html.twig
```diff
@@ -0,0 +1,110 @@
...
+<div class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+  <input
+    id="{{ _id }}"
+    name="{{ _name }}"
+    class="ecl-checkbox__input"
+    type="checkbox"
+    value="{{ _value }}"
+    {{ _disabled ? "disabled" }}
+    {{ _checked ? "checked" }}
+    {{ _helper_id is not empty ? "aria-describedby=\"" ~ _helper_id ~ "\"" -}}
```
sergepavle: May be it should be like:
```suggestion
    {{ _helper_id is not empty ? "aria-describedby=\"" ~ _helper_id ~ "\"" }}
```
?

voidtek: Good catch

PR #308, opened 2019-08-27
Add sidebar to page template

haringsrob:
provides a left sidebar as per the markup here: https://ec.europa.eu/component-library/playground/ec/?path=/story/templates-pages--search

### DIFF
--- a/templates/layout/page.html.twig
+++ b/templates/layout/page.html.twig
@@ -7,6 +7,8 @@
  */
 #}
 
+{% set _main_content_class = page.sidebar ? 'ecl-col-12 ecl-col-lg-9' : 'ecl-col' %}
+
 {% block site_header %}
 <header class="ecl-site-header">
   {{ page.site_top_bar }}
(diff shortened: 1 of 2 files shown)


haringsrob:
Tagged with ecl2 because the target is 2.x

templates/layout/page.html.twig
```diff
@@ -33,7 +35,12 @@
 <section class="main-content ecl-u-mv-l">
   <div class="ecl-container">
     <div class="ecl-row">
-      <div class="ecl-col">
+      {% if page.sidebar %}
+        <div class="ecl-col-12 ecl-col-lg-3">
```
voidtek: According to https://ec.europa.eu/component-library/ec/templates/pages/search/code/ this div should be an `aside` and we should change the main by a `section`.

ademarco:
I would add a sidebar in the main theme. Still this is a good occasion to improve our theme to be more extensible by sub-themes (which, I guess, you are using @haringsrob).

In order to do that we could maybe rework this PR using [Twig blocks](https://twig.symfony.com/doc/1.x/tags/block.html) on the page template, so that you could simply override a `content` block in your page template that extends our.

If you are ok with this approach please consider updating the documentation too.

ademarco (changes requested):
See comment above.

haringsrob:
Can be done by

PR #311, opened 2019-09-06
OPENEUROPA-2280: Not correct spacing between "Listing item block" paragraph.

voidtek:
### Description

Revert wrapper and add extra classes where is needed.


templates/field/field--paragraph--field-oe-text-long.html.twig
```diff
@@ -6,6 +6,6 @@
  * @see ./core/themes/stable/templates/field/field.html.twig
  */
 #}
-<div class="ecl-editor">
```
imanoleguskiza: This is a bit of a problem because oe-text-long is also used inside components (for example, inside the accordion item), which adds extra margins inside them, making them bigger than they should

voidtek: Good catch

voidtek: done

oe_theme.theme
```diff
@@ -79,11 +79,6 @@ function oe_theme_theme_suggestions_field_alter(array &$suggestions, array $vari
     // fields, while keeping all the rest markup-free.
     array_unshift($suggestions, 'field__bare');
   }
-
-  // Add suggestions for all 'Entity reference revisions' for each target type.
-  if ($variables['element']['#field_type'] === 'entity_reference_revisions') {
-    $suggestions[] = 'field__' . $variables['element']['#field_type'] . '__' . $variables['element']['#items']->getFieldDefinition()->getSetting('target_type');
-  }
```
ademarco: We can't remove this one, even if we don't use it anymore, since sites might use this suggestion to style their own things.

templates/field/field--entity-reference-revisions--paragraph.html.twig
```diff
@@ -1,13 +0,0 @@
...
- * Theme override for the entity reference revisions field wich allowed to use paragraphs.
- *
- * @see ./templates/field/field--bare.html.twig
- */
-#}
-{% for item in items %}
-  <div class="ecl-u-mv-s">
-    {{ item.content }}
-  </div>
-{% endfor %}
```
ademarco: Why the decision to remove this? It looked quite convenient to have them applied to all paragraphs, present and future.

oe_theme.theme
voidtek: fair enough.

templates/field/field--entity-reference-revisions--paragraph.html.twig
voidtek: The goal of this template is to add `<div class="ecl-u-mv-s">` foreach item.content, but it's the source of bug of this ticket.
we could keep this template without the div, it seems duplicate code, is it not?

ademarco:
I'm not sure about this implementation, I've re-ran the steps in the ticket and indeed I end up with this situation:

image

This is because we indiscriminately apply that class to every `entity-reference-revisions--paragraph` field, which includes the listing items in the screenshot above.

I'd rather see an implementation that removes that class for specific cases, rather than having to chase down all cases where we need it.

What do you think @voidtek ? I'll continue my investigation to look for alternative solutions.

ademarco:
Ok so I've managed to achieve the result below by changing this:

```php
  // Add suggestions for all 'Entity reference revisions' for each target type.
  if ($variables['element']['#field_type'] === 'entity_reference_revisions') {
    /** @var \Drupal\field\Entity\FieldConfig $field_definition */
    $field_definition = $variables['element']['#items']->getFieldDefinition();
    $suggestions[] = 'field__' . $field_definition->getType() . '__' . $field_definition->getSetting('target_type');
    $suggestions[] = 'field__' . $field_definition->getType() . '__' . $field_definition->getSetting('target_type') . '__' . $field_definition->getName();
  }
```

And by adding the following template file:
```
field--entity-reference-revisions--paragraph--field-oe-paragraphs.html.twig
```

```twig
{% for item in items %}
  {{ item.content }}
{% endfor %}
```


image

ademarco:
Closed in favor of #326

PR #315, opened 2019-09-06
OPENEUROPA-2243: Remove site switcher from oe_theme 1.x

voidtek:
### Description

Remove the block site switcher.


ecl-builder.config.js
```diff
@@ -84,7 +84,6 @@ module.exports = {
     { from: path.resolve(nodeModules, '@ecl'), patterns: 'ec-component-navigation-menu/*.twig', to: path.resolve(outputFolder, 'templates/components') },
     { from: path.resolve(nodeModules, '@ecl'), patterns: 'ec-component-page-header/*.twig', to: path.resolve(outputFolder, 'templates/components') },
     { from: path.resolve(nodeModules, '@ecl'), patterns: 'ec-component-pager/*.twig', to: path.resolve(outputFolder, 'templates/components') },
-    { from: path.resolve(nodeModules, '@ecl'), patterns: 'ec-component-site-switcher/*.twig', to: path.resolve(outputFolder, 'templates/components') },
```
imanoleguskiza: We don't get the ec-component-site-switcher template from ECL anymore but the drupal block template still uses it (see https://github.com/openeuropa/oe_theme/blob/master/templates/navigation/oe-corporate-blocks-site-switcher.html.twig#L13). This means that if I manually add the site switcher block to the site it will crash.

voidtek: Good catch. :D

PR #319, opened 2019-09-09
OPENEUROPA-2326: Add tabs and sub-tabs.

voidtek:
### Description

Add menu local component.


oe_theme.theme
```diff
@@ -65,8 +65,10 @@ function oe_theme_preprocess_menu__main(array &$variables): void {
   // Massage data to be compliant with ECL navigation menu data structure.
   $variables['links'] = array_map(function ($item) {
     return [
-      'href' => $item['url'],
-      'label' => $item['title'],
+      'link' => [
```
imanoleguskiza: This change seems out of scope. In fact the menu--main template uses a component that is no longer there in ecl2. So probably we should change this together with the cleanup of the menu template.

oe_theme.theme
```diff
@@ -75,8 +77,10 @@ function oe_theme_preprocess_menu__main(array &$variables): void {
     $variables['links'][$name]['children_links'] = array_map(function ($item) {
       return [
         'links' => [[
-          'href' => $item['url'],
-          'label' => $item['title'],
+          'link' => [
```
imanoleguskiza: Same as above.

templates/compositions/ec-component-navigation/navigation-list.html.twig
```diff
@@ -0,0 +1,49 @@
+{#
```
imanoleguskiza: I think we should change the name of this composition. It is not the navigation-list composition as defined in the ECL but rather our own implementation. Why not call like using the Drupal naming convention: local-tasks?

imanoleguskiza (changes requested):

oe_theme.theme
voidtek: Fair enough, I remove.

oe_theme.theme
voidtek: Same as above. :p

templates/compositions/ec-component-navigation/navigation-list.html.twig
voidtek: Fair enough.

sergepavle (dismissed):

sergepavle (dismissed):

templates/compositions/ec-component-navigation/local-tasks.html.twig
```diff
@@ -0,0 +1,59 @@
...
+
+{% if _links is not empty and _links is iterable %}
+  <nav class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+    <h2 hidden>{{ _title }}</h2>
+    <ul class="ecl-navigation-list ecl-unordered-list--no-bullet ecl-u-border-bottom ecl-u-border-color-grey-25{{ _list_class }}">
+      {% for _link in _links %}
+        {% set _active_class = _link.is_active is defined and _link.is_active == true ? ' ecl-navigation-list__item--active ecl-u-border-bottom ecl-u-border-color-blue' : '' %}
+        <li class="ecl-navigation-list__item ecl-unordered-list__item ecl-u-d-inline{{ _list_item_class }}{{ _active_class }}">
+          {% include '@ecl/link' with {
+          }|merge(_link|default({})) only %}
```
ademarco: Doesn't this work too?
```
{% include '@ecl/link' with _link only %}
```

templates/compositions/ec-component-navigation/local-tasks.html.twig
```diff
@@ -0,0 +1,59 @@
...
+
+{% if _variant == 'primary' %}
+  {% set _list_class = ' ecl-u-pb-m ecl-u-type-bold' %}
+  {% set _list_item_class = ' ecl-u-pa-m' %}
+{% endif %}
+{% if _variant == 'secondary' %}
+  {% set _list_class = ' ecl-u-pb-s ecl-u-type-s' %}
+  {% set _list_item_class = ' ecl-u-pa-s' %}
+{% endif %}
+{% if extra_classes is defined and extra_classes is not empty %}
```
ademarco: We don't need to "worry too much" if things are not set, defined etc. in Twig, the language parser should take care of that.

Let's just do ` {% if extra_classes is not empty %}`.

templates/compositions/ec-component-navigation/local-tasks.html.twig
```diff
@@ -0,0 +1,59 @@
...
+  {% set _css_class = _css_class ~ ' ' ~ extra_classes %}
+{% endif %}
+
+{% if extra_attributes is defined and extra_attributes is not empty and extra_attributes is iterable %}
```
ademarco: Same as above.

templates/compositions/ec-component-navigation/local-tasks.html.twig
```diff
@@ -0,0 +1,59 @@
...
+  {% for attr in extra_attributes %}
+    {% set _extra_attributes = _extra_attributes ~ ' ' ~ attr.name ~ '="' ~ attr.value ~ '"' %}
+  {% endfor %}
+{% endif %}
+
+{# Print the result #}
+
+{% if _links is not empty and _links is iterable %}
```
ademarco: Same as above.

templates/compositions/ec-component-navigation/local-tasks.html.twig
```diff
@@ -0,0 +1,59 @@
...
+  <nav class="{{ _css_class }}"{{ _extra_attributes|raw }}>
+    <h2 hidden>{{ _title }}</h2>
+    <ul class="ecl-navigation-list ecl-unordered-list--no-bullet ecl-u-border-bottom ecl-u-border-color-grey-25{{ _list_class }}">
+      {% for _link in _links %}
+        {% set _active_class = _link.is_active is defined and _link.is_active == true ? ' ecl-navigation-list__item--active ecl-u-border-bottom ecl-u-border-color-blue' : '' %}
```
ademarco: Same as above, `... = _link.is_active ? ...` is enough.

tests/Kernel/MenuLocalTasksTest.php
```diff
@@ -9,8 +9,6 @@
 
 /**
  * Class MenuLocalTasks.
```
ademarco: Add meaningful comment here.

ademarco (changes requested):

sergepavle (dismissed):

nagyad (dismissed):

ademarco (approved):

imanoleguskiza (approved):

PR #321, opened 2019-09-13
OPENEUROPA-2083: ECL 2.0: Use extra_blocks on form.

voidtek:
### Description

Update search form.

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -16,6 +16,7 @@ use Drupal\oe_theme\ValueObject\FileValueObject;
 use Drupal\oe_theme\ValueObject\DateValueObject;
 use Drupal\oe_theme\ValueObject\GalleryItemValueObject;
 use Drupal\oe_theme_helper\EuropeanUnionLanguages;
+use Drupal\Core\Template\Attribute;
 
 /**
  * Implements hook_preprocess().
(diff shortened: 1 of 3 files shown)


oe_theme.theme
```diff
@@ -940,6 +940,17 @@ function oe_theme_preprocess_form__oe_search_search_form(array &$variables): voi
   $variables['input_array'] = _oe_theme_preprocess_search_input_text($variables['element']['keys']);
   $variables['input_array']['hide_label'] = TRUE;
   $variables['label'] = $variables['input_array']['label'] = t('Search');
+
+  $variables['extra_attributes'] = [];
+  foreach ($variables['attributes'] as $key => $value) {
+    if (is_string($value)) {
```
ademarco: As per parent preprocess `template_preprocess_form()` values on `$variables['attributes']` array are always strings, so we can also not check this with `if (is_string($value))`.

voidtek: fair enough.

oe_theme.theme
```diff
@@ -955,28 +955,14 @@ function oe_theme_preprocess_form__oe_search_search_form(array &$variables): voi
...
     'type' => $element['#attributes']['type'],
     'class' => $element['#attributes']['class'],
   ];
-  $ecl_array['class'][] = 'ecl-search-form__text-input';
-
-  $extra_attributes = !empty($element['#attributes']) ? $element['#attributes'] : [];
-  $extra_attributes = array_diff_key($extra_attributes, $ecl_array);
-
-  $ecl_array['extra_attributes'] = [];
-  foreach ($extra_attributes as $key => $value) {
```
imanoleguskiza: We are removing the extra_attributes from the text_input here, is this intentional?

imanoleguskiza: Apparently the extra_attributes on the text_input have no effect so removing them doesn't affect anything.

PR #324, opened 2019-09-17
OPENEUROPA-2258: Add php 7.2 on CI.

voidtek:
### Description

Add php 7.2 on CI.


yenyasinn:
We don't support PHP 7.2 anymore

PR #325, opened 2019-09-18
Add "Tag" pattern

haringsrob:
### Description

Simple implementation for the tag component

### Change log

- Added: Tag pattern

### DIFF
--- a/templates/patterns/tag/tag.ui_patterns.yml
+++ b/templates/patterns/tag/tag.ui_patterns.yml
@@ -1,6 +1,10 @@
 tag:
   label: "Tag"
   description: "The tag component is used to display structured information."
+  variants:
+    default:
+      label: "Default"
+      description: "Displays the url and label in a tag."
   fields:
     tag_type:
       type: "text"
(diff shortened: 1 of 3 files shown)


templates/patterns/tag/tag.ui_patterns.yml
```diff
@@ -0,0 +1,19 @@
...
+    tag_type:
+      type: "text"
+      label: "Type"
+      description: "Type of tag (link, button, removable)"
+      preview: "link"
+    path:
+      type: "text"
+      label: "Path"
+      description: "Tag path"
+      preview: "#"
```
ademarco: If this is the tag link URL I'd call it `url` instead of `path` as it might contain a full URL, other than a relative path like `/something`.

ademarco:
After checking the ECL documentation it seems that there are two tag variants, link and removable link:
image

I suggest to provide only one variant, the link one, in the scope of this PR, this will allow us to add the second one when we need it and still be backward compatible with what you need in your project.

PR #326, opened 2019-09-18
OPENEUROPA-2280: Not correct spacing between "Listing item block" paragraph

voidtek:
### Description

Not correct spacing between "Listing item block" paragraph


oe_theme.theme
```diff
@@ -82,7 +82,10 @@ function oe_theme_theme_suggestions_field_alter(array &$suggestions, array $vari
 
   // Add suggestions for all 'Entity reference revisions' for each target type.
   if ($variables['element']['#field_type'] === 'entity_reference_revisions') {
-    $suggestions[] = 'field__' . $variables['element']['#field_type'] . '__' . $variables['element']['#items']->getFieldDefinition()->getSetting('target_type');
+    /** @var \Drupal\field\Entity\FieldConfig $field_definition */
+    $field_definition = $variables['element']['#items']->getFieldDefinition();
+    $suggestions[] = 'field__' . $field_definition->getType() . '__' . $field_definition->getSetting('target_type');
+    $suggestions[] = 'field__' . $field_definition->getType() . '__' . $field_definition->getSetting('target_type') . '__' . $field_definition->getName();
```
ademarco: We can actually start each with:
```php
$suggestions[] = 'field__entity_reference_revisions__' . ...
```
Since we are inside that if, it will make everything more readable.

templates/field/field--entity-reference-revisions--paragraph--field-oe-paragraphs.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
+/**
+ * @file
+ * Theme override for the entity reference revisions field wich allowed to use paragraphs.
+ *
+ * @see ./templates/field/field--bare.html.twig
+ */
+#}
+{%- include 'field--bare.html.twig' -%}
```
ademarco: Not sure if this is not too broad, maybe we should have the following suggestions and pick the post restrictive one:

We can use these to complete the chain prefixed by TARGET_TYPE: :point_down: 
```
field__entity_reference_revisions__TARGET_TYPE
field__entity_reference_revisions__TARGET_TYPE__TARGET_BUNDLE
field__entity_reference_revisions__TARGET_TYPE__TARGET_BUNDLE__FIELD_NAME
```
But then we also add those that have precedence over those above:
```
field__entity_reference_revisions__PARENT_ENTITY_TYPE
field__entity_reference_revisions__PARENT_ENTITY_TYPE__PARENT_ENTITY_BUNDLE
field__entity_reference_revisions__PARENT_ENTITY_TYPE__PARENT_ENTITY_BUNDLE__FIELD_NAME
```

And pick up the last one: 

```
field__entity_reference_revisions__PARENT_ENTITY_TYPE__PARENT_ENTITY_BUNDLE__FIELD_NAME
```

Does it make sense?

voidtek: I understand your point.
I tested all the components and configurations, and it seems to work.

PR #328, opened 2019-09-20
OPENEUROPA-2096: ECL 2.0: Table

voidtek:
### Description

ECL 2.0: Table


templates/layout/page.html.twig
```diff
@@ -33,7 +33,7 @@
 {% endblock %}
 
 {% block content %}
-<main class="ecl-u-pv-xl" id="main-content">
```
imanoleguskiza: What is the reasoning behind this?

templates/status/status-messages.html.twig
```diff
@@ -25,6 +25,7 @@
     'description': _self.message_list(messages),
     'variant': (type == 'status' ? 'success' : type),
     'extra_attributes': extra_attributes,
+    'extra_classes': 'ecl-u-mv-l',
```
imanoleguskiza: I assume this is to ofset the class we removed on page.html, but why was that done?

oe_theme.theme
```diff
@@ -1045,3 +1045,41 @@ function oe_theme_preprocess_language_selection_page_content(&$variables) {
...
+ */
+function oe_theme_preprocess_block__oe_theme_main_page_content(&$variables) {
+  $variables['attributes']['class'][] = 'ecl-u-mv-l';
+}
+
+/**
+ * Implements hook_preprocess_table().
+ */
+function oe_theme_preprocess_table(array &$variables): void {
+  $variables['zebra'] = 'true';
```
imanoleguskiza: How come we are always setting the zebra to true?

templates/ui_patterns_library/patterns-overview-page.html.twig
```diff
@@ -24,35 +24,32 @@
   } %}
 
   {% for pattern_name, pattern in patterns %}
+    <div class="ecl-u-mv-l ecl-u-border-color-grey-25 ecl-u-border-bottom">
```
imanoleguskiza: All of the changes done in this file seem unrelated to the porting the table component.

oe_theme.theme
```diff
@@ -1045,3 +1045,41 @@ function oe_theme_preprocess_language_selection_page_content(&$variables) {
...
+
+  $name_col = [];
+  foreach ($variables['header'] as $key => &$cell) {
```
imanoleguskiza: Defining the $key variable is not required.

oe_theme.theme
```diff
@@ -1045,3 +1045,41 @@ function oe_theme_preprocess_language_selection_page_content(&$variables) {
...
+    $cell['label'] = $cell['content'];
+
+    if ($cell['attributes']['colspan']) {
+      $cell['colspan'] = (string) $cell['attributes']['colspan'];
+    }
+
+    for ($i = 1; $i <= (array_key_exists('colspan', $cell) ? intval($cell['colspan']) : 1); $i++) {
+      $cell_label = (array) $cell['label'];
+      $name_col[] = array_key_exists('#markup', $cell_label) ? $cell_label['#markup'] : $cell['label'];
```
imanoleguskiza: Can we add a small comment as to why we are doing this? Something like: "The contents of a drupal cell can be a render array so we need to make sure we have an string to set as the 'data-ecl-table-header' property later"

templates/layout/page.html.twig
voidtek: done

templates/status/status-messages.html.twig
voidtek: removed

templates/ui_patterns_library/patterns-overview-page.html.twig
voidtek: Removed.

oe_theme.theme
voidtek: Removed.

oe_theme.theme
voidtek: done.

oe_theme.theme
voidtek: done.

oe_theme.theme
```diff
@@ -1044,3 +1044,35 @@ function oe_theme_preprocess_language_selection_page_content(&$variables) {
...
+    if ($cell['attributes']['colspan']) {
+      $cell['colspan'] = (string) $cell['attributes']['colspan'];
+    }
+
+    for ($i = 1; $i <= (array_key_exists('colspan', $cell) ? intval($cell['colspan']) : 1); $i++) {
+      $cell_label = (array) $cell['label'];
+      // The contents of a drupal cell can be a render array so we need to make
+      // sure we have an string to set as the 'data-ecl-table-header' property
+      // later.
+      $name_col[] = array_key_exists('#markup', $cell_label) ? $cell_label['#markup'] : $cell['label'];
```
imanoleguskiza: I am still not sure about this approach. We can't really be sure that the cell content is going to be a simple markup render array. For all we know, this may be anything so I think we need to find a more open solution. I added a possible approach here: https://github.com/openeuropa/oe_theme/pull/332/files in which I render the array and extract all tags to end up with the actual text. Let me know what you think

PR #337, opened 2019-10-09
OPENEUROPA-2192: Use ECL background image attribute and document image styles.

imanoleguskiza:
### Description

Use ECL background image attribute and document image styles.

### Change log

- Added: Use ECL background image attribute and document image styles.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/templates/compositions/ec-component-content-item/content-item.html.twig
+++ b/templates/compositions/ec-component-content-item/content-item.html.twig
@@ -49,7 +49,7 @@
 
 {% macro image(src, alt, class1, class2) %}
   <div role="img" aria-label="{{ alt }}" class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl {{ class1 }}"
-       style="background-image:url('{{ src }}');background-size:cover;height:5rem;width:7.5rem"></div>
+       style="background-image:url('{{ src }}');background-size:cover;height:8.75rem;width:13.125rem"></div>
   <div role="img" aria-label="{{ alt }}"
        class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl ecl-u-d-none {{ class2 }}"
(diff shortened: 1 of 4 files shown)


templates/compositions/ec-component-content-item/content-item.html.twig
```diff
@@ -49,10 +49,10 @@
 
 {% macro image(src, alt, class1, class2) %}
   <div role="img" aria-label="{{ alt }}" class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl {{ class1 }}"
-       style="background-image:url('{{ src }}');background-size:contain;height:5rem;width:7.5rem"></div>
+       style="background-image:url('{{ src }}');background-size:cover;height:8.75rem;width:13.125rem"></div>
   <div role="img" aria-label="{{ alt }}"
        class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl ecl-u-d-none {{ class2 }}"
-       style="background-image:url('{{ src }}');background-size:contain;height:8.75rem;width:13.125rem">
+       style="background-image:url('{{ src }}');background-size:cover;height:8.75rem;width:13.125rem">
```
ademarco: I'm wondering what happens here if we have images using special characters like `image(1).png`, we should assess this.

imanoleguskiza: Manually tested on drupal and the resulting elements look like this:
<div role="img" aria-label="" class="ecl-u-ratio-3-2 ecl-u-flex-shrink-0 ecl-u-mr-xl ecl-u-d-lg-block ecl-content-item__image__before" style="background-image:url('/build/sites/default/files/2019-10/cat-ico%27s.png');background-size:cover;height:8.75rem;width:13.125rem"></div>
So Drupal takes care of the dangerous elements :)

PR #341, opened 2019-10-22
OPENEUROPA-2093: Port field pattern to ECL 2.x and add field_list pattern.

ademarco:


tests/Kernel/fixtures/rendering.yml
```diff
@@ -829,18 +829,39 @@
...
+        - label: "Live streaming"
+          body: "08:00 CET | 00d:23h:35m"
+  assertions:
+    count:
+      '.ecl-description-list.ecl-description-list--horizontal': 1
+      '.ecl-description-list__term': 3
+      '.ecl-description-list__definition': 3
+    equals:
+      '.ecl-description-list__term': "When"
+      '.ecl-description-list__definition': "Thursday 15 November, 08:00 AM to Friday 16 November"
```
ademarco: Let's add assertions also for the other two with child selector.

PR #348, opened 2019-10-24
OPENEUROPA-2430: Create an "Icons with text" pattern.

sergepavle:
### Description

We can offer this as a pattern and then use such a pattern as needed when theming content types.

We will implement this as a composition, i.e. we won't be asking ECL team to provide this as a component.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
+++ b/templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
@@ -4,40 +4,31 @@
   /**
    * @file
    * Theme implementation for a ECL list of icons with text.
+   *
+   * Parameters:
+   *   - "items" (array) (default: [])
+   *     - "icon" (string) (default: 'file')
+   *     - "text" (string) (default: '')
    */
-Parameters:
-- "items" (array) (default: [])
-  - "icon" (string) (default: '')
-  - "text" (string) (default: '')
 #}
 
 {# Internal properties #}
 
 {% set _icon_path = icon_path|default('') %}
-{% set _extra_classes = 'ecl-u-mr-s' %}
-
-{# Internal logic - Process properties #}
-
-{% set _items = [] %}
-{% for item in  items %}
-  {% if item.icon and item.text %}
-    {% set _items = _items|merge([item]) %}
-  {% endif %}
-{% endfor %}
 
 {# Print the result #}
 
-{% if _items %}
+{% if items %}
 <ul class="ecl-unordered-list ecl-unordered-list--no-bullet">
-  {% for _item in _items %}
-    <li class="ecl-u-d-flex ecl-u-align-items-center ecl-unordered-list__item{% if not loop.first %} ecl-u-mt-l{% endif %}">
+  {% for _item in items if _item.text %}
+    <li class="ecl-u-d-flex ecl-u-align-items-center ecl-unordered-list__item{{ not loop.first ? ' ecl-u-mt-l' }}">
       {% include '@ecl/icon' with {
         icon: {
           path: _icon_path,
           type: 'general',
-          name: _item.icon
+          name: _item.icon|default('file')
         },
-        extra_classes: _extra_classes
+        extra_classes: 'ecl-u-mr-s'
       } only %}{{- _item.text -}}
     </li>
   {% endfor %}
--- a/templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
+++ b/templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
@@ -18,19 +18,20 @@ Parameters:
 {% set _extra_classes = 'ecl-u-mr-s' %}
 
 {# Print the result #}
-{% if _items not empty %}
+
+{% if _items %}
 <ul class="ecl-unordered-list ecl-unordered-list--no-bullet">
   {% for _item in _items %}
-  <li class="ecl-u-d-flex ecl-u-align-items-center ecl-unordered-list__item{% if not loop.first %} ecl-u-mt-l{% endif %}">
-    {% include '@ecl/icon' with {
-      icon: {
-        path: _icon_path,
-        type: 'general',
-        name: _item.icon
-      },
-      extra_classes: _extra_classes
-    } only %}{{- _item.text -}}
-  </li>
+    <li class="ecl-u-d-flex ecl-u-align-items-center ecl-unordered-list__item{% if not loop.first %} ecl-u-mt-l{% endif %}">
+      {% include '@ecl/icon' with {
+        icon: {
+          path: _icon_path,
+          type: 'general',
+          name: _item.icon
+        },
+        extra_classes: _extra_classes
+      } only %}{{- _item.text -}}
+    </li>
   {% endfor %}
 </ul>
 {% endif %}
--- a/templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
+++ b/templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
@@ -1,0 +1,37 @@
+{% spaceless %}
+
+{#
+  /**
+   * @file
+   * Theme implementation for a ECL list of icons with text.
+   */
+Parameters:
+- "items" (array) (default: [])
+  - "icon" (string) (default: '')
+  - "text" (string) (default: '')
+#}
+
+{# Internal properties #}
+
+{% set _items = items|default([]) %}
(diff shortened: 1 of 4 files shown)


templates/patterns/icons_with_text/pattern-icons-with-text.html.twig
```diff
@@ -0,0 +1,10 @@
+{#
+/**
+ * @file
+ * A list of icons with a text.
+ */
+#}
+{% include '@oe_theme/compositions/ec-component-icons-with-text/icons-with-text.html.twig' with {
+  'items': items|default([]),
```
nagyad: Why do you provide default if you already provide this default in the included twig?

tests/Kernel/fixtures/rendering.yml
```diff
@@ -1371,3 +1371,36 @@
...
       '.ecl-timeline2__item:nth-child(5) .ecl-timeline2__content': "Lorem Ipsum 5"
       '.ecl-timeline2__item:nth-child(6) .ecl-timeline2__content': "Lorem Ipsum 6"
+- array:
+    '#type': pattern
+    '#id': icons_with_text
+    '#fields':
+      items:
+        -
+          icon: "file"
+          text: "Culture | Innovation"
```
nagyad: What will happen if don't provide one of these?

templates/patterns/icons_with_text/pattern-icons-with-text.html.twig
sergepavle: I agree. removed.

tests/Kernel/fixtures/rendering.yml
sergepavle: Updated behavior. If an icon or text is empty, we remove this list item.
@ademarco What do you think, is this correct behavior?

ademarco: I would not display the item only if the text is empty, while if the icon is empty we can default to the `document` one, so we always display one.

ademarco: I still don't see the fallback to `document` icon if none is specify, did we do this?

tests/Kernel/fixtures/rendering.yml
```diff
@@ -1456,3 +1456,59 @@
...
+          icon: "livestreaming"
+          text: "Live streaming available"
+  assertions:
+    count:
+      'ul.ecl-unordered-list--no-bullet': 1
+      'ul li.ecl-unordered-list__item': 2
+      'ul li.ecl-unordered-list__item svg': 2
+    equals:
+      'ul li.ecl-unordered-list__item:nth-child(1)': "Culture | Innovation"
+      'ul li.ecl-unordered-list__item:nth-child(2)': "Live streaming available"
```
ademarco: Assert also that we have a `document` icon if none is specified, in:
```
        -
          text: "15-16 November 2019, 08:00 AM"
```

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
```diff
@@ -0,0 +1,46 @@
...
+{% set _extra_classes = 'ecl-u-mr-s' %}
+
+{# Internal logic - Process properties #}
+
+{% set _items = [] %}
+{% for item in  items %}
+  {% if item.icon and item.text %}
+    {% set _items = _items|merge([item]) %}
+  {% endif %}
+{% endfor %}
```
ademarco: What does this do? Can't we simply loop trough `items` down below?

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
```diff
@@ -0,0 +1,46 @@
...
+
+{#
+  /**
+   * @file
+   * Theme implementation for a ECL list of icons with text.
+   */
+Parameters:
+- "items" (array) (default: [])
+  - "icon" (string) (default: '')
+  - "text" (string) (default: '')
```
ademarco: Please indent "Parameters" within:
```
  /**
   * @file
   * Theme implementation for a ECL list of icons with text.
   */
```

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
```diff
@@ -0,0 +1,46 @@
...
+#}
+
+{# Internal properties #}
+
+{% set _icon_path = icon_path|default('') %}
+{% set _extra_classes = 'ecl-u-mr-s' %}
```
ademarco: We can simply set that string below, no need for an extra variable here.

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
```diff
@@ -0,0 +1,46 @@
...
+
+{% if _items %}
+<ul class="ecl-unordered-list ecl-unordered-list--no-bullet">
+  {% for _item in _items %}
+    <li class="ecl-u-d-flex ecl-u-align-items-center ecl-unordered-list__item{% if not loop.first %} ecl-u-mt-l{% endif %}">
+      {% include '@ecl/icon' with {
+        icon: {
+          path: _icon_path,
+          type: 'general',
+          name: _item.icon
```
ademarco: We should fallback to `document` if none specified.

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
```diff
@@ -0,0 +1,46 @@
...
+        },
+        extra_classes: _extra_classes
```
ademarco: We can simply set this to `'ecl-u-mr-s'`, no need to use a variable here.

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    {% set _items = _items|merge([item]) %}
+  {% endif %}
+{% endfor %}
+
+{# Print the result #}
+
+{% if _items %}
+<ul class="ecl-unordered-list ecl-unordered-list--no-bullet">
+  {% for _item in _items %}
+    <li class="ecl-u-d-flex ecl-u-align-items-center ecl-unordered-list__item{% if not loop.first %} ecl-u-mt-l{% endif %}">
```
ademarco: Use inline conditions inline:
```twig
<li class="ecl-u-d-flex ecl-u-align-items-center ecl-unordered-list__item{{ not loop.first ? ' ecl-u-mt-l' }}">
```

tests/Kernel/fixtures/rendering.yml
sergepavle: Done. But we don't have, unfortunately, `document` icon. So I used `file` icon.

tests/Kernel/fixtures/rendering.yml
sergepavle: Done. but with `file` icon.

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
sergepavle: Removed.

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
sergepavle: Fixed

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
sergepavle: Done. Used `file` icon

templates/compositions/ec-component-icons-with-text/icons-with-text.html.twig
```diff
@@ -0,0 +1,36 @@
...
+{#
+  /**
+   * @file
+   * Theme implementation for a ECL list of icons with text.
+   */
+  Parameters:
+  - "items" (array) (default: [])
+    - "icon" (string) (default: 'file')
+    - "text" (string) (default: '')
+#}
```
ademarco: I meant as follows:
```
  /**
   * @file
   * Theme implementation for a ECL list of icons with text.
   *
   * Parameters:
   *   - "items" (array) (default: [])
   *     - "icon" (string) (default: 'file')
   *     - "text" (string) (default: '')
   */
```

PR #349, opened 2019-10-29
OPENEUROPA-2388: Make sure that all content types are properly themed.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


tests/Kernel/ContentRenderTest.php
```diff
@@ -82,10 +85,14 @@ protected function setUp() {
       'media',
       'filter',
       'rdf_entity',
+      'oe_media',
+      'media_avportal',
+      'oe_media_avportal',
     ]);
 
+    $this->container->get('config.installer')->installDefaultConfig('module', 'oe_theme_helper');
```
imanoleguskiza: This together with the changes to the configuration files are not completely justified. Is there any particular reason to move the configuration files into the helper module? As discussed on slack, the same result will be achieved by doing     $this->container->get('config.installer')->installDefaultConfig('theme', 'oe_theme');

sergepavle:
Moved to #350

PR #350, opened 2019-10-31
OPENEUROPA-2449: Make sure that all content types are properly themed.

sergepavle:
### Description

Version 2.x of the theme is mature enough to attempt poring content type rendering. This means:

Run and fix tests in \Drupal\Tests\oe_theme\Kernel\ContentRenderTest
Manually create an entry for each content type and check if it displays correctly, if it doesn't fix it and add that case to the tests, so we avoid regressions in future

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


tests/Kernel/ContentRenderTest.php
```diff
@@ -115,12 +123,29 @@ protected function setUp() {
    * Tests that the News node type is rendered with the correct ECL markup.
    */
   public function testNews(): void {
+
+    $media = $this->container
+      ->get('entity_type.manager')
+      ->getStorage('media')->create([
+        'bundle' => 'av_portal_photo',
+        'oe_media_avportal_photo' => 'P-038924/00-15',
```
ademarco: This seems to work without an internet connection too, I guess we are not performing any actual call to the AV Portal service.

PR #353, opened 2019-11-04
OPENEUROPA-2144: ECL 2.0: Gallery pattern.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


oe_theme.theme
```diff
@@ -658,9 +658,16 @@ function oe_theme_preprocess_links__oe_multilingual_content_language_block(array
...
-    $item = GalleryItemValueObject::fromArray($item);
+function oe_theme_preprocess_pattern_gallery(array &$variables): void {
+  foreach ($variables['items'] as $i => &$item) {
+    /** @var Drupal\oe_theme\ValueObject\GalleryItemValueObject $gallery_item_object */
+    $gallery_item_object = GalleryItemValueObject::fromArray($item);
+    $item = [
+      'path' => $gallery_item_object->getThumbnail()->getSource(),
+      'alt' => $gallery_item_object->getThumbnail()->getAlt(),
+      'icon' => $gallery_item_object->getIcon(),
+      'description' => $gallery_item_object->getCaption(),
```
ademarco: This needs to happen in the pattern template itself, since this preprocess runs before more specific ones. You can access value object properties bu using dotted notation. Check the `pattern-file-translation.html.twig` for an example of that.

oe_theme.libraries.yml
```diff
@@ -26,3 +26,10 @@ dialogs:
   dependencies:
     - core/drupal
     - oe_theme/base
+
+gallery:
+  js:
+    js/gallery.js: {}
+  dependencies:
+    - core/drupal
+    - oe_theme/base
```
ademarco: Libraries that are only specific to a pattern should be defined on the pattern definition and not globally. For more info refer to the [UI Patterns documentation](https://ui-patterns.readthedocs.io/en/8.x-1.x/content/patterns-definition.html#expose-pattern-assets-as-libraries).

sergepavle: We've fixed autoinit by patch for fixing bug in ecl-twig. Now we don't need to declare aditional library.

ademarco:
We should maybe add a default variant since we just provide a simple gallery:
```
  variants:
    default:
      label: "Default"
      description: "Simple gallery, just displaying a grid of media items, with no additional behaviour"

```

oe_theme.libraries.yml
```diff
@@ -26,3 +26,4 @@ dialogs:
   dependencies:
     - core/drupal
     - oe_theme/base
+
```
ademarco: Let's remove this new line if possible.

templates/patterns/gallery/gallery.ui_patterns.yml
```diff
@@ -1,6 +1,10 @@
 gallery:
   label: "Gallery"
   description: "Gallery of available media."
+  variants:
+    default:
+      label: "Default"
+      description: "Simple gallery, just displaying a grid of media items, with no additional behaviour"
```
ademarco: Since it actually works with full behavior let's change this into:

```
Default gallery displaying a grid of media items.
```

templates/patterns/gallery/gallery.ui_patterns.yml
```diff
@@ -1,6 +1,10 @@
 gallery:
   label: "Gallery"
   description: "Gallery of available media."
```
ademarco: Let's change this one into:
```
   description: "A gallery with a media browser (it currently only support images, other media types will be added later)."
```

PR #355, opened 2019-11-05
OPENEUROPA-2152:  Fix remaining tests on 2.x branch

ademarco:


tests/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -11,6 +11,8 @@
 /**
  * Test content language switcher rendering.
  *
+ * To be re-enabled after OPENEUROPA-2013 is fixed.
```
ademarco: Always use `@todo`:
```
  * @todo: To be re-enabled after OPENEUROPA-2013 is fixed.
```

PR #358, opened 2019-11-08
OPENEUROPA-2478: Add vertical and horizontal variants to fields patterns

nagyad:


templates/patterns/field_list/pattern-field-list--variant-horizontal.html.twig
```diff
@@ -0,0 +1,11 @@
+{#
+/**
+ * @file
+ * Field list horizontal variant.
+ */
+#}
+
+{% include '@ecl/description-list' with {
+  'variant': 'horizontal',
```
imanoleguskiza: We don't need to define a different template for the variant, the variant is a variable that is available so we can simply pass it on the default template.

templates/patterns/field_list/field_list.ui_patterns.yml
```diff
@@ -1,15 +1,22 @@
...
-        - label: "Where"
-          body: "The EGG, Rue Barra 175, 1070 Brussels, Belgium"
-        - label: "Live streaming"
-          body: "08:00 CET | 00d:23h:35m"
+        - term: "When"
+          definition: "Thursday 15 November, 08:00 AM to Friday 16 November"
+        - term: "Where"
+          definition: "The EGG, Rue Barra 175, 1070 Brussels, Belgium"
+        - term: "Live streaming"
+          definition: "08:00 CET | 00d:23h:35m"
```
ademarco: We don't change our field properties names to follow what ECL requires, we instead remap them in the Twig template itself (not in the preprocess as it would be too early).

imanoleguskiza: I think in this case its fine since this is a pure 2.x pattern so the names aren't set on stone yet.

ademarco: Yeah we could do that if they would make sense in our use case but also `term` / `definition` are just confusing.

ademarco: For @nagyad we should have something like this in the twig template (with the correct naming of course):

```twig
{% set _items = [] %}
{% for _item in items %}
  {% set _items = _items|merge([{
    label: _item.label,
    path: _item.href
  }]) %}
{% endfor %}
```

PR #359, opened 2019-11-08
OPENEUROPA-2467: Remove SASS source mapping from final CSS artifacts.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -29,7 +29,7 @@ pipeline:
     image: node:9.5.0
     commands:
       - npm install --unsafe-perm
-      - npm run build
+      - NODE_ENV=production npm run build
 
   composer-install:
     group: prepare
@@ -77,9 +77,6 @@ pipeline:
   before-release:
     image: fpfis/httpd-php-dev:7.1
     commands:
-      # Rebuild styles without Source map for releases.
-      - npm rebuild node-sass
-      - NODE_ENV=production npm run build:styles
       - ./vendor/bin/run release:create-archive --tag=${DRONE_TAG}
     when:
       event: tag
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -77,6 +77,9 @@ pipeline:
   before-release:
     image: fpfis/httpd-php-dev:7.1
     commands:
+      # Rebuild styles without Source map for releases.
+      - npm rebuild node-sass
+      - NODE_ENV=production npm run build:styles
```
ademarco: What I meant was to do this directly on:
```
pipeline:
  npm-build:
    group: prepare
    image: node:9.5.0
    commands:
      - npm install --unsafe-perm
      - npm run build
```

So:
```
pipeline:
  npm-build:
    group: prepare
    image: node:9.5.0
    commands:
      - npm install --unsafe-perm
      - NODE_ENV=production npm run build
```

This is also theoretically more correct since we will test a prod build.

ademarco:
Ok that's good, let's also remove the build "banner" by applying this:
```patch
--- a/ecl-builder.config.js
+++ b/ecl-builder.config.js
@@ -9,12 +9,7 @@ const nodeModules = __dirname + '/node_modules';
 // SCSS includePaths
 const includePaths = [nodeModules];
 
-const banner = `${pkg.name} - ${
-  pkg.version
-} Built on ${new Date().toISOString()}`;
-
 const style_options = {
-  banner,
   includePaths,
   sourceMap: isProd ? 'none' : true,
 };

``

PR #360, opened 2019-11-11
OPENEUROPA-2472: Display more link only when it exists.

22Alexandra:
### Description

Display more link only when it exists on the social media link component.

### Change log

- Fixed: check if more link exists


nagyad (changes requested):
Please add tests.

ademarco:
Closing as we won't support this on 1.x.

PR #362, opened 2019-11-11
OPENEUROPA-2474: Make sure that teaser display modes are properly themed.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


imanoleguskiza (dismissed):

modules/oe_theme_content_news/oe_theme_content_news.post_update.php
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Update dimensions of 'List item' image style.
+ */
+function oe_theme_content_news_post_update_20001(array &$sandbox): void {
+  $image_style = ImageStyle::load('oe_theme_list_item');
+  $effects = $image_style->getEffects();
+  /** @var \Drupal\image\ImageEffectInterface $effect */
+  foreach ($effects as $effect) {
+    if ($effect->getPluginId() == 'image_scale_and_crop') {
```
nagyad: ===

ademarco (changes requested):
This is flagged waiting for input from the PO.

modules/oe_theme_content_news/oe_theme_content_news.post_update.php
sergepavle: Fixed

sergepavle:
> This is flagged waiting for input from the PO.

Already discussed with @kugta and we are agreed that this is not a blocker.

ademarco:
Instead of changing the preset size we could use the largest one for now and then restore the original functionality once the ECL will provide it. Let's just add a comment to explain the reasoning behind using the largest preset we have.

nagyad (dismissed):

ademarco (approved):

ademarco:
As discussed with PO we will now use the largest image stile we have and restore the responsive behavior once ECL will allow us to do it again.

imanoleguskiza (approved):

PR #364, opened 2019-11-11
OPENEUROPA-2013: Add content language switcher

ademarco:


ademarco:
Looking as follows now: 
image

PR #377, opened 2019-11-21
OPENEUROPA-2221: Theme event content type.

nagyad:


modules/oe_theme_content_event/oe_theme_content_event.module
```diff
@@ -0,0 +1,478 @@
...
+    }
+  }
+
+  return FALSE;
+}
+
+/**
+ * Implements hook_preprocess_HOOK() for node--oe-event--teaser.html.twig.
+ */
+function oe_theme_content_event_preprocess_node__oe_event__teaser(&$variables) {
```
imanoleguskiza: DO we need to keep this empty method?

modules/oe_theme_content_event/oe_theme_content_event.module
```diff
@@ -0,0 +1,478 @@
...
+  $node = $variables['node'];
+
+  // Get the summary description of the event.
+  $variables['content']['description_summary'] = $node->get('oe_event_description_summary')->value;
+
+  // Get the summary report of the event.
+  $variables['content']['report_summary'] = $node->get('oe_event_report_summary')->value;
+
+  // Prepare the full text for the description section.
+  $variables['content']['description'] = $node->get('body')->value;
```
imanoleguskiza: We could add this as in an else clause for the next condition right?

modules/oe_theme_content_event/oe_theme_content_event.module
```diff
@@ -0,0 +1,478 @@
...
+
+  // Get the location of the event from the Venue.
+  $postal_address_rendered = _oe_theme_content_event_get_rendered_address($node);
+
+  // Create the event date string depending on if it is on the same day.
+  $event_date_string = $event_dates->isSameDate() ?
+    $date_formatter->format($event_dates->getStartTime(), 'oe_theme_event_date_hour') :
+    t('@start_date to @end_date', [
+      '@start_date' => $date_formatter->format($event_dates->getStartTime(), 'oe_theme_event_date_hour'),
+      '@end_date' => $date_formatter->format($event_dates->getEndTime(), 'oe_theme_event_date'),
```
imanoleguskiza, line 389: Is this intentional? The start date has the hour but the end date doesn't.

modules/oe_theme_content_event/oe_theme_content_event.module
```diff
@@ -0,0 +1,478 @@
...
+ *   The event node object.
+ *
+ * @return array
+ *   The prepared array.
+ */
+function _oe_theme_content_event_prepare_practical_information(NodeInterface $node): array {
+  /** @var \Drupal\Core\Datetime\DateFormatter $date_formatter */
+  $date_formatter = \Drupal::service('date.formatter');
+  /** @var \Drupal\Core\Render\Renderer $renderer */
+  $renderer = \Drupal::service('renderer');
```
imanoleguskiza, line 152: This doesn't seem like a good practice, since this is a preprocess method, why not use the render arrays?

imanoleguskiza (changes requested):
The one thing I found missing from the rendered event is a link to the livestream url, is that intentional? Also there are a lot of helper methods on the module file, I think it would be best if we had a service maybe? That way you could inject all the other services you need there as well.

modules/oe_theme_content_event/oe_theme_content_event.module
nagyad: Yes it is intentional.

modules/oe_theme_content_event/oe_theme_content_event.module
nagyad: No.

modules/oe_theme_content_event/oe_theme_content_event.module
nagyad: We have to do it here, because ECL twig does not support render arrays. I checked this with @ademarco

modules/oe_theme_content_event/oe_theme_content_event.module
nagyad: Yes we could.

nagyad:
We make services to create something reusable. These preprocess functions are tied to Event rendering business logic + ECL twig expectations. I think making a service to wrap those services in play, would not make much sense either, the wrapper service would be called in the preprocess the same way. Since, we are not using formatters in order to have more control over changes without the need of upgrade path each time, that's the sacrifice. We could optimise the code for sure.

modules/oe_theme_content_event/oe_theme_content_event.module
nagyad: Not for now.

PR #382, opened 2019-11-27
OPENEUROPA-2357: Theming and tests theming of share page block.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


ademarco:
After discussing it with @22Alexandra we decided to change approach: we will provide a template on the oe_webtools module and override it here in the theme.

PR #383, opened 2019-11-27
OPENEUROPA-2557: Add wrapper around meta information on page header.

ademarco:


imanoleguskiza (approved):
PR solves the issue although it isn't very elegant (a "title" wrapper sounds like something for the title, it would be nice to have separate wrappers or if they have a different name)

ademarco:
Yep we know, ECL Twig team will take care of fixing it properly in an upcoming release, which would make the whole patch useless.

nagyad (approved):

PR #385, opened 2019-11-28
OPENEUROPA-2221: Theme event content type.

nagyad:


ademarco:
Closing as we changed the way the event content type is built. Check #404.

PR #387, opened 2019-12-04
OPENEUROPA-2393: Make sure that javascript behaviours are bind once.

nagyad:
If ajax is used then the DOM changes and Drupal.behaviours binds event on the element more than once. Resulting in faulty behaviors such as toggles won't work anymore. http://d7.thecarneyeffect.co.uk/applying-javascript-behaviours-once-drupal

"Any object defined as a property of Drupal.behaviors will get its attach() method called when the DOM has loaded both initially and after any AJAX calls."

Check all JS files in oe_theme, and evaluate case by case the usage of .once() context to make sure it binds only once.

https://www.drupal.org/docs/8/api/javascript-api/javascript-api-overview

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -7,6 +7,7 @@
     <env name="SIMPLETEST_BASE_URL" value="${drupal.base_url}"/>
     <env name="SIMPLETEST_DB" value="mysql://${drupal.database.user}:${drupal.database.password}@${drupal.database.host}:${drupal.database.port}/${drupal.database.name}"/>
     <env name="SIMPLETEST_SPARQL_DB" value="sparql://${drupal.sparql.host}:${drupal.sparql.port}/"/>
+    <env name="MINK_DRIVER_ARGS_WEBDRIVER" value='["${selenium.browser}", null, "${selenium.host}:${selenium.port}/wd/hub"]'/>
   </php>
   <testsuites>
     <testsuite>
--- a/runner.yml.dist
+++ b/runner.yml.dist
@@ -41,6 +41,7 @@ drupal:
 selenium:
   host: "http://selenium"
   port: "4444"
+  browser: "chrome"
 
 commands:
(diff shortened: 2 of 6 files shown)


sergepavle (dismissed):

oe_theme.libraries.yml
```diff
@@ -8,6 +8,8 @@ component_library_ec:
     js/ecl_auto_init.js: {}
   dependencies:
     - core/drupal
+    - core/jquery
+    - core/jquery.once
```
sergepavle: Tiny remark. Looks like it is not so necessary to add `core/jquery` because `core/jquery.once` already have it self this dependency. But I saw in core similar like here approach.
So it is fine for me. But if you you wish - you could remove `core/jquery`.

js/ecl_auto_init.js
```diff
@@ -2,10 +2,12 @@
  * @file
  * ECL Auto Init behavior.
  */
-(function (ECL, Drupal) {
+(function ($, ECL, Drupal) {
   Drupal.behaviors.eclAutoInit = {
-    attach: function attach() {
-      window.ECL.autoInit();
+    attach: function attach(context) {
+      $(document, context).once('eclAutoInit').each(function () {
```
brummbar: By putting a once on the `document`, which is the web page itself (see https://developer.mozilla.org/en-US/docs/Web/API/document), it actually prevents any element that is added by ajax to get the behaviours attached. 
Also `document` is the top element, so passing `context` would be wrong because no document can be found inside it.

brummbar (changes requested):
Added a javascript test that verifies the behaviour functionality. The old approach worked because ECL has auto init functionalities, that checks for JS to not get applied twice.
The new approach prevents any element added with Ajax to get behaviours applied.

sergepavle (approved):

brummbar (approved):

PR #392, opened 2019-12-06
ISSUE-386: Add support for extra container classes in select template.

kp77:
### Description

Add support for additional container classes for select element so that container width modifier classes (see here under 'Widths': https://ec.europa.eu/component-library/ec/components/forms/select/code/ )  can be passed to the template.

### Change log

- Changed: 
./templates/form/select.html.twig
- Added 'wrapper_attributes' that accepts 'wrapper_classes' variable.
- Renamed 'extra_attributes' to 'extra_classes'

### DIFF
--- a/templates/form/select.html.twig
+++ b/templates/form/select.html.twig
@@ -11,13 +11,19 @@
   {% set extra_classes = extra_classes|merge(['ecl-select--invalid']) %}
 {% endif %}
 
-{% set wrapper_attributes = create_attribute({'class': 'ecl-select__container'}) %}
+{# Set default wrapper classes. #}
+{# This also allows overriding templates to take control of classes like width modifiers. #}
+{% set wrapper_classes = wrapper_classes|default([
+  'ecl-select__container',
+  'ecl-select__container--m',
+]) %}
+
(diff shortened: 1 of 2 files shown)


ademarco:
Thanks for this @kp77 ! I'm a bit uneasy in adding a variable that is not supported by Drupal core, meaning: Drupal core's Select render element does not support setting a wrapper class from outside.

What's your actual use case? you need to set the select dimension based on some external conditions? can you paste here the code? Maybe we can find an alternative solution, or we can support this by just allowing a template override.

kp77:
Hi @ademarco 
I agree with the point regarding variable not supported by Drupal.

The issue is, that as I understand the width modifier is actually not optional in the component. It looks to me that a default size class is always used on the wrapper in the examples.

ademarco:
Ok the proposed solution allows for a size class override in a template override and still keeps drupal-side of things clean.

ademarco (approved):

kp77:
Thanks a lot for the quick fix!

imanoleguskiza (approved):
Good improvement.

PR #397, opened 2019-12-23
OPENEUROPA-2693: On IE11 the icons are not displayed.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra (dismissed):

js/svg4ie.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @file
+ * ECL fix for SVGs in IE 11.
+ */
+(function (ECL, Drupal) {
```
brummbar: No need to have ECL as parameter.

js/svg4ie.js
```diff
@@ -0,0 +1,13 @@
...
+  Drupal.behaviors.svg4Ie = {
+    attach: function attach() {
+      if (typeof svg4everybody === "function" && !!navigator.userAgent.match(/Trident.*rv\:11\./)) {
```
brummbar: No need to test that `svg4everybody` is defined, we already set the dependency.
Also no need to test for browser, the library does it already.

oe_theme.libraries.yml
```diff
@@ -6,8 +6,10 @@ component_library_ec:
   js:
     dist/ec/scripts/ecl-ec-preset-legacy-website.js: {}
     js/ecl_auto_init.js: {}
+    js/svg4ie.js: { preprocess: false }
```
brummbar: Why disabling the aggregation on this script?

oe_theme.libraries.yml
```diff
@@ -17,8 +19,10 @@ component_library_eu:
   js:
     dist/eu/scripts/ecl-eu-preset-legacy-website.js: {}
     js/ecl_auto_init.js: {}
+    js/svg4ie.js: { preprocess: false }
```
brummbar: Same as above.

oe_theme.libraries.yml
```diff
@@ -30,3 +34,7 @@ dialogs:
     js/dialogs.js: {}
   dependencies:
     - core/drupal
+
+svg4everybody:
```
brummbar: To know which version of the library we included, let's use the version here.

oe_theme.libraries.yml
```diff
@@ -30,3 +34,7 @@ dialogs:
...
+  js:
+    js/svg4everybody.min.js: { preprocess: false }
```
brummbar: Also here, why disabling aggregation?

brummbar (changes requested):
Note from @ademarco: let's add the npm package `svg4everybody` as dependency, and then use the ecl-builder to copy it over in the dist package instead of putting the library file here.

oe_theme.libraries.yml
brummbar: This is not needed anymore, as we will use npm to handle the library.

js/svg4ie.js
brummbar: Since we won't be testing for browser, let's remove the `IE` stuff. It's just `svg4Everybody` the behavior name now.

ademarco (changes requested):
I'd declare one single library as such:
```yml
external_svg:
  js:
    dist/js/svg4everybody.min.js: { minified: true }
    js/external_svg.js: {}
```
And then we can simply load this library via the hteme info.yml file, as we do with the `editor` library:

```yml
libraries:
  - oe_theme/editor
  - oe_theme/external_svg
```

oe_theme.libraries.yml
sergepavle: Left over on try to fix order of js files.
Removed.

js/external_svg.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @file
+ * ECL fix for SVGs in IE 11.
+ */
+(function (Drupal) {
+  Drupal.behaviors.svg4Everybody = {
+    attach: function attach() {
```
brummbar: Missed this last time: no need to name the function.
```suggestion
    attach: function () {
```

ecl-builder.config.js
```diff
@@ -63,5 +63,6 @@ module.exports = {
     { from: path.resolve(nodeModules, '@ecl-twig'), patterns: 'ec-component-tag/*.twig', to: path.resolve(outputFolder, 'components') },
     { from: path.resolve(nodeModules, '@ecl-twig'), patterns: 'ec-component-text-input/*.twig', to: path.resolve(outputFolder, 'components') },
     { from: path.resolve(nodeModules, '@ecl-twig'), patterns: 'ec-component-timeline/*.twig', to: path.resolve(outputFolder, 'components') },
+    { from: path.resolve(nodeModules, 'svg4everybody/dist'), to: path.resolve(outputFolder, 'dist/js') },
```
brummbar: Let's copy only the minified version of the script.

oe_theme.libraries.yml
```diff
@@ -30,3 +30,8 @@ dialogs:
     js/dialogs.js: {}
   dependencies:
     - core/drupal
+
+external_svg:
+  js:
+    dist/js/svg4everybody.js: {}
```
brummbar: So here we put the minified version.

brummbar (changes requested):

brummbar (approved):

ademarco (approved):

PR #399, opened 2020-01-10
OPENEUROPA-2558: Override oe webtools social share template.

22Alexandra:
### Description

Override oe webtools social share template.

### Change log

- Changed: oe webtools social share template
- Fixed: kernel test


composer.json
```diff
@@ -34,6 +34,7 @@
         "openeuropa/oe_multilingual": "~1.0",
         "openeuropa/oe_paragraphs": "~1.3.0",
         "openeuropa/oe_search": "~1.0",
+        "openeuropa/oe_webtools": "dev-OPENEUROPA-2558",
```
nagyad: Use composer patching instead of this.

templates/oe_webtools/oe-webtools-social-share.html.twig
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * @file
+ * Theme override for oe webtools social share block.
+ *
+ * @see ./core/modules/system/templates/block.html.twig
+ */
+#}
+<div class="ecl-social-media-share">
+  <p class="ecl-social-media-share__description">
+      {{ title }}
```
nagyad: Indentation issue. I think it could even go single line like the script tag below.

templates/oe_webtools/oe-webtools-social-share.html.twig
```diff
@@ -0,0 +1,14 @@
+{#
+/**
+ * @file
+ * Theme override for oe webtools social share block.
+ *
+ * @see ./core/modules/system/templates/block.html.twig
```
nagyad: Why do we need this here?

tests/Kernel/Webtools/SocialShareTest.php
```diff
@@ -0,0 +1,46 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme\Kernel\Webtools;
+
+use Drupal\Tests\oe_theme\Kernel\AbstractKernelTestBase;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Test share this block rendering.
```
nagyad: Test social share block rendering. - would be a better fit.

nagyad (changes requested):

composer.json
22Alexandra: I'll patch after oe_webtools PR is merged.

nagyad: You can still patch with this branch and after just change to master. This approach is not correct.

tests/Kernel/Webtools/SocialShareTest.php
```diff
@@ -0,0 +1,46 @@
...
+    ];
+
+    $render = $this->buildBlock('social_share', $config);
+
+    $html = (string) $this->container->get('renderer')->renderRoot($render);
+    $crawler = new Crawler($html);
+
+    // Make sure that social media share block is correctly rendered.
+    $custom_footer = $crawler->filter('p.ecl-social-media-share__description');
+    $this->assertCount(1, $custom_footer);
```
nagyad, line 44: Could we assert the value as well?

nagyad (dismissed):

dxvargas (dismissed):

nagyad (dismissed):

nagyad (approved):

dxvargas (approved):

PR #400, opened 2020-01-10
OPENEUROPA-2642: Create a formatter for address field.

sergepavle:
### Description

It turned out that the address_plain formatter does not pass the entity to the theme function, making it impossible to create the proper suggestions. The only option left here is to create a custom formatter for the address field that will allow us to configure a separator and will print the address simply separated by that, like: Rue de la Loi, 56, 1000, Brussels.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,8 +9,7 @@
         "drupal/core": "^8.7",
         "drupal/ui_patterns": "^1.0",
         "openeuropa/ecl-twig-loader": "~2.0",
-        "php": "^7.1",
-        "twig/twig": "~1.41"
+        "php": "^7.1"
     },
     "require-dev": {
         "cweagans/composer-patches": "~1.0",
@@ -18,6 +17,7 @@
         "consolidation/robo": "~1.4",
         "consolidation/annotated-command": "^2.8.2",
         "drupal-composer/drupal-scaffold": "^2.5.2",
+        "drupal/address": "^1.7",
         "drupal/config_devel": "~1.2",
         "drupal/drupal-driver": "~2.0.0-alpha6",
         "drupal/drupal-extension": "~4.0",
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
         "drupal/core": "^8.7",
         "drupal/ui_patterns": "^1.0",
         "openeuropa/ecl-twig-loader": "~2.0",
-        "php": "^7.1"
+        "php": "^7.1",
+        "twig/twig": "~1.41"
     },
     "require-dev": {
         "cweagans/composer-patches": "~1.0",
--- a/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
+++ b/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
@@ -4,8 +4,10 @@ declare(strict_types = 1);
 
 namespace Drupal\oe_theme_helper\Plugin\Field\FieldFormatter;
 
+use CommerceGuys\Addressing\Locale;
 use Drupal\address\AddressInterface;
 use Drupal\address\Plugin\Field\FieldFormatter\AddressDefaultFormatter;
+use Drupal\Component\Utility\Html;
 use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Language\LanguageInterface;
@@ -92,11 +94,25 @@ class AddressInlineFormatter extends AddressDefaultFormatter implements Containe
     $countries = $this->countryRepository->getList();
     $address_format = $this->addressFormatRepository->get($country_code);
     $values = $this->getValues($address, $address_format);
-    $values['country'] = $countries[$country_code];
+
+    $address_elements['%country'] = Html::escape($countries[$country_code]);
+    foreach ($address_format->getUsedFields() as $field) {
+      $address_elements['%' . $field] = Html::escape($values[$field]);
+    }
+
+    if (Locale::matchCandidates($address_format->getLocale(), $address->getLocale())) {
+      $format_string = '%country' . "\n" . $address_format->getLocalFormat();
+    }
+    else {
+      $format_string = $address_format->getFormat() . "\n" . '%country';
+    }
+
+    $items = self::replacePlaceholders($format_string, $address_elements);
+
     return [
       '#theme' => 'oe_theme_helper_address_inline',
       '#address' => $address,
-      '#address_items' => $values,
+      '#address_items' => $items,
       '#address_delimiter' => $this->getSetting('delimiter'),
       '#cache' => [
         'contexts' => [
@@ -104,6 +120,35 @@ class AddressInlineFormatter extends AddressDefaultFormatter implements Containe
         ],
       ],
     ];
+  }
+
+  /**
+   * Replaces placeholders in the given string.
+   *
+   * @param string $string
+   *   The string containing the placeholders.
+   * @param array $replacements
+   *   An array of replacements keyed by their placeholders.
+   *
+   * @return array
+   *   The exploded lines.
+   */
+  public static function replacePlaceholders($string, array $replacements) {
+    // Make sure the replacements don't have any unneeded newlines.
+    $replacements = array_map('trim', $replacements);
+    $string = strtr($string, $replacements);
+    // Remove noise caused by empty placeholders.
+    $lines = explode("\n", $string);
+    foreach ($lines as $index => $line) {
+      // Remove leading punctuation, excess whitespace.
+      $line = trim(preg_replace('/^[-,]+/', '', $line, 1));
+      $line = preg_replace('/\s\s+/', ' ', $line);
+      $lines[$index] = $line;
+    }
+    // Remove empty lines.
+    $lines = array_filter($lines);
+
+    return $lines;
   }
 
 }
--- a/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
+++ b/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
@@ -4,21 +4,13 @@ declare(strict_types = 1);
 
 namespace Drupal\oe_theme_helper\Plugin\Field\FieldFormatter;
 
-use CommerceGuys\Addressing\AddressFormat\AddressField;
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -9,22 +9,24 @@
         "drupal/core": "^8.7",
         "drupal/ui_patterns": "^1.0",
         "openeuropa/ecl-twig-loader": "~2.0",
-        "php": "^7.1"
+        "php": "^7.1",
+        "twig/twig": "~1.41"
```
ademarco: We shouldn't require a specific version of twig for this, please use whatever available in Drupal core's Twig.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,243 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Language\LanguageInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Plugin implementation of the 'oe_address_inline' formatter.
+ *
+ * @FieldFormatter(
+ *   id = "oe_address_inline",
```
ademarco: We need to prepend this with the module's name `oe_theme_helper_address_inline`

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,243 @@
...
+
+  /**
+   * The subdivision repository.
+   *
+   * @var \CommerceGuys\Addressing\Subdivision\SubdivisionRepositoryInterface
+   */
+  protected $subdivisionRepository;
+
+  /**
+   * Constructs an AddressPlainFormatter object.
```
ademarco: Copy paste leftover.

modules/oe_theme_helper/templates/oe-theme-helper-address-inline.html.twig
```diff
@@ -0,0 +1,24 @@
...
+{%
+  set ordered_address = [
+  address_items.addressLine1,
+  address_items.addressLine2,
+  address_items.postalCode,
+  address_items.locality,
+]
+%}
+<p class="address" translate="no">
+  {{ ordered_address|filter(v => v is not empty)|join(address_delimiter) }}
```
ademarco: Let's use here only what Twig in Drupal core provides us.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,243 @@
...
+   *
+   * @param \Drupal\address\AddressInterface $address
+   *   The address.
+   * @param \CommerceGuys\Addressing\AddressFormat\AddressFormat $address_format
+   *   The address format.
+   *
+   * @return array
+   *   The values, keyed by address field.
+   */
+  protected function getValues(AddressInterface $address, AddressFormat $address_format) {
```
ademarco: Can't we extend `AddressDefaultFormatter` and inherit this method from it? If we need to modify its values we can always do in the current class, I mean:

```php
protected function getValues(AddressInterface $address, AddressFormat $address_format) {
    $values = parent::getValues(...);
    // modify values here...
    return $values;

```

ademarco (changes requested):

nagyad (dismissed):

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * @return array
+   *   A renderable array.
+   */
+  protected function viewElement(AddressInterface $address, $langcode) {
+    $country_code = $address->getCountryCode();
+    $countries = $this->countryRepository->getList();
+    $address_format = $this->addressFormatRepository->get($country_code);
+    $values = $this->getValues($address, $address_format);
+    $values['country'] = $countries[$country_code];
```
ademarco: We need to use the address format contained in the `addressFormatRepository` and use that to render our address. We could check what happens here `\Drupal\address\Plugin\Field\FieldFormatter\AddressDefaultFormatter::viewElement()` and simply replace `\n` with our separator and not use `html_tag` render arrays since we only need to display a simple string separated by a comma, at the end.

ademarco (changes requested):

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,154 @@
...
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
+    $form = parent::settingsForm($form, $form_state);
+
+    $form['delimiter'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Delimiter'),
+      '#default_value' => $this->getSetting('delimiter'),
+      '#description' => $this->t('Specify delimiter between address items.'),
```
ademarco, line 46: Make this required.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,154 @@
...
+   */
+  public function settingsSummary() {
+    $settings_summary = parent::settingsSummary();
+    $example_address_items = [
+      'Rue de la Loi',
+      '56',
+      '1000',
+      'Brussels',
+    ];
+    $settings_summary[] = implode($this->getSetting('delimiter'), $example_address_items);
```
ademarco: I would remove the example and just print what the delimiter is, to have on the ui something like 
```
Delimiter: ,
```

modules/oe_theme_helper/templates/oe-theme-helper-address-inline.html.twig
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Available variables:
+ *   - address: Address object.
+ *   - address_items: Address items.
+ *   - address_delimiter: Delimiter between address items.
+ *
+ * @ingroup themeable
+ */
+#}
+<p class="address" translate="no">
```
ademarco: Make this a `span` and this is an inline formatter, and remove the class name, just ` <span translate="no">...`

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,154 @@
...
+use Drupal\address\AddressInterface;
+use Drupal\address\Plugin\Field\FieldFormatter\AddressDefaultFormatter;
+use Drupal\Component\Utility\Html;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Language\LanguageInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+
+/**
+ * Plugin implementation of the 'oe_theme_helper_address_inline' formatter.
```
ademarco: Use a more descriptive text here, like what this actually does:

```
Format an address inline, using its locale format and a configurable separator.
```

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * @FieldFormatter(
+ *   id = "oe_theme_helper_address_inline",
+ *   label = @Translation("Inline address"),
+ *   field_types = {
+ *     "address",
+ *   },
+ * )
+ */
+class AddressInlineFormatter extends AddressDefaultFormatter implements ContainerFactoryPluginInterface {
```
ademarco: Since now we don't use anything from the base class `AddressDefaultFormatter` we can also not extend it and extend the default class.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultSettings() {
+    return [
+      'delimiter' => ', ',
+    ] + parent::defaultSettings();
```
ademarco: Same here, since we don't use anything from `AddressDefaultFormatter` we can just have our own settings and not inherit `parent::defaultSettings()`.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
+    $form = parent::settingsForm($form, $form_state);
```
ademarco: Same here, since we don't use anything from `AddressDefaultFormatter` we can just have our own settings and not inherit `parent::defaultSettings()`.

modules/oe_theme_helper/tests/src/Kernel/Plugin/Field/FieldFormatter/AddressInlineFormatterTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Drupal\Tests\oe_theme_helper\Kernel\Plugin\Field\FieldFormatter;
+
+use Drupal\entity_test\Entity\EntityTest;
+use Drupal\Tests\address\Kernel\Formatter\FormatterTestBase;
+
+/**
+ * Test AddressInlineFormatter plugin.
+ */
+class AddressInlineFormatterTest extends FormatterTestBase {
```
ademarco, line 13: Nice catch using `FormatterTestBase`!

ademarco (changes requested):

ademarco (approved):

nagyad (approved):

PR #403, opened 2020-01-20
OPENEUROPA-2674: Remove dialog.js library.

sergepavle:
### Description

Dialogs have been deprecated in ecl 2.x . They are present only in the language switcher, but not as standalone. Thus the dialog.js file that we added to initialise dialogs is not needed anymore.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


brummbar (changes requested):
The `js/dialogs.js` needs to be removed too.

brummbar (approved):

nagyad (approved):

PR #420, opened 2020-02-05
OPENEUROPA-2809: Add Ratio 3:2 medium image style.

22Alexandra:
### Description

Creates new Ratio 3:2 medium image style and styles image of text with featured media paragraph using oe_theme_medium_no_crop style.

### Change log

- Added: oe_theme_ratio_3_2_medium image style
- Changed: 
- Deprecated:
- Removed:
- Fixed: image style of text with featured media paragraph
- Security:

### Commands

```sh

```

### DIFF
--- a/modules/oe_theme_helper/oe_theme_helper.post_update.php
+++ b/modules/oe_theme_helper/oe_theme_helper.post_update.php
@@ -139,7 +139,13 @@ function oe_theme_helper_post_update_20006() {
 /**
  * Add Ratio 3:2 medium image style.
  */
-function oe_theme_helper_post_update_20007(array &$sandbox): void {
+function oe_theme_helper_post_update_20007(array &$sandbox) {
+  // If the image style already exists, we bail out.
+  $style = \Drupal::entityTypeManager()->getStorage('image_style')->load('oe_theme_ratio_3_2_medium');
+  if ($style) {
+    return 'The image style was previously created.';
+  }
+
   // Create image style.
   $image_style = ImageStyle::create([
     'name' => 'oe_theme_ratio_3_2_medium',
--- a/modules/oe_theme_helper/oe_theme_helper.post_update.php
+++ b/modules/oe_theme_helper/oe_theme_helper.post_update.php
@@ -135,3 +135,29 @@ function oe_theme_helper_post_update_20006() {
     $config_factory->getEditable($block)->save();
   }
 }
+
+/**
+ * Add Ratio 3:2 medium image style.
+ */
+function oe_theme_helper_post_update_20007(array &$sandbox): void {
+  // Create image style.
+  $image_style = ImageStyle::create([
+    'name' => 'oe_theme_ratio_3_2_medium',
+    'label' => 'Ratio 3:2 medium',
+  ]);
+
+  // Create effect.
+  $effect = [
+    'id' => 'image_scale_and_crop',
+    'weight' => 1,
+    'data' => [
+      'anchor' => 'center-center',
+      'width' => '600',
+      'height' => '400',
+    ],
+  ];
+
+  // Add effect to the image style and save.
+  $image_style->addImageEffect($effect);
+  $image_style->save();
+}
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -12,6 +12,7 @@ use Drupal\Component\Utility\Html;
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Render\Element;
 use Drupal\Core\Url;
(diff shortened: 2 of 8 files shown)


oe_theme.theme
```diff
@@ -925,25 +929,11 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
...
-    $uri = $file->getFileUri();
-    $cacheability->addCacheableDependency($file);
-  }
+  $thumbnail = $media->get('thumbnail')->first();
+  $object = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_medium_no_crop');
   $variables['image'] = [
-    'src' => file_create_url($uri),
-    'alt' => $media->get($field_name)->alt,
+    'src' => $object['src'],
+    'alt' => $thumbnail->alt,
```
ademarco: Can't we just do the following?
```php
$variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_medium_no_crop');
```

oe_theme.theme
```diff
@@ -913,6 +913,10 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
 
   /** @var \Drupal\media\Entity\Media $media */
   $media = $paragraph->get('field_oe_media')->entity;
+  if (!$media instanceof MediaInterface) {
+    // The media entity is not available anymore, bail out.
+    return;
```
imanoleguskiza, line 918: I'm not sure if this will ever be the case, we already check whether the field is empty or not and the field itself is a reference field that only allows MediaInterface entities. If we decide to keep this check, then we should define the $cacheability variable after this check, not before.

modules/oe_theme_helper/oe_theme_helper.post_update.php
```diff
@@ -135,3 +135,35 @@ function oe_theme_helper_post_update_20006() {
     $config_factory->getEditable($block)->save();
   }
 }
+
+/**
+ * Add Ratio 3:2 medium image style.
+ */
+function oe_theme_helper_post_update_20007(array &$sandbox) {
```
imanoleguskiza, line 142: I believe the configuration update needs to take place in ewcms and not in the component.

imanoleguskiza: To be confirmed by the coordination team.

tests/Kernel/ValueObject/ImageTest.php
```diff
@@ -0,0 +1,93 @@
...
+    ])->save();
+
+    // Copy file in public files to use it for styling.
+    \Drupal::service('file_system')->copy(__DIR__ . '/../../fixtures/example_1.jpeg', 'public://example_1.jpg');
+  }
+
+  /**
+   * Test the image value object has the correct style applied.
+   */
+  public function testFromStyledImageItem() {
```
imanoleguskiza, line 62: Since we are adding tests for the ImageValueObject, lets add a test for fromImageItem as well.

imanoleguskiza (changes requested):

tests/Kernel/ValueObject/ImageTest.php
22Alexandra: It is tested in here: https://github.com/openeuropa/oe_theme/blob/2.x/tests/Unit/ValueObject/ImageValueObjectTest.php#L107

oe_theme.theme
22Alexandra: If you remove the image, the entity is there but the image is not.

modules/oe_theme_helper/oe_theme_helper.post_update.php
imanoleguskiza: Discussed with the coordination team, the update path is here since other components might rely on the image style.

imanoleguskiza (dismissed):

src/ValueObject/ImageValueObject.php
```diff
@@ -144,4 +144,25 @@ public static function fromImageItem(ImageItem $image_item): ValueObjectInterfac
     );
   }
 
+  /**
+   * Construct object from a styled image.
```
nagyad: Here you are still constructing from a Drupal image field and pass it through an image style.
I would change the comment according to this. For the function name I could not come up with something better either. :+1:

nagyad (changes requested):
Good job so far! :)

nagyad (dismissed):

brummbar (approved):

hernani (approved):

PR #425, opened 2020-02-06
SGECI2-370: Displaying wrong language (file entity not translatable).

golddragon007:
### Description

If a media type is translatable and we translate the file, it will be displayed wrongly the translated file language. File entity not translatable, so we need to use the media entity language to determine what language the file has.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed: oe_theme_preprocess_media__document__default
- Security:


imanoleguskiza:
Fix was implemented in a separate PR

PR #438, opened 2020-02-25
OPENEUROPA-2803: Add a length field for featured items and list items.

imanoleguskiza:
### Description

Add a new field to specific patterns to allow trimming the contents when using them.

### Change log

- Added: Add a length field for featured items and list items.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -7,10 +7,12 @@
 
 declare(strict_types = 1);
 
-use Drupal\Core\Form\FormStateInterface;
 use Drupal\Component\Utility\Html;
+use Drupal\Component\Render\MarkupInterface;
 use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Element;
+use Drupal\Core\Template\Attribute;
 use Drupal\Core\Url;
 use Drupal\media\MediaInterface;
 use Drupal\media\Plugin\media\Source\Image;
@@ -20,7 +22,6 @@ use Drupal\oe_theme\ValueObject\DateValueObject;
 use Drupal\oe_theme\ValueObject\GalleryItemValueObject;
 use Drupal\oe_theme\ValueObject\ImageValueObject;
 use Drupal\oe_theme_helper\EuropeanUnionLanguages;
-use Drupal\Core\Template\Attribute;
 
 /**
  * Implements hook_form_FORM_ID_alter().
@@ -1246,10 +1247,9 @@ function oe_theme_preprocess_pattern_link__preview(array &$variables): void {
 
 /**
  * Implements hook_preprocess_HOOK().
- *
- * Trim the description field on Featured items.
  */
 function oe_theme_preprocess_pattern_featured_item(array &$variables): void {
+  // If the length value is set, trim the description value.
   if (empty($variables['length'])) {
     return;
   }
@@ -1258,10 +1258,9 @@ function oe_theme_preprocess_pattern_featured_item(array &$variables): void {
 
 /**
  * Implements hook_preprocess_HOOK().
- *
- * Trim the detail field on link list items.
  */
 function oe_theme_preprocess_pattern_list_item(array &$variables): void {
+  // If the length value is set, trim the detail value.
   if (empty($variables['length'])) {
     return;
   }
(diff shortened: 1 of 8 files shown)


oe_theme.theme
```diff
@@ -1243,3 +1243,56 @@ function oe_theme_preprocess_pattern_link__preview(array &$variables): void {
   // Pattern preview will pass along a \Drupal\Core\Render\Markup object.
   $variables['url'] = Url::fromUserInput((string) $variables['url']);
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ *
+ * Trim the description field on Featured items.
```
upchuk: This preprocessor can do multiple things. Among them is the trimming. So the comment here should be either just the first line or if we wanna be more descriptive we can say it preprocesses the featured item pattern. But you can leave it out since the rest don't seem to have it either. Same for the list item.

oe_theme.theme
```diff
@@ -1243,3 +1243,56 @@ function oe_theme_preprocess_pattern_link__preview(array &$variables): void {
...
+ */
+function oe_theme_preprocess_pattern_list_item(array &$variables): void {
+  if (empty($variables['length'])) {
+    return;
+  }
+  $variables['detail'] = _oe_theme_trim_textfield($variables['detail'], $variables['length']);
+}
+
+/**
+ * Trims the contents of a text field.
```
upchuk, line 1272: Let's add here a bit more detail as to what we are doing seeing that we have multiple cases of teaser.

oe_theme.theme
```diff
@@ -1243,3 +1243,56 @@ function oe_theme_preprocess_pattern_link__preview(array &$variables): void {
...
+ *
+ * @param string|array $teaser
+ *   The teaser to be trimmed.
+ * @param int $size
+ *   The length of the trimmed teaser.
+ *
+ * @return string|array
+ *   The trimmed teaser.
+ */
+function _oe_theme_trim_textfield($teaser, $size = 150) {
```
upchuk: Again, missing type hints, return types...

oe_theme.theme
```diff
@@ -1243,3 +1243,56 @@ function oe_theme_preprocess_pattern_link__preview(array &$variables): void {
...
+  if ((!is_array($teaser)) && ((!is_object($teaser) && settype($teaser, 'string') !== FALSE) || (is_object($teaser) && method_exists($teaser, '__toString')))) {
```
upchuk: What is this overcomplication? What was wrong with checking if it's a simple string?

oe_theme.theme
```diff
@@ -1243,3 +1243,56 @@ function oe_theme_preprocess_pattern_link__preview(array &$variables): void {
...
+    $teaser = text_summary((string) $teaser, NULL, $size);
+  }
+  elseif (is_array($teaser)) {
```
upchuk: Instead of putting 10 elseif, just return the teaser if a case matches

oe_theme.theme
```diff
@@ -1243,3 +1243,56 @@ function oe_theme_preprocess_pattern_link__preview(array &$variables): void {
...
+    if (isset($teaser['#markup'])) {
+      $teaser['#markup'] = text_summary($teaser['#markup'], NULL, $size);
+    }
+    elseif (isset($variable['#plain_text'])) {
+      $variable['#plain_text'] = text_summary($teaser['#plain_text'], NULL, $size);
```
upchuk: What is this $variable?!

tests/Kernel/fixtures/rendering.yml
```diff
@@ -388,6 +388,27 @@
       'article.ecl-content-item div.ecl-content-item__meta': "Meta 1 | Meta 2"
       'article.ecl-content-item div.ecl-content-item__title .ecl-link': "Item title"
       'article.ecl-content-item div.ecl-content-item__description': "Item text"
+- array:
```
upchuk, line 391: Any way we can add coverage also for non string cases? Because as you can see, the lack of it caused bugs in the actual trimming function..

upchuk (changes requested):

oe_theme.theme
imanoleguskiza: The problem is that when testing using our renderingtest class, the $teaser is going to be a Markup object. So I wanted to cover all cases where we have strings or objects that can be strings.

oe_theme.theme
imanoleguskiza: I added the type hint for $size but since the $teaser and return can be of various types I don't think I can type hint them?

oe_theme.theme
imanoleguskiza: You are right, I keep it consistent now and add a small comment inside the method.

oe_theme.theme
imanoleguskiza: I added a more detailed description and also more comments on the method itself, let me know what you think.

oe_theme.theme
imanoleguskiza: I refactored the method to take this into account.

oe_theme.theme
imanoleguskiza: Fixed the error.

tests/Kernel/fixtures/rendering.yml
imanoleguskiza: I added three more cases to check for the different render arrays.

upchuk (dismissed):

sergepavle (changes requested):
I think we lost the upgrade path for enabling module smart_trim.
everything else looks good for me.

sergepavle (approved):

upchuk (approved):

PR #443, opened 2020-03-12
OPENEUROPA-3009: Update ECL version.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra (dismissed):

nagyad:
I found this on the components page:
image

Looks like there are no styles for these anymore in ECL. We need to check what to do with these.

nagyad (changes requested):
Rejecting due to missing styles for certain button variants.

ademarco:
It seems that this is not related to upgrade to 2.25, we already had it, I'll fix it anyway as it is perceived as a bug.

ademarco (dismissed):

ademarco (approved):

nagyad (approved):

PR #445, opened 2020-03-17
OPENEUROPA-2860: [POC] Support multiple EU and EC branding

ademarco:


imanoleguskiza:
Functionality was implemented

PR #447, opened 2020-03-19
OPENEUROPA-3020: Style Timeline paragraph.

22Alexandra:
### Description

Style Timeline paragraph.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


oe_theme.theme
```diff
@@ -1357,3 +1357,19 @@ function _oe_theme_trim_textfield($text, int $size = 150) {
...
 }
+
+/**
+ * Implements hook_preprocess_paragraph().
+ */
+function oe_theme_preprocess_paragraph__oe_timeline(array &$variables): void {
+  $paragraph = $variables['paragraph'];
+  $variables['limit'] = $paragraph->get('field_oe_timeline_expand')->value;
+  $items = $paragraph->get('field_oe_timeline')->getValue();
+  foreach ($items as $key => $item) {
```
imanoleguskiza: There doesn't seem to be any transformation going on here, why not simply do $variables['items'] = $paragraph->get('field_oe_timeline')->getValue(); ?

tests/Kernel/Paragraphs/ParagraphsTestBase.php
```diff
@@ -34,6 +34,9 @@ abstract class ParagraphsTestBase extends AbstractKernelTestBase {
     'oe_paragraphs',
     'allowed_formats',
     'typed_link',
+    'oe_paragraphs_timeline',
```
imanoleguskiza: Instead of adding extra dependencies (and adding more complexity to tests that don't need it), I think we should create a TimelineParagraphTest class and extend the dependencies there.

tests/Kernel/Paragraphs/ParagraphsTestBase.php
```diff
@@ -45,7 +48,13 @@ protected function setUp() {
     $this->installEntitySchema('paragraph');
     $this->installEntitySchema('file');
     $this->installSchema('file', ['file_usage']);
-    $this->installConfig(['oe_paragraphs', 'filter']);
+    $this->installConfig([
```
imanoleguskiza: Same as before, lets do this in a dedicate class to not polute the rest of the tests.

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

tests/Kernel/Paragraphs/TimelineParagraphTest.php
```diff
@@ -0,0 +1,117 @@
...
+        ],
+      ],
+    ]);
+
+    $paragraph->save();
+    $html = $this->renderParagraph($paragraph);
+    $crawler = new Crawler($html);
+
+    $this->assertCount(1, $crawler->filter('ol.ecl-timeline2'));
+    $this->assertCount(7, $crawler->filter('ol.ecl-timeline2 li.ecl-timeline2__item'));
```
nagyad, line 83: I see 6 items above, why do you assert for 7?

nagyad: Below you assert for 6 correctly.

nagyad: I see the button has the same class. So it's 6 items + button.

tests/Kernel/Paragraphs/TimelineParagraphTest.php
```diff
@@ -0,0 +1,117 @@
...
+    $this->assertEquals('Show 3 more items', trim($crawler->filter('button.ecl-button.ecl-button--secondary.ecl-timeline2__toggle span.ecl-button__container span.ecl-button__label')->text()));
+
+    $paragraph->set('field_oe_timeline_expand', '6');
+    $paragraph->save();
+    $html = $this->renderParagraph($paragraph);
+    $crawler = new Crawler($html);
+
+    $this->assertCount(6, $crawler->filter('ol.ecl-timeline2 li.ecl-timeline2__item'));
+    $this->assertCount(0, $crawler->filter('ol.ecl-timeline2 li.ecl-timeline2__item.ecl-timeline2__item--collapsed'));
+    $this->assertCount(0, $crawler->filter('ol.ecl-timeline2 li.ecl-timeline2__item.ecl-timeline2__item--toggle'));
```
nagyad, line 114: Test misses one case. We should test when the button is not collapsed and it stays on the page with the text "Hide 3 items".

nagyad (changes requested):

nagyad (dismissed):
After discussing we agreed that the effort to test the library behavior is not worth to move this to FunctionalJavascript that is considerably slower as well. The test is fine. The front end we will test by behat once it reaches EWCMS.

brummbar (dismissed):

brummbar (approved):

imanoleguskiza (approved):

PR #449, opened 2020-03-23
OPENEUROPA-3052: Remove ecl-editor wrapper from timeline tpl.

22Alexandra:
### Description

Remove ecl-editor wrapper from timeline tpl.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/templates/field/timeline.html.twig
+++ b/templates/field/timeline.html.twig
@@ -4,10 +4,8 @@
  * Theme override for the timeline field template.
  */
 #}
-<div class="ecl-editor">
 {{ pattern('timeline', {
   'limit': limit,
   'button_label': show_more,
   'items': items,
 }) }}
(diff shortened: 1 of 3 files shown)


nagyad (dismissed):

ademarco (changes requested):
This will prevent HTML in content field to be styled correctly.

ademarco:
To fix it we need to wrap each description with `ecl-editor` as timeline expects single elements to receive ECL classes in order to be properly styled (like `ecl-link`).

See below the desired result using Firefox dev console.

image

ademarco (approved):

nagyad (approved):

PR #453, opened 2020-03-26
OPENEUROPA-3057: Use skos for Event types.

nagyad:


sergepavle (dismissed):

imanoleguskiza (dismissed):

modules/oe_theme_content_event/src/Plugin/PageHeaderMetadata/EventContentType.php
```diff
@@ -39,7 +39,7 @@ public function getMetadata(): array {
       ->getOptionsProvider('value', $node);
 
     $metadata['metas'] = [
-      $provider->getPossibleOptions()[$node->get('oe_event_type')->value],
+      $provider->getPossibleOptions()[$node->get('oe_event_type')->target_id],
```
ademarco: This is wrong here, we need to render the label as the field type is not `list_string` anymore, check how we do it here with a similar field type `\Drupal\oe_theme_content_event\Plugin\ExtraField\Display\DetailsExtraField::getRenderableSubject`

ademarco (changes requested):

modules/oe_theme_content_event/src/Plugin/PageHeaderMetadata/EventContentType.php
```diff
@@ -32,14 +32,16 @@ public function applies(): bool {
...
 
-    // Use business logic from "list_default" formatter.
     $node = $this->getNode();
-    $provider = $node->get('oe_event_type')->getFieldDefinition()
-      ->getFieldStorageDefinition()
-      ->getOptionsProvider('value', $node);
+    $metadata['metas'] = [];
 
+    if ($node->get('oe_event_type')->isEmpty()) {
+      return $metadata;
```
ademarco: What if we have summary but no meta? we don't display the summary?

ademarco (changes requested):

modules/oe_theme_content_event/src/Plugin/PageHeaderMetadata/EventContentType.php
```diff
@@ -32,15 +32,15 @@ public function applies(): bool {
...
-    $provider = $node->get('oe_event_type')->getFieldDefinition()
-      ->getFieldStorageDefinition()
-      ->getOptionsProvider('value', $node);
+    $metadata['metas'] = [];
 
-    $metadata['metas'] = [
-      $provider->getPossibleOptions()[$node->get('oe_event_type')->value],
-    ];
+    if (!$node->get('oe_event_type')->isEmpty()) {
+      $translated_value = \Drupal::service('entity.repository')->getTranslationFromContext($node->get('oe_event_type')->entity);
```
ademarco: Inject this service.

ademarco (changes requested):

modules/oe_theme_content_event/src/Plugin/PageHeaderMetadata/EventContentType.php
ademarco: Ok, now that it's injected let's use it.

ademarco (changes requested):

modules/oe_theme_content_event/src/Plugin/PageHeaderMetadata/EventContentType.php
```diff
@@ -32,15 +75,15 @@ public function applies(): bool {
   public function getMetadata(): array {
     $metadata = parent::getMetadata();
 
-    // Use business logic from "list_default" formatter.
     $node = $this->getNode();
-    $provider = $node->get('oe_event_type')->getFieldDefinition()
-      ->getFieldStorageDefinition()
-      ->getOptionsProvider('value', $node);
+    $metadata['metas'] = [];
```
ademarco: Let's remove this, it's not needed. Also isn't oe_event_type required? We weren't checking for it's presence before:
```php
    $metadata['metas'] = [
      $provider->getPossibleOptions()[$node->get('oe_event_type')->value],
    ];
```

ademarco (changes requested):

modules/oe_theme_content_event/src/Plugin/PageHeaderMetadata/EventContentType.php
nagyad: Yes, but if we create an event with api, like behat without type it will fail.

ademarco (approved):

imanoleguskiza (approved):

PR #454, opened 2020-03-27
OPENEUROPA-3059: Improve registration button messages.

nagyad:


sergepavle (dismissed):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/RegistrationButtonExtraField.php
```diff
@@ -99,12 +100,22 @@ public function viewElements(ContentEntityInterface $entity) {
...
-      $date_diff = $this->dateFormatter->formatDiff($this->requestTime, $datetime_start->getTimestamp());
-      $build['#description'] = $this->t('Registration will open in @time_left. You can register from @start_date, until @end_date.', [
-        '@time_left' => $date_diff,
-        '@start_date' => $this->dateFormatter->format($datetime_start->getTimestamp(), 'oe_event_date_hour'),
-        '@end_date' => $this->dateFormatter->format($datetime_end->getTimestamp(), 'oe_event_date_hour'),
-      ]);
+      $request_datetime = DrupalDateTime::createFromTimestamp($this->requestTime);
+      $date_diff = $request_datetime->diff($datetime_start);
+      if ($date_diff->days === 0) {
+        $build['#description'] = $this->t('Registration will close today, @end_date.', [
```
imanoleguskiza: I don't think this message is very useful as it is. It will be shown to a user when it arrives at an event the day the registration opens but at an earlier hour. Therefore I think we should inform the user when they can expect to be able to register: "Registration will open today at @hour." sounds more appropriate.

imanoleguskiza (dismissed):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/RegistrationButtonExtraField.php
nagyad: Yes, you are right. I got confused. Good catch!

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/RegistrationButtonExtraField.php
```diff
@@ -99,12 +100,22 @@ public function viewElements(ContentEntityInterface $entity) {
...
       $datetime_end = $event->getRegistrationEndDate();
-      $date_diff = $this->dateFormatter->formatDiff($this->requestTime, $datetime_start->getTimestamp());
-      $build['#description'] = $this->t('Registration will open in @time_left. You can register from @start_date, until @end_date.', [
-        '@time_left' => $date_diff,
-        '@start_date' => $this->dateFormatter->format($datetime_start->getTimestamp(), 'oe_event_date_hour'),
-        '@end_date' => $this->dateFormatter->format($datetime_end->getTimestamp(), 'oe_event_date_hour'),
-      ]);
+      $request_datetime = DrupalDateTime::createFromTimestamp($this->requestTime);
+      $date_diff = $request_datetime->diff($datetime_start);
+      if ($date_diff->days === 0) {
```
imanoleguskiza: This condition has a problem because if the time remaining is less than 24h it will trigger, even though it might be a different day.

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

sergepavle (dismissed):

sergepavle (approved):

imanoleguskiza (approved):

PR #458, opened 2020-04-06
OPENEUROPA-2950: Refactor theme trimming function as a Twig filter.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +493,59 @@ public function toEclIcon(array $context, $icon): array {
...
+   * @param string|array|object $text
+   *   The text to be trimmed.
+   *
+   * @return string|array
+   *   The trimmed text.
+   */
+  public function trimTextfield(array $context, $text) {
+    $size = $context['length'] ?? NULL;
+
+    // Return original text if length not available.
```
nagyad: Return original text if length is not available.

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
```diff
@@ -300,4 +303,110 @@ public function toEclIconProvider(): array {
     ];
   }
 
+  /**
+   * Tests trimming of textfield.
+   *
+   * @param mixed $textfield
+   *   The string, object or render array.
+   * @param array $context
+   *   The icon array to be rendered.
```
nagyad: Please update param comment.

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
```diff
@@ -300,4 +303,110 @@ public function toEclIconProvider(): array {
...
+   * @param mixed $expected_result
+   *   The icon array to be rendered.
```
nagyad: Please update param comment.

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
```diff
@@ -300,4 +303,110 @@ public function toEclIconProvider(): array {
...
+    $context['input_data'] = $textfield;
+    $result = $this->twig->render("{{ trim_textfield(input_data)|json_encode() }}", $context);
+    $this->assertEquals(Html::escape(Json::encode($expected_result)), $result);
+  }
+
+  /**
+   * Returns test cases for ::testTrimTextfield().
+   *
+   * @return array[]
+   *   An icon array.
```
nagyad: Please update parameter comment.

nagyad (changes requested):

nagyad (dismissed):

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
```diff
@@ -300,4 +303,110 @@ public function toEclIconProvider(): array {
...
+   * @param array $context
+   *   The context array.
+   * @param mixed $expected_result
+   *   The expected result.
+   *
+   * @covers ::trimTextfield
+   * @dataProvider trimTextfieldProvider
+   */
+  public function testTrimTextfield($textfield, array $context, $expected_result): void {
+    $context['input_data'] = $textfield;
```
22Alexandra: The function needs the context for the length, not for the input data. Please set the length in the context from the data provider.

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
```diff
@@ -300,4 +303,110 @@ public function toEclIconProvider(): array {
...
+    $result = $this->twig->render("{{ trim_textfield(input_data)|json_encode() }}", $context);
```
22Alexandra: Call the function using the $textfield from the data provider (as the function uses the parameter) instead of setting it in the $context.

22Alexandra (changes requested):

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
```diff
@@ -300,4 +303,110 @@ public function toEclIconProvider(): array {
...
+   *   The string, object or render array.
+   * @param array $context
+   *   The context array.
+   * @param mixed $expected_result
+   *   The expected result.
+   *
+   * @covers ::trimTextfield
+   * @dataProvider trimTextfieldProvider
+   */
+  public function testTrimTextfield($textfield, array $context, $expected_result): void {
```
22Alexandra: No need to send the $context, you can send only the $length and then set it in the $context array.

22Alexandra (changes requested):

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
sergepavle: Could you explain more.

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
22Alexandra: Nevermind..better send the context already prepared with the length.

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
22Alexandra: Talked to Sergii and the result would not be the same.

22Alexandra (dismissed):

upchuk (dismissed):
Not sure this is the right approach. Will check. Please don't merge.

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -55,6 +57,7 @@ public function getFilters(): array {
   public function getFunctions(): array {
     return [
       new \Twig_SimpleFunction('to_ecl_icon', [$this, 'toEclIcon'], ['needs_context' => TRUE]),
+      new \Twig_SimpleFunction('trim_textfield', [$this, 'trimTextfield'], ['needs_context' => TRUE]),
```
ademarco: Let's not use needs_context here as we don't really need it, we just need to pass the length as a parameter.

Also let's call it `smart_trim` as it's really the implementation of that behavior.

ademarco (changes requested):

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +493,57 @@ public function toEclIcon(array $context, $icon): array {
...
+   *
+   * The method expects to receive one of the following: a string,
+   * a render array or an object that implements MarkupInterface.
+   * It will then try to trim the associated string to a size defined when
+   * invoking the method.
+   *
+   * @param string|array|object $text
+   *   The text to be trimmed.
+   * @param int|null $size
+   *   The length.
```
ademarco: Call this 
```
   * @param int $limit
   *   Amount of text to allow.
```
As in `truncateChars()`.

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -55,6 +57,7 @@ public function getFilters(): array {
   public function getFunctions(): array {
     return [
       new \Twig_SimpleFunction('to_ecl_icon', [$this, 'toEclIcon'], ['needs_context' => TRUE]),
+      new \Twig_SimpleFunction('smart_trim', [$this, 'trimTextfield']),
```
ademarco: Call the method `smartTrim()` for consistency.

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +493,57 @@ public function toEclIcon(array $context, $icon): array {
...
+    // Return original text if length is not available.
+    if ($size === NULL) {
+      return $text;
+    }
+
+    $truncate = new TruncateHTML();
+
+    if (is_string($text) || $text instanceof MarkupInterface) {
+      // If $text is a string or a Markup object, trim and return the string.
+      return $output = $truncate->truncateChars($text, $size);
```
ademarco: You don't need this assignment, just return:
```
       return $truncate->truncateChars($text, $size);
```

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +493,57 @@ public function toEclIcon(array $context, $icon): array {
...
+      return $text;
+    }
+
+    if (isset($text['#plain_text'])) {
+      $text['#plain_text'] = $truncate->truncateChars($text['#plain_text'], $size);
+      return $text;
+    }
+
+    if (isset($text['#type']) && $text['#type'] == 'processed_text') {
+      $text['#text'] = $truncate->truncateChars($text['#text'], $size);
```
ademarco: This might contain in-line filters that need to be rendered before right? Like embedded media or so.

In general we should render a render array before truncating it, even if it's `#markup`, as it might have sensible `#prefix` or `#suffix` that can contain anything, including readable text.

ademarco (changes requested):
I think this is not the right way of doing it, the output of that function must be a simple string, meaning that rendering should happens before trimming it.

We can also leave this to the calling template, meaning, first it must call `|render` and then `|smart_trim`, so it can be done case by case by the calling context. This would simplify this PR considerably.

Also we need to turn this into a filter, and not a function, so we can pipe it. Meaning `foo|render|smart_trim(10)`

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
sergepavle: done

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
sergepavle: done

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
sergepavle: fixed

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +494,37 @@ public function toEclIcon(array $context, $icon): array {
...
+  /**
+   * Trims the contents of a text field.
+   *
+   * The method expects to receive a string (already rendered).
+   * Inside a template we can use '|render' for passing to filter
+   * only string like in example: {{ long_text|render|smart_trim(10) }}.
+   *
+   * @param mixed $text
+   *   The text to be trimmed.
+   * @param mixed $limit
```
22Alexandra: The $limit should be always int.

22Alexandra (changes requested):

22Alexandra (dismissed):

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +494,37 @@ public function toEclIcon(array $context, $icon): array {
...
+
+    if (is_string($text) || $text instanceof MarkupInterface) {
+      // If $text is a string or a Markup object, trim
+      // and return the Markup object.
+      return Markup::create($truncate->truncateChars($text, $limit));
+    }
+
+    // Return the unchanged $text since we don't support other scenarios.
+    return $text;
+  }
```
ademarco, line 528: Let's always return a string, not at times a string and at time a markup object, I'd simply do:

```php
public function smartTrim($text, $limit = NULL) {
  // Truncate only if if length is not available or $text is string-like.
  if ($limit !== NULL && (is_string($text) || $text instanceof MarkupInterface)) {
    $truncate = new TruncateHTML();
    return $truncate->truncateChars($text, $limit);
  }

  return $text;
}
```

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +494,37 @@ public function toEclIcon(array $context, $icon): array {
...
+    if ($limit === NULL) {
+      return $text;
+    }
+
+    $truncate = new TruncateHTML();
+
+    if (is_string($text) || $text instanceof MarkupInterface) {
+      // If $text is a string or a Markup object, trim
+      // and return the Markup object.
+      return Markup::create($truncate->truncateChars($text, $limit));
```
ademarco: If we get a string we shouldn't turn that into a markup, as this would mean that a string with `<javascript>...</javascript>` would suddenly become executable code.

Let's simply do
```php
public function smartTrim($text, $limit) {
  $truncate = new TruncateHTML();
  return $truncate->truncateChars($text, $limit);
}
```
truncateChars handles all cases anyway, then let's have a test that ensure that passing a MarkupInterface behaves correctly.

ademarco (changes requested):

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
ademarco: That didn't work because then the string would be stripped when rendered. We then changed approach, the only way is to respect the input.

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +510,72 @@ public function toEclIcon(array $context, $icon): array {
...
+   *   Current Twig environment.
+   * @param mixed $input
+   *   Input to be trimmed, it can be a string, an object or a render array.
+   * @param int $limit
+   *   Amount of text to allow.
+   *
+   * @return mixed
+   *   The trimmed output.
+   */
+  public function smartTrim(\Twig_Environment $env, $input, $limit) {
```
brummbar, line 526: Can we put the type hinting on the limit?

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +510,72 @@ public function toEclIcon(array $context, $icon): array {
...
+    // Bubbles Twig template argument's cacheability & attachment metadata.
+    $this->bubbleArgMetadata($input);
+
+    // Render $input, so we can truncate its output.
```
brummbar: This comment is probably old as we don't render here anymore.

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -490,4 +510,72 @@ public function toEclIcon(array $context, $icon): array {
...
+    $truncate = new TruncateHTML();
+
+    // If $input is a Markup object, trim it and return it as such.
```
brummbar:
```suggestion
    // If input is a Markup object, trim it and return it as such.
```

brummbar (changes requested):

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
ademarco: I'm a bit hesitant on this as it might come in a form of a template variable as a string.

brummbar: Yep indeed, let's leave it like this.

brummbar (approved):

ademarco (approved):

PR #459, opened 2020-04-08
OPENEUROPA-3095: Adapt theme to support full width elements.

22Alexandra:


nagyad (dismissed):

oe_theme.theme
```diff
@@ -1484,3 +1484,20 @@ function oe_theme_preprocess_paragraph__oe_banner(array &$variables): void {
 
   $cacheability->applyTo($variables);
 }
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function oe_theme_theme_suggestions_page_alter(array &$suggestions, array $variables) {
+  if ($node = \Drupal::routeMatch()->getParameter('node')) {
```
ademarco: Better checking if node is an actual node object, since we are gonna invoke a method on it below:

```php
  $node = \Drupal::routeMatch()->getParameter('node');
  if ($node instanceof \Drupal\node\NodeInterface) {
```

Also, will this work when asking for a revision of a node? Now sure if in that case the route has an element `node` containing the node object, we should manually check that.

oe_theme.theme
```diff
@@ -1484,3 +1484,20 @@ function oe_theme_preprocess_paragraph__oe_banner(array &$variables): void {
...
+    $content_type = $node->bundle();
+    $suggestions[] = 'page__' . $content_type;
+  }
+}
+
+/**
+ * Implements hook_preprocess_page().
+ */
+function oe_theme_preprocess_page(array &$variables): void {
+  $variables['main_attributes'] = new Attribute(['class' => ['ecl-u-pv-2xl ecl-u-pv-md-3xl']]);
```
ademarco: One class per array element, so extending themes can remove one but not the other.

templates/layout/page.html.twig
```diff
@@ -33,7 +33,7 @@
 {% endblock %}
 
 {% block content %}
-<main class="ecl-u-pv-xl" id="main-content">
+<main{{ main_attributes }}>
```
ademarco, line 36: `id` got lost in the process, you should add it in the preprocess.

ademarco (changes requested):

ademarco:
I'm also gonna add the following changes in this PR, so that removing top margin will be much easier in EWCMS:

image

upchuk (dismissed):

upchuk (approved):

ademarco (approved):

PR #463, opened 2020-04-15
OPENEUROPA-3096: Update banner paragraph and patterns to support full width.

22Alexandra:


templates/patterns/banners/banner_hero.ui_patterns.yml
```diff
@@ -57,3 +57,9 @@ banner_hero:
       description: "Define if banner should be centered. Deprecated, use the Alignment field instead."
       preview:
       escape: false
+    full_width:
+      type: "boolean"
+      label: "Full width"
+      description: "Define if the banner should be extended full width."
```
nagyad: > Define if the banner should be extended to full width.

templates/patterns/banners/banner_page.ui_patterns.yml
```diff
@@ -46,3 +46,9 @@ banner_page:
       label: "Alignment"
       description: "The Page banner can be 'center' or 'left' (aligned left)."
       preview: "left"
+    full_width:
+      type: "boolean"
+      label: "Full width"
+      description: "Define if the banner should be extended full width."
```
nagyad: > Define if the banner should be extended to full width.

oe_theme.theme
```diff
@@ -1423,6 +1423,10 @@ function oe_theme_preprocess_paragraph__oe_banner(array &$variables): void {
   $variables['url'] = Url::fromUri($link['uri']);
   $variables['label'] = $link['title'];
 
+  if ($paragraph->get('field_oe_banner_full_width')->value === 'yes') {
+    $variables['full_width'] = TRUE;
+  }
```
nagyad: I'm not sure about this. I know that in the previous oe_paragraphs ticket it was asked to do list_text with Yes/No, but actually looking at the solution here it's meant to be bool. Using strings for deciding a bool is not correct. I checked with @ademarco and we agreed to have a checkbox instead on the paragraph. Please apply the change on the paragraphs side within this ticket so we make this right. :)

tests/Kernel/fixtures/rendering.yml
```diff
@@ -176,12 +177,14 @@
       label: "Subscribe"
       description: "Banner description"
       centered: false
+      full_width: true
```
nagyad, line 180: Here you are using bool, it is more relevant.

nagyad (changes requested):

oe_theme.theme
22Alexandra: Did the change but boolean field returns 1/0 so it's the same approach.

oe_theme.theme
```diff
@@ -1423,7 +1423,7 @@ function oe_theme_preprocess_paragraph__oe_banner(array &$variables): void {
   $variables['url'] = Url::fromUri($link['uri']);
   $variables['label'] = $link['title'];
 
-  if ($paragraph->get('field_oe_banner_full_width')->value === 'yes') {
+  if ($paragraph->get('field_oe_banner_full_width')->value === '1') {
     $variables['full_width'] = TRUE;
```
nagyad: Instead of this, you could just assign the value as bool.

nagyad (changes requested):

nagyad (dismissed):

oe_theme.theme
```diff
@@ -1495,3 +1497,21 @@ function oe_theme_preprocess_paragraph__oe_banner(array &$variables): void {
 
   $cacheability->applyTo($variables);
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_pattern_banner_hero(array &$variables): void {
+  if ($variables['full_width'] === TRUE) {
+    $variables['div_attributes'] = new Attribute(['class' => 'ecl-page-banner--full-width']);
```
ademarco: Let's call this: `container_attributes`

ademarco (changes requested):

ademarco (dismissed):

nagyad:
After oe_paragraphs was merged, we need to fix the composer patch.

nagyad (dismissed):

ademarco:
I'm also gonna add the following changes in this PR, so that removing top margin will be much easier in EWCMS:

image

ademarco (approved):

upchuk (approved):

PR #465, opened 2020-04-20
OPENEUROPA-3091: Adapt banner preprocess for optional link.

22Alexandra:


nagyad (dismissed):

nagyad (dismissed):

templates/patterns/banners/pattern-banner-hero.html.twig
```diff
@@ -5,7 +5,11 @@
  */
 #}
 
-{% set _link = {'path' : url, 'label' : label, icon_position: 'after'} -%}
+{% if url is not empty and label is not empty %}
```
imanoleguskiza, line 8: We can simplify this like this: 

```
{% if url is not empty and label is not empty %}
  {% set _link = {'link' : {'path' : url, 'label' : label, icon_position: 'after'}} -%}
  {% set _link = _link|merge({'icon' : {'path' : ecl_icon_path, 'type' : "ui", 'name' : "corner-arrow", 'transform' : "rotate-90", 'size' : "xs"}}) %}
{% endif %}
```
That way we don't need to declare link_

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

nagyad (dismissed):

imanoleguskiza (approved):

nagyad (approved):

PR #466, opened 2020-04-22
Increase mysql max packet

Tiago-FM:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


ademarco:
@Tiago-FM @voidtek was this just a test? Can we close it? We never have mysql going away on components as they have quite a small footprint.

imanoleguskiza:
Fix is already present

PR #489, opened 2020-05-26
OPENEUROPA-3006: Add support for event listing to patterns

ademarco:


nagyad (dismissed):

tests/Kernel/fixtures/rendering.yml
```diff
@@ -1717,6 +1759,34 @@
...
+- array:
+    '#type': pattern
+    '#id': featured_item
+    '#variant': "default"
+    '#fields':
+      title: "Example title"
+      description: "Example description"
+  assertions:
+    count:
+      "footer.ecl-card__footer": 0
```
imanoleguskiza: We could just probably add this assertion to one of the previous tests and save creating one array just for this.

templates/patterns/list_item/list_item.ui_patterns.yml
```diff
@@ -23,6 +20,18 @@ list_item:
     thumbnail_secondary:
       label: "List item: thumbnail with secondary image"
       description: "Thumbnail list item, contains a secondary image, metadata information and a linked title."
+    date:
+      label: "List item: event"
```
imanoleguskiza, line 24: Its weird to call this pattern "event" since it doesn't necessarily involve an event.

imanoleguskiza (changes requested):

templates/patterns/list_item/list_item.ui_patterns.yml
nagyad: Probably it's because it is used on an event listing item at the moment.

ademarco: You are right @imanoleguskiza I'll change their name to past date, upcoming date etc.

tests/Kernel/fixtures/rendering.yml
ademarco: Ok I'll move it up.

templates/patterns/list_item/list_item.ui_patterns.yml
ademarco: Still ongoing date doesn't make much sense :D

imanoleguskiza: Actually you are right, the date blocks are just for events: https://ec.europa.eu/component-library/ec/components/date-block/usage/

ademarco: We agreed with @imanoleguskiza to keep it event. resolving.

imanoleguskiza (approved):

nagyad (approved):

PR #494, opened 2020-06-03
OPENEUROPA-3180: Fix venue active language.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DetailsExtraField.php
+++ b/modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DetailsExtraField.php
@@ -6,8 +6,11 @@ namespace Drupal\oe_theme_content_event\Plugin\ExtraField\Display;
 
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityRepositoryInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\oe_content_entity_venue\Entity\VenueInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Extra field displaying event details as a list of icons and text.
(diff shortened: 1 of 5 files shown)


modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DetailsExtraField.php
```diff
@@ -119,6 +119,11 @@ protected function addRenderableLocation(array &$build, ContentEntityInterface $
       return;
     }
 
+    // Inherit language from parent entity.
+    if ($venue->hasTranslation($entity->language()->getId())) {
+      $venue = $venue->getTranslation($entity->language()->getId());
+    }
+
```
ademarco: Instead of this inject the entity repository service and do
```php
  protected function addRenderableLocation(array &$build, ContentEntityInterface $entity): void {
    if ($entity->get('oe_event_venue')->isEmpty()) {
      return;
    }

    $venue = $entity->get('oe_event_venue')->entity;
    if (!$venue instanceof VenueInterface) {
      return;
    }

    $venue = $this->entityRepository->getTranslationFromContext($venue);
    $cacheability = CacheableMetadata::createFromRenderArray($build);
```

ademarco (changes requested):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DetailsExtraField.php
sergepavle: done

modules/oe_theme_helper/tests/src/Kernel/Plugin/FieldFormatter/AddressInlineFormatterTest.php
```diff
@@ -2,9 +2,9 @@
 
 declare(strict_types = 1);
 
-namespace Drupal\Tests\oe_theme_helper\Kernel\Plugin\Field\FieldFormatter;
+namespace Drupal\Tests\oe_theme_helper\Kernel\Plugin\FieldFormatter;
```
ademarco: Why we moved the file to this namespace? It was supposed to mirror 1 to 1 the class that it tests, which is `\Drupal\oe_theme_helper\Plugin\Field\FieldFormatter\AddressInlineFormatter`, so the namespace and file location were correcnt.

modules/oe_theme_helper/tests/src/Kernel/Plugin/FieldFormatter/AddressInlineFormatterTest.php
```diff
@@ -31,7 +31,7 @@ protected function setUp(): void {
    * Tests formatting of address.
    */
   public function testInlineFormatterAddress() {
-    $entity = EntityTest::create([]);
+    $entity = EntityTestMul::create([]);
```
ademarco: Any reason why we changed the entity type here? We don't use anything of `EntityTestMul` in the test.

ademarco (changes requested):

modules/oe_theme_helper/tests/src/Kernel/Plugin/FieldFormatter/AddressInlineFormatterTest.php
sergepavle: I eventually have found that we have tests for field formatted in different folders:
https://github.com/openeuropa/oe_theme/tree/2.x/modules/oe_theme_helper/tests/src/Kernel/Plugin
For some reasons I moved address inline formatter. If you think that better to move tests of other formatter I could do. Normally we are not following mirroring of test classes.

modules/oe_theme_helper/tests/src/Kernel/Plugin/FieldFormatter/AddressInlineFormatterTest.php
sergepavle: EntityTest - it is not translatable entity type
But EntityTestMul is translatable.
I have found that not translatable entity type does not pass to formatter class a langcode (passing null).

sergepavle: Because we still use \Drupal\Tests\address\Kernel\Formatter\FormatterTestBase as parent class. But it is changed in the patch of contribs module.
I did this to have the possibility to test address fields with translatable entity types.
EntityTest - do not allow to do this.

modules/oe_theme_helper/tests/src/Kernel/Plugin/FieldFormatter/AddressInlineFormatterTest.php
sergepavle: I reverted moving of `AddressInlineFormatterTest` and moved to the same folder the `SocialMediaLinksFormatterTest`  class.

modules/oe_theme_helper/tests/src/Kernel/Plugin/FieldFormatter/AddressInlineFormatterTest.php
ademarco: Ok but here, in this test, we are not testing its translatablility right? That's why I was thinking to just use EntityTest

ademarco: Ah I see it, it's because of `$this->renderEntityFields($entity, $this->display);`. Resolving.

ademarco (approved):

22Alexandra (approved):

PR #497, opened 2020-06-12
OPENEUROPA-3180: Fix issue with failed patch for oe_content.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


composer.json
```diff
@@ -33,14 +33,15 @@
         "drupal/styleguide": "~1.0-alpha3",
         "drupal/typed_link": "~1.0",
         "drush/drush": "~9.0",
+        "openeuropa/rdf_skos": "~0.8",
```
ademarco: We can remove this as it has been fixed on oe_content now.

composer.json
```diff
@@ -33,14 +33,15 @@
...
         "easyrdf/easyrdf": "0.10.0-alpha.1 as 0.9.1",
         "friends-of-behat/service-container-extension": "^1.0",
         "guzzlehttp/guzzle": "~6.3",
         "nikic/php-parser": "~3.0",
         "openeuropa/behat-transformation-context": "~0.1",
         "openeuropa/code-review": "~1.1",
         "openeuropa/drupal-core-require-dev": "^8.7",
-        "openeuropa/oe_content": "dev-master",
+        "openeuropa/oe_content": "1.8.1",
```
ademarco: We can also discard this as that has been fixed on oe_content master now, right? ~I think this PR is not necessary anymore.~ We can update typed_link in the scope of this PR.

ademarco (changes requested):

composer.json
```diff
@@ -33,14 +33,15 @@
         "drupal/styleguide": "~1.0-alpha3",
         "drupal/typed_link": "~1.0",
```
ademarco: We need this to be moved to ~1.1.

ademarco (changes requested):

composer.json
sergepavle: Updated

composer.json
sergepavle: Done

ademarco (approved):

hernani (approved):

22Alexandra (approved):

PR #503, opened 2020-06-17
OPENEUROPA-3282: Add oe_contact_forms integration.

drishu:
### Description

Integrate oe_contact_forms for themeing.

### Change log

- Added: oe_theme_contact_forms companion module
- Changed:
  - add oe_contact_forms to composer.json
  - preprocess_form_element to add vertical spacing between elements
  - hook_preprocess_input to add ecl clases to tel input fields
  - runner.yml to enable the new module
  - checkbox composition to add required class if set

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "openeuropa/oe_paragraphs": "dev-master",
         "openeuropa/oe_search": "~1.0",
         "openeuropa/oe_webtools": "dev-master",
-        "openeuropa/oe_contact_forms": "dev-OPENEUROPA-3282B",
+        "openeuropa/oe_contact_forms": "dev-master",
         "openeuropa/task-runner": "^1.0.0-beta5",
         "phpunit/phpunit": "~6.0",
         "symfony/dom-crawler": "~3.4"
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,7 @@
         "openeuropa/oe_paragraphs": "dev-master",
         "openeuropa/oe_search": "~1.0",
         "openeuropa/oe_webtools": "dev-master",
+        "openeuropa/oe_contact_forms": "dev-OPENEUROPA-3282B",
         "openeuropa/task-runner": "^1.0.0-beta5",
         "phpunit/phpunit": "~6.0",
         "symfony/dom-crawler": "~3.4"
--- a/modules/oe_theme_contact_forms/README.md
+++ b/modules/oe_theme_contact_forms/README.md
@@ -1,0 +1,15 @@
+# OpenEuropa Contact Forms companion module
+
+This module is a theming companion module to the [OpenEuropa Contact Forms](https://github.com/openeuropa/oe_contact_forms) module.
+It provides the logic needed to theme the corporate contact forms.
+
+## Installation
+
+Make sure you you have read the OpenEuropa Contact Forms's [README.md](https://github.com/openeuropa/oe_contact_forms/blob/master/README.md)
+before enabling this module.
+
+After enabling this module make sure you assign the following permissions to the anonymous user role, so visitors can
+correctly access corporate contact forms.
+
+- `Access corporate contact form`
+- `View published SKOS Concept entities`
--- a/modules/oe_theme_contact_forms/README.md
+++ b/modules/oe_theme_contact_forms/README.md
@@ -1,12 +1,11 @@
 # OpenEuropa Contact Forms companion module
 
-This module is a theming companion module to the [OpenEuropa Contact Forms](https://github.com/openeuropa/oe_contact_forms) module.
+This module is a theming companion module to the [OpenEuropa Contact Forms](https://github.com/openeuropa/oe_contact_forms) component.
 It provides the logic needed to theme the corporate contact forms.
 
 ## Installation
 
-Make sure you you have read the OpenEuropa Contact Forms's [README.md](https://github.com/openeuropa/oe_contact_forms/blob/master/README.md)
-before enabling this module.
+Make sure you have read the OpenEuropa Contact Forms's [README.md](https://github.com/openeuropa/oe_contact_forms/blob/master/README.md).
 
 After enabling this module make sure you assign the following permissions to the anonymous user role, so visitors can
 correctly access corporate contact forms.
--- a/modules/oe_theme_contact_forms/oe_theme_contact_forms.module
+++ b/modules/oe_theme_contact_forms/oe_theme_contact_forms.module
@@ -99,7 +99,7 @@ function oe_theme_contact_forms_preprocess_status_messages(&$variables) {
     foreach ($fields as $field_name) {
       /** @var \Drupal\Core\Field\FieldItemList $field */
       $field = $contact_message->get($field_name);
-      if ($field->isEmpty() || !$field->access()){
+      if ($field->isEmpty() || !$field->access()) {
         continue;
       }
 
--- a/modules/oe_theme_contact_forms/oe_theme_contact_forms.module
+++ b/modules/oe_theme_contact_forms/oe_theme_contact_forms.module
@@ -106,6 +106,11 @@ function oe_theme_contact_forms_preprocess_status_messages(&$variables) {
       }
 
       $field_view = $field->view($display_settings);
+
+      if (!isset($field_view['#title'])) {
+        continue;
+      }
+
       $items[] = [
         'label' => $field_view['#title'],
         'body' => $field_view,
--- a/modules/oe_theme_contact_forms/oe_theme_contact_forms.module
+++ b/modules/oe_theme_contact_forms/oe_theme_contact_forms.module
@@ -21,23 +21,29 @@ function oe_theme_contact_forms_form_contact_message_form_alter(&$form, FormStat
   /** @var \Drupal\contact\Entity\ContactForm $contact_form */
   $contact_form = ContactForm::load($contact_message->bundle());
 
+  $is_corporate_form = (boolean) $contact_form->getThirdPartySetting('oe_contact_forms', 'is_corporate_form', FALSE);
+  if (!$is_corporate_form) {
+    // We only care about the corporate contact forms.
+    return;
+  }
+
   // Display header as ecl paragraph.
-  $form['header']['#markup'] = '<p class="ecl-u-type-paragraph">' . $form['header']['#markup'] . '</p>';
-  // Logged in user name should display in its own line.
-  if ($form['name']['#type'] === 'item') {
-    $form['name']['#field_prefix'] = '<div class="ecl-u-mt-xs">';
-    $form['name']['#field_suffix'] = '</div>';
+  if (isset($form['header'])) {
+    $form['header'] = [
+      '#type' => 'inline_template',
+      '#template' => '<p class="ecl-u-type-paragraph">{{ header }}</p>',
+      '#context' => [
+        'header' => $form['header'],
+      ],
+    ];
   }
-  // Logged in user mail should display in its own line.
-  if ($form['mail']['#type'] === 'item') {
-    $form['mail']['#field_prefix'] = '<div class="ecl-u-mt-xs">';
-    $form['mail']['#field_suffix'] = '</div>';
-  }
+
   // Add ecl link classes to the privacy link.
   $privacy_link = $contact_form->getThirdPartySetting('oe_contact_forms', 'privacy_policy', FALSE);
   $privacy_title = t('I have read and agree with the <a class="ecl-u-ml-2xs ecl-link ecl-link--default" href="@link" target="_blank">data protection terms</a></label>', [
     '@link' => $privacy_link,
   ]);
+
   $form['privacy_policy']['#title'] = $privacy_title;
 }
 
@@ -50,7 +56,7 @@ function oe_theme_contact_forms_preprocess_status_messages(&$variables) {
   }
 
   foreach ($variables['message_list']['status'] as $key => $value) {
-    // Target contact_messages.
+    // Target contact_messages which are being rendered in the status message.
     if (!is_array($value) || !isset($value['#contact_message'])) {
       continue;
     }
@@ -66,23 +72,15 @@ function oe_theme_contact_forms_preprocess_status_messages(&$variables) {
       continue;
     }
 
-    $optional_selected = $contact_form->getThirdPartySetting('oe_contact_forms', 'optional_fields', []);
+    $optional_fields = $contact_form->getThirdPartySetting('oe_contact_forms', 'optional_fields', []);
     $items = [];
-    $fields = [
+    $fields = array_merge([
       'name',
       'mail',
       'subject',
       'message',
       'oe_topic',
-    ];
-
-    // Include the optional fields.
-    if (in_array('oe_country_residence', $optional_selected)) {
-      $fields[] = 'oe_country_residence';
-    }
-    if (in_array('oe_telephone', $optional_selected)) {
-      $fields[] = 'oe_telephone';
-    }
+    ], array_values($optional_fields));
 
     // Build the label body pairs for the field_list pattern.
     foreach ($fields as $field_name) {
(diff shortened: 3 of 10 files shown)


modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,107 @@
...
+  $form_object = $form_state->getFormObject();
+  /** @var \Drupal\contact\MessageInterface $contact_message */
+  $contact_message = $form_object->getEntity();
+  /** @var \Drupal\contact\Entity\ContactForm $contact_form */
+  $contact_form = ContactForm::load($contact_message->bundle());
+
+  // Display header as ecl paragraph.
+  $form['header']['#markup'] = '<p class="ecl-u-type-paragraph">' . $form['header']['#markup'] . '</p>';
+  // Loged in user name should display in its own line.
+  if ($form['name']['#type'] == 'item') {
```
nagyad: We use === if we are certain of a value type to avoid type juggling, like in this case.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,107 @@
...
+  /** @var \Drupal\Core\Entity\ContentEntityForm $form_object */
+  $form_object = $form_state->getFormObject();
+  /** @var \Drupal\contact\MessageInterface $contact_message */
+  $contact_message = $form_object->getEntity();
+  /** @var \Drupal\contact\Entity\ContactForm $contact_form */
+  $contact_form = ContactForm::load($contact_message->bundle());
+
+  // Display header as ecl paragraph.
+  $form['header']['#markup'] = '<p class="ecl-u-type-paragraph">' . $form['header']['#markup'] . '</p>';
+  // Loged in user name should display in its own line.
```
nagyad: Typo.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,107 @@
...
+  if ($form['name']['#type'] == 'item') {
+    $form['name']['#field_prefix'] = '<div class="ecl-u-mt-xs">';
+    $form['name']['#field_suffix'] = '</div>';
+  }
+  // Loged in user mail should display in its own line.
```
nagyad: Typo.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,107 @@
...
+ * Module file used for theming the corporate contact forms.
+ */
+
+declare(strict_types = 1);
+
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\contact\Entity\ContactForm;
+
+/**
+ * Implements hook_form_FORM_ID_alter() for contact_message_form().
```
nagyad, line 15: Implements hook_form_FORM_ID_alter(). <-- I would keep it like this.
If you want to document just do it either one line below, or simply as a comment in the function.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,107 @@
...
+  }
+  // Loged in user mail should display in its own line.
+  if ($form['mail']['#type'] == 'item') {
+    $form['mail']['#field_prefix'] = '<div class="ecl-u-mt-xs">';
+    $form['mail']['#field_suffix'] = '</div>';
+  }
+  // Add ecl link classes to the privacy link.
+  $privacy_link = $contact_form->getThirdPartySetting('oe_contact_forms', 'privacy_policy', FALSE);
+  $privacy_title = t('I have read and agree with the <a class="ecl-u-ml-2xs ecl-link ecl-link--default" href="@link" target="_blank">data protection terms</a></label>', [
+    '@link' => $privacy_link,
```
nagyad, line 45: What happens if privacy_link is false as you specified in the getThirdPartySetting()?

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,107 @@
...
+      'name',
+      'mail',
+      'subject',
+      'message',
+      'oe_topic',
+    ];
+
+    // Include the optional fields.
+    if (in_array('oe_country_residence', $optional_selected)) {
+      $fields[] = 'oe_country_residence';
```
nagyad, line 91: I don't know why the mockup shows this element in the status message clickable. But currently is leading to the skos entity that I not correct, also this results in anonymous user can't see the country in the status message.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,107 @@
...
+  $privacy_title = t('I have read and agree with the <a class="ecl-u-ml-2xs ecl-link ecl-link--default" href="@link" target="_blank">data protection terms</a></label>', [
+    '@link' => $privacy_link,
+  ]);
+  $form['privacy_policy']['#title'] = $privacy_title;
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_contact_forms_preprocess_status_messages(&$variables) {
```
nagyad, line 54: I tried to insert some HTML into the message and I pressed preview.
image

nagyad: Same for the status message, I think the content of the field should be escaped during submit.
image

modules/oe_theme_contact_forms/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Installation
+
+Make sure you you have read the OpenEuropa Contact Forms's [README.md](https://github.com/openeuropa/oe_contact_forms/blob/master/README.md)
+before enabling this module.
+
+After enabling this module make sure you assign the following permissions to the anonymous user role, so visitors can
+correctly access corporate contact forms.
+
+- `Access corporate contact form`
```
nagyad, line 13: I created a corporate contact form with alias, and I can access the form as anonymous user without this permission.
image

composer.json
```diff
@@ -49,6 +49,7 @@
         "openeuropa/oe_paragraphs": "dev-master",
         "openeuropa/oe_search": "~1.0",
         "openeuropa/oe_webtools": "dev-master",
+        "openeuropa/oe_contact_forms": "dev-OPENEUROPA-3282",
```
nagyad: This should be put on master after the https://github.com/openeuropa/oe_contact_forms/pull/12 is merged.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,107 @@
...
+  if ($form['mail']['#type'] == 'item') {
+    $form['mail']['#field_prefix'] = '<div class="ecl-u-mt-xs">';
+    $form['mail']['#field_suffix'] = '</div>';
+  }
+  // Add ecl link classes to the privacy link.
+  $privacy_link = $contact_form->getThirdPartySetting('oe_contact_forms', 'privacy_policy', FALSE);
+  $privacy_title = t('I have read and agree with the <a class="ecl-u-ml-2xs ecl-link ecl-link--default" href="@link" target="_blank">data protection terms</a></label>', [
+    '@link' => $privacy_link,
+  ]);
+  $form['privacy_policy']['#title'] = $privacy_title;
```
nagyad, line 48: As anonymous user I can't Send a copy to myself but as logged in user I can. Not sure if it just a confugration.
image

nagyad (changes requested):
I could not find tests that would assert the status message has all the fields filled in with the correct values.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
drishu: in oe_contact_forms reviews it was requested, we should try to be consistent, what do you think ?

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
drishu: privacy_link is mandatory, it will never be false

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
drishu: if you can see the markup as text, that means its escaped no ?

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
drishu: thats the way contact forms core is built (not very well I might add)

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
nagyad: In my opinion such text should not be sent at all, I would strip-tags.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
nagyad: If it was requested, then keep it as is. For me it's fine.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
upchuk: Yes, indeed, SKOS references should not show as links as there is no link to go to.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
upchuk: It is escaped, exactly for the reason @drishu says.

modules/oe_theme_contact_forms/README.md
upchuk: Please check. If it's true, there is an issue in oe_contact_forms.

composer.json
upchuk: I think after we ensure we are more or less done there, we will be able to release and use a tag as wel.

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
drishu: done

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
drishu: fixed

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
drishu: fixed

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
drishu: fixed

modules/oe_theme_contact_forms/README.md
drishu: fixed

composer.json
drishu: fixes where needed so I've put a different branch for now

upchuk (dismissed):

templates/compositions/ec-component-checkbox/checkbox.html.twig
```diff
@@ -82,7 +86,7 @@
     {{ _extra_input_attributes|raw }}
   />
 
-  <label for="{{ _id }}" class="ecl-checkbox__label">
+  <label for="{{ _id }}" class="ecl-checkbox__label {{ _required ? "form-required" }}">
```
ademarco: Let's move the space inside the class string, so that, it not required, we don't print a trailing space, like below:

```
   <label for="{{ _id }}" class="ecl-checkbox__label{{ _required ? " form-required" }}">
```

oe_theme.theme
```diff
@@ -1120,6 +1130,16 @@ function oe_theme_preprocess_input__search(array &$variables): void {
...
 
+/**
+ * Implements hook_preprocess_input().
+ */
+function oe_theme_preprocess_input(array &$variables) {
+  if ($variables['element']['#type'] === 'tel') {
+    $variables['attributes']['class'][] = 'ecl-text-input';
+    $variables['attributes']['size'] = 60;
+  }
+}
```
ademarco: We can use here `oe_theme_preprocess_input__tel()` or, if possible, use `input--tel.html.twig`.

oe_theme.theme
```diff
@@ -496,6 +496,16 @@ function oe_theme_preprocess_form_element(array &$variables): void {
...
   }
+  // Ensure we have class attribute.
+  $variables = array_merge_recursive($variables, [
+    'attributes' => [
+      'class' => [],
+    ],
+  ]);
+  // Add top and bottom margin.
+  if (!in_array('ecl-u-mt-l', $variables['attributes']['class'])) {
+    $variables['attributes']['class'][] = 'ecl-u-mv-m';
```
ademarco: Can't we use `form-element.html.twig` instead?

image

modules/oe_theme_contact_forms/oe_theme_contact_forms.module
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  foreach ($variables['message_list']['status'] as $key => $value) {
+    // Target contact_messages which are being rendered in the status message.
+    if (!is_array($value) || !isset($value['#contact_message'])) {
+      continue;
+    }
+
+    /** @var \Drupal\contact\MessageInterface $contact_message */
+    $contact_message = $value['#contact_message'];
```
ademarco, line 70: I would also add here a check like `$value['#contact_message'] instanceof MessageInterface`, so we make sure #contact_message refers to a message entity.

ademarco (changes requested):
First review done, we need to add tests, we can try with Kernel tests. We would need to test that all the theming CSS classes and templates are applied correctly.

ademarco (changes requested):
I've found an error, reproducible as shown below:

Peek 2020-06-30 11-13

tests/Kernel/ContactFormRenderTest.php
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Tests that corporate contact form is rendered with the correct ECL markup.
+   */
+  public function testContactForm(): void {
+    $contact_form = ContactForm::create(['id' => 'oe_contact_form']);
+    $contact_form->setThirdPartySetting('oe_contact_forms', 'is_corporate_form', TRUE);
+    $header = 'this is a test header';
+    $contact_form->setThirdPartySetting('oe_contact_forms', 'header', $header);
```
ademarco: You use the $header variable only on this line, let's not set useless variables, can't we just set the value right away?

tests/Kernel/ContactFormRenderTest.php
```diff
@@ -0,0 +1,101 @@
...
+    $this->installConfig(['system', 'image', 'responsive_image', 'contact']);
+    $this->container->get('theme_installer')->install(['oe_theme']);
+    $this->container->get('theme_handler')->setDefault('oe_theme');
+    $this->container->set('theme.registry', NULL);
+  }
+
+  /**
+   * Tests that corporate contact form is rendered with the correct ECL markup.
+   */
+  public function testContactForm(): void {
```
ademarco, line 44: This does not test that, after submitting the form, the status message is properly themed as we want.

ademarco (changes requested):

tests/Kernel/ContactFormRenderTest.php
drishu: thats this part `$messenger->addMessage($full_view);`

ademarco: I mean the full thing: I get on the form, submit it and get that message properly styled. but OK if it's done elsewhere.

ademarco (dismissed):

ademarco (approved):

modules/oe_theme_contact_forms/README.md
22Alexandra: This permission is also needed "Use the site-wide contact form"

22Alexandra (approved):

nagyad (approved):

PR #511, opened 2020-06-30
OPENEUROPA-3387: Fix translated media rendering for paragraphs and Event.

22Alexandra:


imanoleguskiza (changes requested):
The fixes are good but we need to add tests on oe_theme. I sugges creating kernel tests for these.

22Alexandra:
> The fixes are good but we need to add tests on oe_theme. I sugges creating kernel tests for these.

I added kernel test for the paragraphs but as discussed it was tricky testing the media rendering in kernel for event and in behat the file was not uploaded.

oe_theme.theme
```diff
@@ -962,6 +962,9 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
   /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
   $paragraph = $variables['paragraph'];
 
+  // Retrieve the correct paragraph translation.
+  $paragraph = \Drupal::service('entity.repository')->getTranslationFromContext($paragraph, $paragraph->language()->getId());
```
upchuk: Why are you passing a language here? It defaults to the current content language but you are passing the original paragraph language so it will do...nothing?

oe_theme.theme
```diff
@@ -1403,6 +1410,8 @@ function oe_theme_preprocess_paragraph__oe_timeline(array &$variables): void {
 function oe_theme_preprocess_paragraph__oe_banner(array &$variables): void {
   /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
   $paragraph = $variables['paragraph'];
+  // Retrieve the correct paragraph translation.
+  $paragraph = \Drupal::service('entity.repository')->getTranslationFromContext($paragraph, $paragraph->language()->getId());
```
upchuk: Same here on the paragraph translation.

tests/Kernel/ContentEventRenderTest.php
```diff
@@ -0,0 +1,249 @@
...
+use Drupal\file\Entity\File;
+use Drupal\Tests\rdf_entity\Traits\RdfDatabaseConnectionTrait;
+use Drupal\user\Entity\Role;
+use Drupal\user\RoleInterface;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Tests that our Event content type render.
+ *
+ * @todo: Extend this test with more scenarios.
```
upchuk: What scenarios is it missing? This todo is quite meaningless.

tests/Kernel/ContentEventRenderTest.php
```diff
@@ -0,0 +1,249 @@
...
+ */
+class ContentEventRenderTest extends AbstractKernelTestBase {
```
upchuk: Cant we extend from `ContentRenderTest ` ?

tests/Kernel/Paragraphs/ContentRowTest.php
```diff
@@ -152,9 +151,6 @@ public function testRendering(): void {
    * Tests that side menu title translations are rendered correctly.
    */
   public function testSideMenuTranslation(): void {
-    // Create "French" language.
-    ConfigurableLanguage::createFromLangcode('fr')->save();
```
upchuk, line 156: Why did you remove? Where is this created?

upchuk (changes requested):

tests/Kernel/Paragraphs/ContentRowTest.php
22Alexandra: The languages are created in ParagraphsTestBase.

tests/Kernel/ContentEventRenderTest.php
nagyad: Because here we don't need news, pages, publication and policy content types to be installed, but many other for event content type related modules only. Therefore the parent::setup method can't be executed but the one in one more level down, the AbstractKernelTest.

oe_theme.theme
nagyad: You can see now the tests are failing. I noticed during debugging without that it get's reset to the English version so the English get's rendered instead of the Bulgarian.

tests/Kernel/ContentEventRenderTest.php
ademarco: Can't we turn this into a browser test? I've tried it locally and it sets up in 85 seconds. 

image

All you need to do is to replace the initial part of this test with the following:

```php
class ContentEventRenderTest extends BrowserTestBase {

  /**
   * {@inheritdoc}
   */
  public static $modules = [
    'config',
    'system',
    'oe_theme_helper',
    'path',
    'oe_theme_content_event',
    'content_translation',
  ];

  /**
   * {@inheritdoc}
   */
  protected function setUp() {
    parent::setUp();

    // Enable and set OpenEuropa Theme as default.
    $this->container->get('theme_installer')->install(['oe_theme']);

    // Make event node and image media translatable.
    $this->container->get('content_translation.manager')->setEnabled('node', 'oe_event', TRUE);
    $this->container->get('content_translation.manager')->setEnabled('media', 'image', TRUE);
    $this->container->get('router.builder')->rebuild();
  }
```

nagyad: Yes that was my initial try, but to debug this is a pure nightmare. Your example is missing couple things. The whole operation to have the event node created with files and translations, takes me 2.40mins to run. To make this test work I need to run it several times, I can't afford 20 minutes of waiting if I run this test 10 times. To debug it's double the time easily, and it's counter-productive. The only way I could imagine to go for such robust tests is with a local class that skips the setup after the first run and restores a database dump from the point when the setup finished and before anything ran. I tried, really and I lost hours...

nagyad: Also, for single scenario it could be okay, but as soon as you need to check 4-5 content types and different test scenarios the time quickly increases from 84 seconds to 500 seconds impacting the project test time on drone.

oe_theme.theme
```diff
@@ -975,6 +975,10 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
     // The media entity is not available anymore, bail out.
     return;
   }
+
+  // Retrieve the correct media translation.
+  $media = \Drupal::service('entity.repository')->getTranslationFromContext($media, $paragraph->language()->getId());
```
ademarco, line 980: We don't need this right, as it will get it from the context?
```
$paragraph->language()->getId()
```

oe_theme.theme
```diff
@@ -1452,6 +1456,9 @@ function oe_theme_preprocess_paragraph__oe_banner(array &$variables): void {
     return;
   }
 
+  // Retrieve the correct translation to display.
+  $media = \Drupal::service('entity.repository')->getTranslationFromContext($media, $paragraph->language()->getId());
```
ademarco, line 1460: Same as above.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -0,0 +1,137 @@
...
+      'uri' => 'public://example_1.jpeg',
+    ]);
+    $en_image->save();
+
+    // Create image media that we will use for the Bulgarian translation.
+    $this->container->get('file_system')->copy(drupal_get_path('theme', 'oe_theme') . '/tests/fixtures/placeholder.png', 'public://placeholder.png');
+    $bg_image = File::create([
+      'uri' => 'public://placeholder.png',
+    ]);
+    $bg_image->save();
```
ademarco: Let's use the same as we use below, for consistency and it's also more appropriate (with the _en,  _bg etc.):

```php
    $en_file = file_save_data(file_get_contents(drupal_get_path('theme', 'oe_theme') . '/tests/fixtures/example_1.jpeg'), 'public://example_1_en.jpeg');
    $en_file->setPermanent();
    $en_file->save();

    // Create Bulgarian file.
    $bg_file = file_save_data(file_get_contents(drupal_get_path('theme', 'oe_theme') . '/tests/fixtures/example_1.jpeg'), 'public://example_1_bg.jpeg');
    $bg_file->setPermanent();
    $bg_file->save();
```

tests/Functional/ContentEventRenderTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+    $file_urls = [
+      'en' => $en_image->createFileUrl(),
+      'bg' => $bg_image->createFileUrl(),
+    ];
+
+    foreach ($node->getTranslationLanguages() as $node_langcode => $node_language) {
+      $node = $this->container->get('entity.repository')->getTranslationFromContext($node, $node_langcode);
+      $this->drupalGet($node->toUrl());
+      $this->assertSession()->elementExists('css', 'img[src*="' . $file_urls[$node_langcode] . '"]');
```
ademarco: Let's restrict the scope of the selector in this test to the actual element where the image is supposed to be displayed, this is a bit too broad: `img[src*="' . $file_urls[$node_langcode] . '"]`

ademarco (changes requested):

oe_theme.theme
nagyad: If I remove these, tests are failing as it loads the English version of the file.

ademarco (approved):

imanoleguskiza (approved):

upchuk (approved):

PR #513, opened 2020-07-07
OPENEUROPA-3380: Project CT bottom region theming.

yenyasinn:
### Description

To style stakeholders use the Organisation teaser pattern to style the default view mode

### Change log

- Added:
Stakeholder group with Participants and Coordinators fields.

Related PR - https://github.com/openeuropa/oe_content/pull/224


templates/field/field--node--oe-project-participants--oe-project.html.twig
```diff
@@ -0,0 +1,11 @@
+{#
+/**
+ * @file
+ * Project participant field template.
```
enriquelacoma: The field name is "Participants | oe_project_participants", i would put "Project participants"

tests/Behat/ProjectContext.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+class ProjectContext extends RawDrupalContext {
+
+  /**
+   * Grant all necessary permissions so that anonymous users can see projects.
+   *
+   * @Given anonymous users can see projects
+   */
+  public function anonymousUsersCanSeeEvents(): void {
+    user_role_grant_permissions(Role::ANONYMOUS_ID, [
```
yenyasinn, line 21: @ademarco, I would prefer to add to the vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/DrupalContext.php statement "I am an anonymous user with the :permissions permission(s)". It could be useful for oe_theme/tests/Behat/EventContext.php also (and for future needs)

sass/components/_content_project.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * @file
+ * This file is used to styles the Project content page.
+ */
+#project-stakeholder {
+  .ecl-description-list__definition {
```
yenyasinn, line 8: It is a workaround of https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1279

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.install
```diff
@@ -0,0 +1,36 @@
...
+
+declare(strict_types = 1);
+
+use Drupal\Core\Config\FileStorage;
+use Drupal\Core\Entity\Entity\EntityViewDisplay;
+
+/**
+ * Implements hook_install().
+ */
+function oe_theme_content_entity_organisation_install() {
```
yenyasinn, line 15: Such code exists in 

- oe_theme_content_entity_venue_install()
- oe_theme_content_entity_contact_install()
- oe_theme_content_event_install()
- oe_theme_content_project_install()

Might be we should use one function from oe_theme (oe_theme_override_configuration(array $configurations)) to set configuration instead of duplicating code?

drishu: yes, thats a good idea, include the isConfigSyncing check, with parameters or FileStorage path and displays

sass/components/_content_project.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * @file
+ * This file is used to styles the Project content page.
```
enriquelacoma: I would put "This file is used to style the Project content page."

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.install
ademarco: We should do this indeed, feel free to propose a refactoring ticket about it.

yenyasinn: Ticket has been created - https://citnet.tech.ec.europa.eu/CITnet/jira/browse/D8TEWPPE-53.

tests/features/project.feature
```diff
@@ -1,24 +0,0 @@
...
-    Then I should see the text "Start date"
-    And I should see the text "01.07.2010"
-
-    # Change Project period end date.
-    When the Project Content "Project page" is updated as follows:
-      | Project period start date | 2010-07-01 |
-      | Project period end date   | 2012-01-01 |
-    And I reload the page
-    Then I should see the text "Project duration"
-    And I should see the text "01.07.2010 - 01.01.2012"
```
ademarco, line 24: Was this second part ported to the functional test?

templates/oe_organisation/oe-organisation.html.twig
```diff
@@ -0,0 +1,37 @@
+{#
+/**
+ * @file
+ * Default theme implementation to present a corporate content entity.
```
drishu: an organisation entity.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -18,6 +21,13 @@ class ContentProjectRenderTest extends BrowserTestBase {
    */
   protected $nodeStorage;
 
+  /**
+   * The organisation storage.
+   *
+   * @var \Drupal\oe_content_entity_organisation\Entity\OrganisationInterface
+   */
+  protected $organisationStorage;
```
drishu: I see this called only once, do we need it as a class property ? same question for nodeStorage

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +210,114 @@ public function testProjectTopGroupRender(): void {
...
+    // Assert Participants field.
+    $project_stakeholder = $this->assertSession()->elementExists('css', 'div#project-stakeholder');
+    $stakeholder_sub_headers = $project_stakeholder->findAll('css', 'h3');
+    $this->assertCount(1, $stakeholder_sub_headers);
+    $this->assertEqual($stakeholder_sub_headers[0]->getText(), 'Participants');
+    $this->assertOrganisation($project_stakeholder, 'participant');
+  }
+
+  /**
+   * Creates organisation entity.
```
drishu: parameter and return please

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +210,114 @@ public function testProjectTopGroupRender(): void {
...
+
+  /**
+   * Tests Organisation entity.
+   *
+   * @param \Behat\Mink\Element\NodeElement $group
+   *   Stakeholder group.
+   * @param string $name
+   *   Name of the entity.
+   */
+  protected function assertOrganisation(NodeElement $group, $name) {
```
drishu: :void

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.install
```diff
@@ -0,0 +1,36 @@
...
+    return;
+  }
+
+  // Override entity view displays.
+  $storage = new FileStorage(drupal_get_path('module', 'oe_theme_content_entity_organisation') . '/config/overrides');
+  $displays = [
+    'core.entity_view_display.oe_organisation.oe_stakeholder.default',
+  ];
+
+  foreach ($displays as $display) {
```
nagyad: The foreach is not required for one item.

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.install
```diff
@@ -0,0 +1,36 @@
...
+    $values = $storage->read($display);
+    $config = EntityViewDisplay::load($values['id']);
+    foreach ($values as $key => $value) {
+      $config->set($key, $value);
+    }
+    $config->save();
```
nagyad: Instead of this let's use the storage itself to update records such as:
`$view_display_values = $storage->read($view_display_id);`
`$view_display = EntityViewDisplay::load($view_display_values['id']);`
`$updated_view_display = $entity_storage->updateFromStorageRecord($view_display, $view_display_values);`
`$updated_view_display->save();`

nagyad: Make sure that the variable names are truly reflect their content, such as the $display is in your code is a $display_id to be more precise.

oe_theme.theme
```diff
@@ -1529,3 +1530,62 @@ function oe_theme_preprocess_page(array &$variables): void {
...
+      $variables[$pattern_field] = $entity->get($entity_field)->value;
+    }
+  }
+
+  // Extract the image if present.
+  if (!$entity->get('oe_logo')->isEmpty()) {
+    /** @var \Drupal\file\Plugin\Field\FieldType\FileItem $image_item */
+    $media_image = $entity->get('oe_logo')->entity;
+
+    if (!empty($media_image) && !$media_image->get('oe_media_image')->isEmpty()) {
```
nagyad: Why do you check here if image field is not empty? It is a required field in the image media bundle.

tests/Functional/ContentProjectRenderTest.php
nagyad: I agree, let's use a local variable.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +210,114 @@ public function testProjectTopGroupRender(): void {
...
+    $stakeholder_sub_headers = $project_stakeholder->findAll('css', 'h3');
+    $this->assertCount(1, $stakeholder_sub_headers);
+    $this->assertEqual($stakeholder_sub_headers[0]->getText(), 'Participants');
+    $this->assertOrganisation($project_stakeholder, 'participant');
+  }
+
+  /**
+   * Creates organisation entity.
+   */
+  protected function getOrganisationEntity($name, $status) {
```
nagyad: Here as well, missing parameter and return types.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +210,114 @@ public function testProjectTopGroupRender(): void {
...
+      'name' => "Test image $name",
+      'oe_media_image' => [
+        'target_id' => (int) $file->id(),
+      ],
+      'uid' => 0,
+      'status' => 1,
+    ]);
+    $media->save();
+
+    $organisation = $this->organisationStorage->create([
```
nagyad: Use here the storage the same way as for media, and remove the class property.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +210,114 @@ public function testProjectTopGroupRender(): void {
...
+      'oe_website' => ['uri' => "http://www.example.com/website_$name"],
+      'status' => $status,
+    ]);
+    $organisation->save();
+
+    return $organisation;
+  }
+
+  /**
+   * Tests Organisation entity.
```
nagyad: Asserts Organisation entity.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +210,114 @@ public function testProjectTopGroupRender(): void {
...
+  /**
+   * Creates organisation entity.
+   */
+  protected function getOrganisationEntity($name, $status) {
+    // Create image for logo.
+    $file = file_save_data(file_get_contents(drupal_get_path('theme', 'oe_theme') . '/tests/fixtures/placeholder.png'), "public://placeholder_$name.png");
+    $file->setPermanent();
+    $file->save();
+
+    $media = $this->container->get('entity_type.manager')->getStorage('media')->create([
```
nagyad: Media storage could be a good candidate for class property.

nagyad (changes requested):

tests/features/project.feature
yenyasinn: Fixed

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.install
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Implements hook_install().
+ */
+function oe_theme_content_entity_organisation_install() {
+  // If we are installing from config, we bail out.
+  if (\Drupal::isConfigSyncing() === TRUE) {
+    return;
+  }
+
+  // Override entity view displays.
```
yenyasinn: @nagyad I've provided a little bit another solution than you asked. This code can work with any types of configurations, not only with entity view display configurations.

oe_theme.theme
yenyasinn: It is safer. Media entity can be removed. In this case page will be unavailable due to thrown exception.

tests/Functional/ContentProjectRenderTest.php
yenyasinn: Done

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.install
```diff
@@ -0,0 +1,40 @@
...
```
nagyad: You are overriding one entity view display. Should not be plural.

oe_theme.theme
nagyad: "Safer"? For me you sound like you are uncertain of the outcome. Assumption and uncertainty is not a good reason for doing extra checks. You are checking the media entity not to be empty and that is one thing, but my remark was regarding the media entity's image field that is required and it will be filled in so most likely won't be empty, instead when you retrieve the image from the field, assert for more concrete like `instanceof ImageItem`.
Another thing, above you you already check if the 'oe_logo' field isEmpty(), and after you assign the entity to a variable then you check if that is empty. Instead of this you would be better off to check if `instanceof MediaInterface`.
Like this:
`$media_image = $entity->get('oe_logo')->entity;`
`  if (!$media_image instanceof MediaInterface) {`
   ` return;`
 ` }`
Try to avoid wrapping logic into nested ifs.

sass/components/_content_project.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * @file
+ * This file is used to style the Project content page.
+ */
+#project-stakeholder {
+  .ecl-description-list__definition {
+    width: 259px;
```
nagyad, line 9: Is this responsive?

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +205,122 @@ public function testProjectTopGroupRender(): void {
...
+   *
+   * @var string $name
+   *   Name of the entity. Is used as a parameter for test data.
+   * @var int $status
+   *   Entity status. 1 - published, 0 - unpublished.
+   *
+   * @return \Drupal\oe_content_entity_organisation\Entity\OrganisationInterface
+   *   Organisation entity.
+   */
+  protected function getOrganisationEntity($name, $status): OrganisationInterface {
```
nagyad: This can't be called as get. Get methods are getting retrieving something from somewhere. Here you clearly creating something, so please name method as createStakeholderOrganisationEntity().
Also I miss type hint of the arguments.

nagyad: As a plus, it's not mandatory but since we have a method for this and to make the test nicely readable without looking into this function, I would use `published` and `unpublished` as arguments and map them to the correct integer value within this method. With the correct method name this would leave no doubt behind when reading the test and what are we doing.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +205,122 @@ public function testProjectTopGroupRender(): void {
...
+
+  /**
+   * Asserts Organisation entity.
+   *
+   * @param \Behat\Mink\Element\NodeElement $group
+   *   Stakeholder group.
+   * @param string $name
+   *   Name of the entity.
+   */
+  protected function assertOrganisation(NodeElement $group, $name): void {
```
nagyad: I find this very confusing to call the variable group and in the annotation call it Stakeholder group and the method organisation. Let's call this as assertStakeholderOrganisationRendering() or assertStakeholderRendering() and rename $group to $rendered_stakeholder_element or something.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +205,122 @@ public function testProjectTopGroupRender(): void {
...
+      'oe_website' => ['uri' => "http://www.example.com/website_$name"],
+      'status' => $status,
+    ]);
+    $organisation->save();
+
+    return $organisation;
+  }
+
+  /**
+   * Asserts Organisation entity.
```
nagyad: You are asserting the rendering of a Stakeholder.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +205,122 @@ public function testProjectTopGroupRender(): void {
...
+    // Assert Participants field.
+    $project_stakeholder = $this->assertSession()->elementExists('css', 'div#project-stakeholder');
+    $stakeholder_sub_headers = $project_stakeholder->findAll('css', 'h3');
+    $this->assertCount(1, $stakeholder_sub_headers);
+    $this->assertEqual($stakeholder_sub_headers[0]->getText(), 'Participants');
+    $this->assertOrganisation($project_stakeholder, 'participant');
+  }
+
+  /**
+   * Creates organisation entity.
```
nagyad: Creates a stakeholder organisation entity.

nagyad (changes requested):
In general my advice is to please put some consistency into naming things, because you are calling participants and coordinators as organisation or stakeholder randomly. This is really confusing me.

tests/Functional/ContentProjectRenderTest.php
yenyasinn: @nagyad , let's extend OrganisationInterface and add there:
```
interface OrganisationInterface extends CorporateEntityInterface {

  /**
   * Denotes that the organisation is not published.
   */
  const NOT_PUBLISHED = 0;

  /**
   * Denotes that the organisation is published.
   */
  const PUBLISHED = 1;
}
```
then we can invoke createStakeholderOrganisationEntity('name', OrganisationInterface::PUBLISHED). Is it appropriate solution?

nagyad: No. Why would we do that? That would be wrong in this case, we are in a scope of a test class and a custom method to create something very specific, that should not affect the implementation of a component. Simply map the string with the value in the method.

yenyasinn: Using constants to determine status of the entity is a pattern that is used in NodeInterface or CommentInterface. Such approach
* defines status clearly
* increases readability of the code
* can be reused in different places.

From my point of view mapping for the status increases function complexity and can be useful just in one particular case.

sass/components/_content_project.scss
yenyasinn: Yes, it is responsive.

Screenshot 2020-07-16 at 16 36 02

tests/Functional/ContentProjectRenderTest.php
nagyad: The things you wrote are true, we could consider to have this in the CorporateEntityInterface rather than on organisation as this is true for all corporate entities. But I would have done it as soon as it supports a feature rather than just a test case.

ademarco: We can put them on `CorporateEntityInterface`.

oe_theme.theme
```diff
@@ -1529,3 +1531,60 @@ function oe_theme_preprocess_page(array &$variables): void {
 function oe_theme_preprocess_pattern_organisation_teaser__preview(array &$variables): void {
   $variables['logo'] = ImageValueObject::fromArray($variables['logo']);
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_oe_organisation(&$variables): void {
```
nagyad: Please move this hook into the organisation submodule.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -4,19 +4,23 @@
...
+use Drupal\oe_content_entity_organisation\Entity\OrganisationInterface;
 
 /**
  * Tests that our Project content type renders correctly.
  */
 class ContentProjectRenderTest extends BrowserTestBase {
 
   /**
-   * The node storage.
+   * The organisation storage.
```
ademarco:
```
 * The media storage.
```

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -69,9 +78,14 @@ public function testProjectTopGroupRender(): void {
...
 
+    // Create organisations for Coordinators and Participants fields.
+    // Unpublished entity should not be shown.
+    $coordinator_organisation = $this->createStakeholderOrganisationEntity('coordinator', 1);
+    $participant_organisation = $this->createStakeholderOrganisationEntity('participant', 0);
+
     // Create a Project node.
     /** @var \Drupal\node\Entity\Node $node */
-    $node = $this->nodeStorage->create([
+    $node = $this->container->get('entity_type.manager')->getStorage('node')->create([
```
ademarco: Let's bring the node storage class property back, since we are doing it for media.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +205,122 @@ public function testProjectTopGroupRender(): void {
...
+      'name' => "Test image $name",
+      'oe_media_image' => [
+        'target_id' => (int) $file->id(),
+      ],
+      'uid' => 0,
+      'status' => 1,
+    ]);
+    $media->save();
+
+    $organisation = $this->container->get('entity_type.manager')->getStorage('oe_organisation')->create([
```
ademarco: Same here, since we are doing it for media, let's also have the organisation storage as a class property.

ademarco (changes requested):
Let's also remove the links from here, as they don't link to anything relevant for the user. I know it's out of scope, but this is already merged, so let's fix it here.

image

sass/components/_content_project.scss
ademarco: Let's put a comment here to remind ourselves to remove this when that issue is fixed:
```
// Remove this when FRONT-1279 is fixed.
```
Also please add the same comment on _description-list.scss, I forgot about it.

tests/Functional/ContentProjectRenderTest.php
yenyasinn: As it was discussed organisationStorage is a class property.

yenyasinn:
> Let's also remove the links from here, as they don't link to anything relevant for the user. I know it's out of scope, but this is already merged, so let's fix it here.
> 
> image

Fixed.

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
```diff
@@ -0,0 +1,72 @@
...
+    $file = $image_item->get('entity')->getValue();
+
+    if ($file instanceof FileInterface) {
+      $variables['logo'] = ImageValueObject::fromImageItem($image_item);
+      $cacheability->addCacheableDependency($file);
+    }
+    $cacheability->applyTo($variables);
+  }
+
+  // Set contact page url and icon.
```
imanoleguskiza: There is no icon to be set (maybe you are referring to the logo that is set before?

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
```diff
@@ -0,0 +1,72 @@
...
+use Drupal\file\FileInterface;
+use Drupal\media\MediaInterface;
+use Drupal\oe_theme\ValueObject\ImageValueObject;
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_content_entity_organisation_preprocess_oe_organisation(&$variables): void {
+  $entity = $variables['entity'];
+  // Prepare fields.
```
imanoleguskiza: This comment doesn't really say anything, either rewrite it or remove it.

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
```diff
@@ -0,0 +1,72 @@
...
+    $link = $entity->get('oe_contact_url')->first()->getValue();
+    if (UrlHelper::isExternal($link['uri'])) {
+      $variables['contact_page_url'] = $link['uri'];
+    }
+    else {
+      $variables['contact_page_url'] = Url::fromUri($link['uri'])->toString();
+    }
+  }
+
+  // Process contact item.
```
imanoleguskiza: Maybe better "Process organisation contact details" or something like that.

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
```diff
@@ -0,0 +1,72 @@
...
+  $variables['organisation_details'] = [];
+  $contact_fields = ['oe_address', 'oe_website'];
+  foreach ($contact_fields as $field) {
+    if (empty($variables['content'][$field]['#title'])) {
```
imanoleguskiza: How come you don't use the field api for this? $entity->get($field)->isEmpty()

tests/Functional/ContentProjectRenderTest.php
imanoleguskiza: This comment has not been resolved.

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
```diff
@@ -0,0 +1,72 @@
...
+  // Prepare fields.
+  $fields_values = ['name' => 'name', 'oe_acronym' => 'acronym'];
+  foreach ($fields_values as $entity_field => $pattern_field) {
+    if (!$entity->get($entity_field)->isEmpty()) {
+      $variables[$pattern_field] = $entity->get($entity_field)->value;
+    }
+  }
+
+  // Extract the image if Media and File entities exist.
+  $media_image = $entity->get('oe_logo')->entity;
```
imanoleguskiza, line 28: We need to make sure that the entity is in the correct language here. Use \Drupal::service('entity.repository')->getTranslationFromContext().

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +223,122 @@ public function testProjectTopGroupRender(): void {
...
+    $this->assertEquals('Website', $labels[1]->getText());
+    $values = $description_lists[0]->findAll('css', 'dd.ecl-description-list__definition');
+    $this->assertCount(2, $values);
+    $this->assertEquals("Address $name, 1001 Brussels, Belgium", $values[0]->getText());
+    $values[1]->hasLink("http://www.example.com/website_$name");
+
+    // Assert contact link.
+    $contact_links = $rendered_stakeholder_element->findAll('css', '.ecl-u-mt-l.ecl-u-type-bold');
+    $this->assertCount(1, $contact_links);
+    $contact_links[0]->hasLink('Contact organisation');
```
imanoleguskiza: Maybe we should also assert the target of the link.

imanoleguskiza (changes requested):

tests/Functional/ContentProjectRenderTest.php
yenyasinn: Antonio asked to return it back.

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
yenyasinn: Fixed

templates/content/field-group-html-element--node--oe-project--group-stakeholder.html.twig
```diff
@@ -0,0 +1,19 @@
+{#
+/**
+ * @file
+ * Stakeholder field group theme implementation.
+ *
+ * @see ./modules/contrib/field_group/templates/field-group-html-element.html.twig
+ */
+#}
+
+<div id="project-stakeholder">
```
ademarco: Let's rename this into `project-stakeholders`, plural.

modules/oe_theme_content_project/config/overrides/core.entity_view_display.node.oe_project.default.yml
```diff
@@ -102,6 +103,25 @@ third_party_settings:
         id: ''
         classes: ''
       label: Results
+    group_stakeholder:
```
ademarco: Let's rename this into `group_stakeholders`, plural.

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
```diff
@@ -0,0 +1,72 @@
...
+  $variables['contact_page_url'] = '';
+  if (!$entity->get('oe_contact_url')->isEmpty()) {
+    $link = $entity->get('oe_contact_url')->first()->getValue();
+    if (UrlHelper::isExternal($link['uri'])) {
+      $variables['contact_page_url'] = $link['uri'];
+    }
+    else {
+      $variables['contact_page_url'] = Url::fromUri($link['uri'])->toString();
+    }
+  }
```
ademarco, line 49: Let's replace the above with this, as the field object can already access the URL object and its toString method:

```php
  if (!$entity->get('oe_contact_url')->isEmpty()) {
    /** @var \Drupal\link\Plugin\Field\FieldType\LinkItem $link */
    $link = $entity->get('oe_contact_url')->first();
    $variables['contact_page_url'] = $link->getUrl()->toString();
  }
```

ademarco, line 49: Let's replace the above with this, as the field object can already access the URL object and its toString method:

```php
  if (!$entity->get('oe_contact_url')->isEmpty()) {
    /** @var \Drupal\link\Plugin\Field\FieldType\LinkItem $link */
    $link = $entity->get('oe_contact_url')->first();
    $variables['contact_page_url'] = $link->getUrl()->toString();
  }
```

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +199,138 @@ public function testProjectTopGroupRender(): void {
...
+   *
+   * @var string $name
+   *   Name of the entity. Is used as a parameter for test data.
+   * @var int $status
+   *   Entity status. 1 - published, 0 - unpublished.
+   *
+   * @return \Drupal\oe_content_entity_organisation\Entity\OrganisationInterface
+   *   Organisation entity.
+   */
+  protected function createStakeholderOrganisationEntity($name, $status): OrganisationInterface {
```
ademarco: Type hinting is missing:
```php
   protected function createStakeholderOrganisationEntity(string $name, int $status): OrganisationInterface {
```

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +199,138 @@ public function testProjectTopGroupRender(): void {
...
+
+  /**
+   * Asserts rendering of a Stakeholder.
+   *
+   * @param \Behat\Mink\Element\NodeElement $rendered_stakeholder_element
+   *   Stakeholder group.
+   * @param string $name
+   *   Name of the entity.
+   */
+  protected function assertStakeholderOrganisationRendering(NodeElement $rendered_stakeholder_element, $name): void {
```
ademarco: Type hinting missing:
```php
   protected function assertStakeholderOrganisationRendering(NodeElement $rendered_stakeholder_element, string $name): void {
```

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -177,6 +199,138 @@ public function testProjectTopGroupRender(): void {
...
+  /**
+   * Gets the entity type's storage.
+   *
+   * @param string $entity_type_id
+   *   The entity type ID to get a storage for.
+   *
+   * @return \Drupal\Core\Entity\EntityStorageInterface
+   *   The entity type's storage.
+   */
+  protected function getStorage($entity_type_id): EntityStorageInterface {
```
ademarco: Type hinting missing:
```php
   protected function getStorage(string $entity_type_id): EntityStorageInterface {
```

ademarco (changes requested):

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
yenyasinn: Fixed

ademarco (dismissed):

modules/oe_theme_content_entity_organisation/oe_theme_content_entity_organisation.module
```diff
@@ -0,0 +1,66 @@
...
+      continue;
+    }
+    $variables['organisation_details'][] = [
+      'label' => $variables['content'][$field]['#title'],
+      'body' => [
+        '#label_display' => 'hidden',
+      ] + $variables['content'][$field],
+    ];
+  }
+
```
nagyad: Empty line before the end of the function, please remove.

nagyad (changes requested):

imanoleguskiza (approved):

nagyad (approved):

PR #516, opened 2020-07-15
EWPP-30: [oe_theme] Add field list to list item pattern

enriquelacoma:
### Description

Add field list on the default, thumbnail-primary and thumbnail-secondary list item variants

### Change log

- Added:
- Changed: list_item patterns
- Removed: compositions/ec-component-content-item/content-item-date.html.twig and compositions/ec-component-content-item/content-item.html.twig

### Commands

```sh
[Insert commands here]

```


tests/Kernel/fixtures/rendering.yml
```diff
@@ -905,6 +943,7 @@
       title: "Item title"
       url: "http://example.com"
       detail: "EU economy, the euro, and practical information for EU businesses and entrepreneurs."
+      lenght: 85
```
ademarco: Why are we adding all these `length` values to these tests? Existing tests shouldn't be changed, or else we risk changing behavior that we want to keep. In this case if length is not set then no trim should happen, do we still ensure that?

tests/Kernel/fixtures/rendering.yml
```diff
@@ -775,28 +798,41 @@
...
+      'article>div .ecl-description-list .ecl-description-list__term': 2
+      'article>div .ecl-description-list .ecl-description-list__definition': 2
     equals:
       'article.ecl-content-item div.ecl-content-item__meta': "Meta 1 | Meta 2"
       'article.ecl-content-item div.ecl-content-item__title .ecl-link': "Item title"
       'article.ecl-content-item div.ecl-content-item__description': "Item text"
+      'article>div .ecl-description-list .ecl-description-list__term:nth-child(1)': "Countries"
+      'article>div .ecl-description-list .ecl-description-list__definition:nth-child(2)': "Spain, France, Italy"
+      'article>div .ecl-description-list .ecl-description-list__term:nth-child(3)': "Deadline dates"
+      'article>div .ecl-description-list .ecl-description-list__definition:nth-child(4)': "1 April 2017 / 1 July 2017"
```
ademarco: Let's put spaces around `>` so it's easier to read, like this:
```
       'article > div .ecl-description-list .ecl-description-list__definition:nth-child(4)': "1 April 2017 / 1 July 2017"
```

tests/Kernel/fixtures/rendering.yml
```diff
@@ -642,15 +643,14 @@
...
-        - "Meta 2"
   assertions:
     count:
       'article.ecl-content-item div.ecl-content-item__title a.ecl-link[href="http://example.com"]': 1
       'article.ecl-content-item div[role="img"].ecl-u-d-lg-block': 0
       'article.ecl-content-item div.ecl-content-item__image__before': 0
       'article.ecl-content-item div.ecl-content-item__image__after': 0
     equals:
-      'article.ecl-content-item div.ecl-content-item__meta': "Meta 1 | Meta 2"
+      'article.ecl-content-item div.ecl-content-item__meta': "Meta 1"
```
ademarco, line 652: Why are we removing one of the meta here?

tests/Kernel/fixtures/rendering.yml
```diff
@@ -642,15 +643,14 @@
       length: 6
       meta:
         - "Meta 1"
-        - "Meta 2"
   assertions:
     count:
       'article.ecl-content-item div.ecl-content-item__title a.ecl-link[href="http://example.com"]': 1
       'article.ecl-content-item div[role="img"].ecl-u-d-lg-block': 0
       'article.ecl-content-item div.ecl-content-item__image__before': 0
       'article.ecl-content-item div.ecl-content-item__image__after': 0
```
ademarco, line 650: Also let's add a test that proves that, when no field list field is passed, then we don't print any field list.

templates/patterns/list_item/list_item.ui_patterns.yml
```diff
@@ -76,3 +76,12 @@ list_item:
       preview:
         src: "https://placeimg.com/1200/900/arch"
         alt: "Alternative text for image"
+    field_list:
+        type: "array"
+        label: "Items"
+        description: "Label of the field"
```
ademarco: Label and description are not correct, let's use:
```
        label: "List of fields"
        description: "List of fields to be displayed under the item description. Only used on XXX"
```
The use the variants where this is used instead of `XXX`

ademarco (changes requested):

tests/Kernel/fixtures/rendering.yml
enriquelacoma: To test "|" is not there when using only one meta

tests/Kernel/fixtures/rendering.yml
enriquelacoma: They were not needed, i removed them

tests/Kernel/fixtures/rendering.yml
enriquelacoma: Updated

tests/Kernel/fixtures/rendering.yml
enriquelacoma: I added a test checking 
   'article > div .ecl-description-list': 0

templates/patterns/list_item/list_item.ui_patterns.yml
enriquelacoma: Updated

templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
```diff
@@ -4,10 +4,8 @@
...
-{% include '@oe_theme/compositions/ec-component-content-item/content-item.html.twig' with {
-  title: {
-    label: title,
-    path: url
-  },
-  description: detail|smart_trim(length)
+{% include '@oe_theme/patterns/list_item/pattern-list-item.html.twig' with {
+  title: title,
+  url: url,
+  detail: detail|smart_trim(length),
```
imanoleguskiza: We apply the same filter in the base pattern so I think we can skip it here.

templates/patterns/list_item/pattern-list-item--variant-thumbnail-primary.html.twig
```diff
@@ -4,13 +4,12 @@
...
-{% include '@oe_theme/compositions/ec-component-content-item/content-item.html.twig' with {
-  title: {
-    label: title,
-    path: url
-  },
-  description: detail|smart_trim(length),
+{% include '@oe_theme/patterns/list_item/pattern-list-item.html.twig' with {
+  title: title,
+  url: url,
+  detail: detail|smart_trim(length),
```
imanoleguskiza: Same as above.

templates/patterns/list_item/pattern-list-item--variant-thumbnail-secondary.html.twig
```diff
@@ -4,13 +4,12 @@
...
-{% include '@oe_theme/compositions/ec-component-content-item/content-item.html.twig' with {
-  title: {
-    label: title,
-    path: url
-  },
-  description: detail|smart_trim(length),
+{% include '@oe_theme/patterns/list_item/pattern-list-item.html.twig' with {
+  title: title,
+  url: url,
+  detail: detail|smart_trim(length),
```
imanoleguskiza: Same as above.

imanoleguskiza (changes requested):

templates/patterns/list_item/pattern-list-item--variant-navigation.html.twig
enriquelacoma: Removed

templates/patterns/list_item/pattern-list-item--variant-thumbnail-primary.html.twig
enriquelacoma: Removed

templates/patterns/list_item/pattern-list-item--variant-thumbnail-secondary.html.twig
enriquelacoma: Removed

imanoleguskiza (approved):

ademarco (approved):

PR #520, opened 2020-07-22
EWPP-55: [oe_theme] Create in-page navigation field group formatter

enriquelacoma:
### Description

Create group for in-page navigation.

### Change log

- Added: FieldInPageNavigationFormatter.php

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_theme_helper/src/Element/InLinePaginationElement.php
+++ b/modules/oe_theme_helper/src/Element/InLinePaginationElement.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+
+/**
+ * Provides a render element for a html element.
+ *
+ * @FormElement("field_group_in_page_navigation")
+ */
+class InLinePaginationElement extends RenderElement {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getInfo() {
+    $class = get_class($this);
+
+    return [
+      '#process' => [
+        [$class, 'processGroup'],
+        [$class, 'processHtmlElement'],
+      ],
+      '#pre_render' => [
+        [$class, 'preRenderGroup'],
+      ],
+      '#theme_wrappers' => ['field_group_in_page_navigation'],
+    ];
+  }
+
+}
--- a/modules/oe_theme_helper/src/Element/InPageNavigationElement.php
+++ b/modules/oe_theme_helper/src/Element/InPageNavigationElement.php
@@ -7,11 +7,11 @@ namespace Drupal\oe_theme_helper\Element;
 use Drupal\Core\Render\Element\RenderElement;
 
 /**
- * Provides a render element for a html element.
+ * Provides a render element for an html element.
  *
  * @FormElement("field_group_in_page_navigation")
  */
-class InLinePaginationElement extends RenderElement {
+class InPageNavigationElement extends RenderElement {
 
   /**
    * {@inheritdoc}
@@ -22,7 +22,6 @@ class InLinePaginationElement extends RenderElement {
     return [
       '#process' => [
         [$class, 'processGroup'],
-        [$class, 'processHtmlElement'],
       ],
       '#pre_render' => [
         [$class, 'preRenderGroup'],
--- a/modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
+++ b/modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
@@ -1,0 +1,140 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Plugin\field_group\FieldGroupFormatter;
+
+use Drupal\field_group\FieldGroupFormatterBase;
+
+/**
+ * Format a field group using the field list pattern.
+ *
+ * @FieldGroupFormatter(
+ *   id = "oe_theme_helper_in_page_navigation",
+ *   label = @Translation("Field In-page navigation"),
+ *   description = @Translation("Format field group for ecl-inpage-navigation."),
+ *   supported_contexts = {
+ *     "view"
+ *   }
+ * )
+ */
+class FieldInPageNavigationFormatter extends FieldGroupFormatterBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function process(&$element, $processed_object) {
+    $element += [
+      '#type' => 'field_group_in_page_navigation',
+    ];
+    // Get children elements in in page navigation group.
+    $field_elements = $processed_object['#fieldgroups'][$element['#group_name']]->children;
+    // Add anchors links and anchors to children with labels.
+    $links = [];
+    $i = 1;
+    foreach ($field_elements as $key => $value) {
+      // Check if we should add the field to page contents.
+      if (!is_array($element[$value]) || $this->excludeField($value, $element[$value], $processed_object['#fieldgroups']) === TRUE) {
+        continue;
+      }
+      // We will add anchors when the element, field or group, has a label.
+      $label = $this->fieldGetLabel($value, $processed_object);
+      // Add menu element and anchor.
+      if ($label !== '') {
+        // If the group has a default id, use it as anchor link.
+        if (isset($processed_object['#fieldgroups'][$value]) && isset($processed_object['#fieldgroups'][$value]->format_settings['id']) && $processed_object['#fieldgroups'][$value]->format_settings['id'] !== '') {
+          $id = $processed_object['#fieldgroups'][$value]->format_settings['id'];
+        }
+        else {
+          $id = "inline-nav-" . $i++;
+        }
+        $links[] = [
+          'href' => "#" . $id,
+          'label' => $label,
+        ];
+        $element[$value]['#id'] = $id;
+        $element[$value]['#attributes']['id'] = $id;
+      }
+    }
+    $element['#links'] = $links;
+    if (count($links) > 0) {
+      $element['#inline_title'] = t('Page contents');
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function preRender(&$element, $rendering_object) {
+    parent::preRender($element, $rendering_object);
+    $this->process($element, $rendering_object);
+  }
+
+  /**
+   * Check if the field should create an in page navigation entry.
+   *
+   * @param string $field_name
+   *   Name of the field.
+   * @param array $field
+   *   The field to check.
+   * @param array $fieldgroups
+   *   List of groups.
+   *
+   * @return bool
+   *   Include or not the field.
+   */
+  private function excludeField(string $field_name, array $field, array $fieldgroups) :bool {
+    // Exclude empty fields.
+    if (!in_array($field_name, array_keys($fieldgroups)) && !isset($field['#items'])) {
+      return TRUE;
+    }
+    // Exclude empty groups.
+    if (in_array($field_name, array_keys($fieldgroups)) && count($field) < 1) {
+      return TRUE;
+    }
+    // Exclude groups with only empty fields.
+    if (in_array($field_name, array_keys($fieldgroups))) {
+      foreach ($field as $value) {
+        // If there is a field with values, include the group.
+        if (isset($value['#items'])) {
+          return FALSE;
+        }
+      }
+      return TRUE;
+    }
+    return FALSE;
+  }
+
+  /**
+   * Get the label used on the field.
+   *
+   * @param string $field_name
+   *   Name of the field.
+   * @param array $processed_object
+   *   Entity processed.
+   *
+   * @return string
+   *   Label of the field.
+   */
+  private function fieldGetLabel(string $field_name, array $processed_object): string {
+    $label = '';
+    if (isset($processed_object['#fieldgroups'][$field_name]->label)) {
+      if ($processed_object['#fieldgroups'][$field_name]->label !== '') {
+        $label = $processed_object['#fieldgroups'][$field_name]->label;
+        // We will assume groups labels are shown unless it is empty or
+        // explicitly hidden.
+        if (isset($processed_object['#fieldgroups'][$field_name]->format_settings['show_label']) && $processed_object['#fieldgroups'][$field_name]->format_settings['show_label'] === FALSE) {
+          return '';
+        }
+      }
+    }
+    else {
+      // For fields, check if the label is not hidden.
+      if (!in_array($processed_object[$field_name]['#label_display'], ['hidden', 'visually_hidden'])) {
+        return $processed_object[$field_name]['#title'];
+      }
+    }
+    return $label;
+  }
+
+}
--- a/modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
+++ b/modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
@@ -5,14 +5,16 @@ declare(strict_types = 1);
 namespace Drupal\oe_theme_helper\Plugin\field_group\FieldGroupFormatter;
 
 use Drupal\field_group\FieldGroupFormatterBase;
+use Drupal\Core\Render\Element;
+use Drupal\Component\Utility\Html;
 
 /**
- * Format a field group using the field list pattern.
+ * Display a field group using the ECL in-page navigation component.
  *
  * @FieldGroupFormatter(
  *   id = "oe_theme_helper_in_page_navigation",
- *   label = @Translation("Field In-page navigation"),
- *   description = @Translation("Format field group for ecl-inpage-navigation."),
+ *   label = @Translation("In-page navigation"),
+ *   description = @Translation("Display a field group using the ECL in-page navigation component."),
  *   supported_contexts = {
  *     "view"
  *   }
@@ -27,36 +29,40 @@ class FieldInPageNavigationFormatter extends FieldGroupFormatterBase {
     $element += [
       '#type' => 'field_group_in_page_navigation',
     ];
-    // Get children elements in in page navigation group.
-    $field_elements = $processed_object['#fieldgroups'][$element['#group_name']]->children;
-    // Add anchors links and anchors to children with labels.
+
+    // Get children elements from the in-page navigation group.
     $links = [];
-    $i = 1;
-    foreach ($field_elements as $key => $value) {
-      // Check if we should add the field to page contents.
-      if (!is_array($element[$value]) || $this->excludeField($value, $element[$value], $processed_object['#fieldgroups']) === TRUE) {
+    foreach (Element::children($element) as $value) {
+      $type = $this->groupOrField($value, $processed_object['#fieldgroups']);
+      if (!$this->hasVisibleLabel($type, $value, $processed_object)) {
         continue;
       }
-      // We will add anchors when the element, field or group, has a label.
-      $label = $this->fieldGetLabel($value, $processed_object);
-      // Add menu element and anchor.
-      if ($label !== '') {
-        // If the group has a default id, use it as anchor link.
-        if (isset($processed_object['#fieldgroups'][$value]) && isset($processed_object['#fieldgroups'][$value]->format_settings['id']) && $processed_object['#fieldgroups'][$value]->format_settings['id'] !== '') {
-          $id = $processed_object['#fieldgroups'][$value]->format_settings['id'];
-        }
-        else {
-          $id = "inline-nav-" . $i++;
-        }
-        $links[] = [
-          'href' => "#" . $id,
-          'label' => $label,
-        ];
-        $element[$value]['#id'] = $id;
-        $element[$value]['#attributes']['id'] = $id;
+      if ($type == 'field') {
+        $label = $processed_object[$value]['#title'];
       }
+      elseif ($type == 'group') {
+        $label = $processed_object['#fieldgroups'][$value]->label;
+      }
+      // If a group has a default id, use it as anchor link.
+      if (!empty($processed_object['#fieldgroups'][$value]->format_settings['id'])) {
+        $id = $processed_object['#fieldgroups'][$value]->format_settings['id'];
+      }
+      else {
+        $id = "inline-nav-" . Html::cleanCssIdentifier($label);
+      }
+
+      $links[] = [
+        'href' => "#" . $id,
+        'label' => $label,
+        'field' => $value,
+      ];
+      $element[$value]['#id'] = $id;
+      $element[$value]['#attributes']['id'] = $id;
+      $element[$value]['#in_page_navigation'] = TRUE;
     }
+
     $element['#links'] = $links;
+
     if (count($links) > 0) {
       $element['#inline_title'] = t('Page contents');
     }
@@ -71,70 +77,56 @@ class FieldInPageNavigationFormatter extends FieldGroupFormatterBase {
   }
 
   /**
-   * Check if the field should create an in page navigation entry.
+   * Returns if the field is a group or a field.
    *
    * @param string $field_name
-   *   Name of the field.
-   * @param array $field
-   *   The field to check.
-   * @param array $fieldgroups
+   *   The name of the field.
+   * @param array $field_groups
    *   List of groups.
    *
-   * @return bool
-   *   Include or not the field.
+   * @return string
+   *   The type of field, group or field.
    */
-  private function excludeField(string $field_name, array $field, array $fieldgroups) :bool {
-    // Exclude empty fields.
-    if (!in_array($field_name, array_keys($fieldgroups)) && !isset($field['#items'])) {
-      return TRUE;
+  protected function groupOrField(string $field_name, array $field_groups): string {
+    if (in_array($field_name, array_keys($field_groups))) {
+      return 'group';
     }
-    // Exclude empty groups.
-    if (in_array($field_name, array_keys($fieldgroups)) && count($field) < 1) {
-      return TRUE;
-    }
-    // Exclude groups with only empty fields.
-    if (in_array($field_name, array_keys($fieldgroups))) {
-      foreach ($field as $value) {
-        // If there is a field with values, include the group.
-        if (isset($value['#items'])) {
-          return FALSE;
-        }
-      }
-      return TRUE;
-    }
-    return FALSE;
+    return 'field';
   }
 
   /**
-   * Get the label used on the field.
+   * Returns if the field or group label is visible.
    *
+   * @param string $type
+   *   The type of field (field or group).
    * @param string $field_name
-   *   Name of the field.
+   *   The name of the field.
    * @param array $processed_object
-   *   Entity processed.
+   *   The object / entity beïng processed.
    *
-   * @return string
-   *   Label of the field.
+   * @return bool
+   *   Returns if the label is visible or not.
    */
-  private function fieldGetLabel(string $field_name, array $processed_object): string {
-    $label = '';
-    if (isset($processed_object['#fieldgroups'][$field_name]->label)) {
-      if ($processed_object['#fieldgroups'][$field_name]->label !== '') {
-        $label = $processed_object['#fieldgroups'][$field_name]->label;
-        // We will assume groups labels are shown unless it is empty or
-        // explicitly hidden.
-        if (isset($processed_object['#fieldgroups'][$field_name]->format_settings['show_label']) && $processed_object['#fieldgroups'][$field_name]->format_settings['show_label'] === FALSE) {
-          return '';
-        }
-      }
+  protected function hasVisibleLabel(string $type, string $field_name, array $processed_object): bool {
+    if (
+      $type === 'field' &&
+      in_array($processed_object[$field_name]['#label_display'], ['above', 'inline'])
+    ) {
+      return TRUE;
     }
-    else {
-      // For fields, check if the label is not hidden.
-      if (!in_array($processed_object[$field_name]['#label_display'], ['hidden', 'visually_hidden'])) {
-        return $processed_object[$field_name]['#title'];
-      }
+
+    if (
+      $type === 'group' &&
+      !empty($processed_object['#fieldgroups'][$field_name]) &&
+      (
+        !isset($processed_object['#fieldgroups'][$field_name]->format_settings['show_label']) ||
+        $processed_object['#fieldgroups'][$field_name]->format_settings['show_label'] !== FALSE
+      )
+    ) {
+      return TRUE;
     }
-    return $label;
+
+    return FALSE;
   }
 
 }
--- a/modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
+++ b/modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
@@ -1,0 +1,71 @@
+{#
+/**
+ * @file
+ * Theme override for a field.
+ *
+ * To override output, copy the "field.html.twig" from the templates directory
+ * to your theme's directory and customize it, just like customizing other
+ * Drupal templates such as page.html.twig or node.html.twig.
+ *
+ * Instead of overriding the theming for all fields, you can also just override
+ * theming for a subset of fields using
+ * @link themeable Theme hook suggestions. @endlink For example,
+ * here are some theme hook suggestions that can be used for a field_foo field
+ * on an article node type:
+ * - field--node--field-foo--article.html.twig
+ * - field--node--field-foo.html.twig
+ * - field--node--article.html.twig
+ * - field--field-foo.html.twig
+ * - field--text-with-summary.html.twig
+ * - field.html.twig
+ *
+ * Available variables:
+ * - attributes: HTML attributes for the containing element.
+ * - label_hidden: Whether to show the field label or not.
+ * - title_attributes: HTML attributes for the title.
+ * - label: The label for the field.
+ * - multiple: TRUE if a field can contain multiple items.
+ * - items: List of all the field items. Each item contains:
+ *   - attributes: List of HTML attributes for each item.
+ *   - content: The field item's content.
+ * - entity_type: The entity type to which the field belongs.
+ * - field_name: The name of the field.
+ * - field_type: The type of the field.
+ * - label_display: The display settings for the label.
+ *
+ * @see template_preprocess_field()
+ */
+#}
+{%
+  set title_classes = [
+  'ecl-u-type-heading-2',
+  label_display == 'visually_hidden' ? 'visually-hidden',
+]
+%}
+
+{% if label_hidden %}
+  {% if multiple %}
+    <div{{ attributes }}>
+      {% for item in items %}
+        <div{{ item.attributes }}>{{ item.content }}</div>
+      {% endfor %}
+    </div>
+  {% else %}
+    {% for item in items %}
+      <div{{ attributes }}>{{ item.content }}</div>
+    {% endfor %}
+  {% endif %}
+{% else %}
+  <div{{ attributes }}>
+    <div{{ title_attributes.addClass(title_classes) }}>{{ label }}</div>
+    {% if multiple %}
+    <div>
+      {% endif %}
+      {% for item in items %}
+        <div{{ item.attributes }}>{{ item.content }}</div>
+      {% endfor %}
+      {% if multiple %}
+    </div>
+    {% endif %}
+  </div>
+{% endif %}
--- a/modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
+++ b/modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
@@ -80,42 +80,42 @@ class FieldInPageNavigationFormatterTest extends BrowserTestBase {
         'label' => 'Field 1 label',
         'visibility' => 'above',
         'type' => 'string',
-        'weight' => 10,
+        'weight' => 11,
       ],
       'field_test_2' => [
         'content' => 'Field 2 content',
         'label' => 'Field 2 label',
         'visibility' => 'inline',
         'type' => 'string',
-        'weight' => 11,
+        'weight' => 21,
       ],
       'field_test_3' => [
         'content' => '',
         'label' => 'Field 3 label',
         'visibility' => 'inline',
         'type' => 'string',
-        'weight' => 12,
+        'weight' => 31,
       ],
       'field_test_4' => [
         'content' => 'Field 4 content',
         'label' => 'Field 4 label',
         'visibility' => 'above',
         'type' => 'string',
-        'weight' => 13,
+        'weight' => 41,
       ],
       'field_test_5' => [
         'content' => 'Field 5 content',
         'label' => 'Field 5 label',
         'visibility' => 'above',
         'type' => 'string',
-        'weight' => 14,
+        'weight' => 61,
       ],
       'field_test_6' => [
         'content' => 'Field 6 content',
         'label' => 'Field 6 label',
         'visibility' => 'hidden',
         'type' => 'string',
-        'weight' => 15,
+        'weight' => 62,
       ],
     ];
     $node_values = ['type' => "test"];
@@ -161,27 +161,8 @@ class FieldInPageNavigationFormatterTest extends BrowserTestBase {
   public function testFieldInPageNavigationFormmaterTest() {
     // Create display groups.
     $data = [
-      'label' => 'In-page navigation group',
-      'weight' => 0,
-      'group_name' => 'group_inpage_nav',
-      'children' => [
-        'group_1',
-        'group_2',
-        'group_3',
-        'group_4',
-        'group_5',
-        'group_6',
-        'field_test_5',
-        'field_test_6',
-      ],
-      'format_type' => 'oe_theme_helper_in_page_navigation',
-    ];
-    $this->createGroup('node', 'test', 'view', 'default', $data);
-    $data = [
-      'weight' => 1,
-      'group_name' => 'group_1',
+      'weight' => 10,
       'label' => 'Group 1',
-      'parent' => 'group_inpage_nav',
       'children' => [
         0 => 'field_test_1',
       ],
@@ -190,12 +171,10 @@ class FieldInPageNavigationFormatterTest extends BrowserTestBase {
         'open' => TRUE,
       ],
     ];
-    $this->createGroup('node', 'test', 'view', 'default', $data);
+    $group_1 = $this->createGroup('node', 'test', 'view', 'default', $data);
     $data = [
-      'weight' => 2,
+      'weight' => 20,
       'label' => 'Group 2',
-      'group_name' => 'group_2',
-      'parent' => 'group_inpage_nav',
       'children' => [
         0 => 'field_test_2',
       ],
@@ -205,12 +184,10 @@ class FieldInPageNavigationFormatterTest extends BrowserTestBase {
         'open' => TRUE,
       ],
     ];
-    $this->createGroup('node', 'test', 'view', 'default', $data);
+    $group_2 = $this->createGroup('node', 'test', 'view', 'default', $data);
     $data = [
-      'weight' => 2,
+      'weight' => 30,
       'label' => 'Group 3',
-      'group_name' => 'group_3',
-      'parent' => 'group_inpage_nav',
       'children' => [
         0 => 'field_test_3',
       ],
@@ -219,12 +196,10 @@ class FieldInPageNavigationFormatterTest extends BrowserTestBase {
         'show_label' => TRUE,
       ],
     ];
-    $this->createGroup('node', 'test', 'view', 'default', $data);
+    $group_3 = $this->createGroup('node', 'test', 'view', 'default', $data);
     $data = [
-      'weight' => 3,
+      'weight' => 40,
       'label' => 'Group 4',
-      'group_name' => 'group_4',
-      'parent' => 'group_inpage_nav',
       'children' => [
         0 => 'field_test_4',
       ],
@@ -233,29 +208,42 @@ class FieldInPageNavigationFormatterTest extends BrowserTestBase {
         'open' => TRUE,
       ],
     ];
-    $this->createGroup('node', 'test', 'view', 'default', $data);
+    $group_4 = $this->createGroup('node', 'test', 'view', 'default', $data);
     $data = [
-      'weight' => 4,
+      'weight' => 50,
       'label' => 'Group 5',
-      'group_name' => 'group_5',
-      'parent' => 'group_inpage_nav',
       'format_type' => 'details',
       'format_settings' => [
         'open' => TRUE,
       ],
     ];
-    $this->createGroup('node', 'test', 'view', 'default', $data);
+    $group_5 = $this->createGroup('node', 'test', 'view', 'default', $data);
     $data = [
-      'weight' => 5,
+      'weight' => 60,
       'label' => 'Group 6',
-      'group_name' => 'group_6',
-      'parent' => 'group_inpage_nav',
       'format_type' => 'html_element',
       'format_settings' => [
         'show_label' => FALSE,
       ],
     ];
-    $this->createGroup('node', 'test', 'view', 'default', $data);
+    $group_6 = $this->createGroup('node', 'test', 'view', 'default', $data);
+    $data = [
+      'label' => 'In-page navigation group',
+      'weight' => 0,
+      'group_name' => 'group_inpage_nav',
+      'children' => [
+        $group_1->group_name,
+        $group_2->group_name,
+        $group_3->group_name,
+        $group_4->group_name,
+        $group_5->group_name,
+        $group_6->group_name,
+        'field_test_5',
+        'field_test_6',
+      ],
+      'format_type' => 'oe_theme_helper_in_page_navigation',
+    ];
+    $inline = $this->createGroup('node', 'test', 'view', 'default', $data);
     $html = $this->drupalGet('node/' . $this->node->id());
     $crawler = new Crawler($html);
     // The content is break into two elements, for navigation and content.
--- a/modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
+++ b/modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
@@ -1,0 +1,305 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme_helper\Functional\Plugin\field_group;
+
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\field_group\Functional\FieldGroupTestTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Test the in page navigation field group formatter.
+ */
+class FieldInPageNavigationFormatterTest extends BrowserTestBase {
+
+  use FieldGroupTestTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'node',
+    'text',
+    'field_ui',
+    'field_group',
+    'oe_theme_helper',
+  ];
+
+  /**
+   * A node to use for testing.
+   *
+   * @var \Drupal\node\NodeInterface
+   */
+  protected $node;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    // Enable oe_theme and set it as default.
+    $this->assertTrue($this->container->get('theme_installer')->install(['oe_theme']));
+    $this->container->get('config.factory')
+      ->getEditable('system.theme')
+      ->set('default', 'oe_theme')
+      ->save();
+
+    // Rebuild the ui_pattern definitions to collect the ones provided by
+    // oe_theme itself.
+    $this->container->get('plugin.manager.ui_patterns')->clearCachedDefinitions();
+
+    // Create test user.
+    $admin_user = $this->drupalCreateUser([
+      'access content',
+      'administer content types',
+      'administer node fields',
+      'administer node form display',
+      'administer node display',
+      'bypass node access',
+    ]);
+    $this->drupalLogin($admin_user);
+
+    // Create content type.
+    $this->drupalCreateContentType([
+      'name' => 'Test',
+      'type' => 'test',
+    ]);
+
+    /** @var \Drupal\Core\Entity\Display\EntityViewDisplayInterface $display */
+    $display = \Drupal::entityTypeManager()
+      ->getStorage('entity_view_display')
+      ->load('node.test.default');
+
+    // Create test fields.
+    $fields = [
+      'field_test_1' => [
+        'content' => 'Field 1 content',
+        'label' => 'Field 1 label',
+        'visibility' => 'above',
+        'type' => 'string',
+        'weight' => 10,
+      ],
+      'field_test_2' => [
+        'content' => 'Field 2 content',
+        'label' => 'Field 2 label',
+        'visibility' => 'inline',
+        'type' => 'string',
+        'weight' => 11,
+      ],
+      'field_test_3' => [
+        'content' => '',
+        'label' => 'Field 3 label',
+        'visibility' => 'inline',
+        'type' => 'string',
+        'weight' => 12,
+      ],
+      'field_test_4' => [
+        'content' => 'Field 4 content',
+        'label' => 'Field 4 label',
+        'visibility' => 'above',
+        'type' => 'string',
+        'weight' => 13,
+      ],
+      'field_test_5' => [
+        'content' => 'Field 5 content',
+        'label' => 'Field 5 label',
+        'visibility' => 'above',
+        'type' => 'string',
+        'weight' => 14,
+      ],
+      'field_test_6' => [
+        'content' => 'Field 6 content',
+        'label' => 'Field 6 label',
+        'visibility' => 'hidden',
+        'type' => 'string',
+        'weight' => 15,
+      ],
+    ];
+    $node_values = ['type' => "test"];
+
+    foreach ($fields as $field_name => $field) {
+      $field_storage = FieldStorageConfig::create([
+        'field_name' => $field_name,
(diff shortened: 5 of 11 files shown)


modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,140 @@
...
+    foreach ($field_elements as $key => $value) {
+      // Check if we should add the field to page contents.
+      if (!is_array($element[$value]) || $this->excludeField($value, $element[$value], $processed_object['#fieldgroups']) === TRUE) {
+        continue;
+      }
+      // We will add anchors when the element, field or group, has a label.
+      $label = $this->fieldGetLabel($value, $processed_object);
+      // Add menu element and anchor.
+      if ($label !== '') {
+        // If the group has a default id, use it as anchor link.
```
drishu: we could do a continue

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,140 @@
...
+   *   Name of the field.
+   * @param array $processed_object
+   *   Entity processed.
+   *
+   * @return string
+   *   Label of the field.
+   */
+  private function fieldGetLabel(string $field_name, array $processed_object): string {
+    $label = '';
+    if (isset($processed_object['#fieldgroups'][$field_name]->label)) {
```
drishu: can we combine the ifs?

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,140 @@
...
+      if ($processed_object['#fieldgroups'][$field_name]->label !== '') {
+        $label = $processed_object['#fieldgroups'][$field_name]->label;
+        // We will assume groups labels are shown unless it is empty or
+        // explicitly hidden.
+        if (isset($processed_object['#fieldgroups'][$field_name]->format_settings['show_label']) && $processed_object['#fieldgroups'][$field_name]->format_settings['show_label'] === FALSE) {
+          return '';
+        }
+      }
+    }
+    else {
```
drishu: else if

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,293 @@
...
+   */
+  protected $node;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setUp() {
+    parent::setUp();
+    // Enable oe_theme and set it as default.
+    $this->assertTrue($this->container->get('theme_installer')->install(['oe_theme']));
```
drishu: we need to clarify if we can use the containers in setup or not

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,293 @@
...
+        'weight' => $field['weight'],
+      ]);
+      // Create values for the testing node.
+      if ($field['content'] !== '') {
+        $node_values[$field_name][0]['value'] = $field['content'];
+      }
+    }
+    // Save display + create node.
+    $display->save();
+    $this->node = $this->drupalCreateNode($node_values);
```
drishu: I'd rather you have $fields as class attribute instead of a node, create the content type here, but the node you create before testing against

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  /**
+   * Test the in page navigation field group formatter.
+   */
+  public function testFieldInPageNavigationFormmaterTest() {
+    // Create display groups.
```
drishu: which is the logic behind creating fields in setup, but groups here ?
it would make sense only if you try different setups of groups I think

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,293 @@
...
+      ->getEditable('system.theme')
+      ->set('default', 'oe_theme')
+      ->save();
+
+    // Rebuild the ui_pattern definitions to collect the ones provided by
+    // oe_theme itself.
+    $this->container->get('plugin.manager.ui_patterns')->clearCachedDefinitions();
+
+    // Create test user.
+    $admin_user = $this->drupalCreateUser([
```
drishu: this is not needed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,140 @@
...
+   *   Entity processed.
+   *
+   * @return string
+   *   Label of the field.
+   */
+  private function fieldGetLabel(string $field_name, array $processed_object): string {
+    $label = '';
+    if (isset($processed_object['#fieldgroups'][$field_name]->label)) {
+      if ($processed_object['#fieldgroups'][$field_name]->label !== '') {
+        $label = $processed_object['#fieldgroups'][$field_name]->label;
```
drishu: can we make sure this returns a string

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,140 @@
...
+        // explicitly hidden.
+        if (isset($processed_object['#fieldgroups'][$field_name]->format_settings['show_label']) && $processed_object['#fieldgroups'][$field_name]->format_settings['show_label'] === FALSE) {
+          return '';
+        }
+      }
+    }
+    else {
+      // For fields, check if the label is not hidden.
+      if (!in_array($processed_object[$field_name]['#label_display'], ['hidden', 'visually_hidden'])) {
+        return $processed_object[$field_name]['#title'];
```
drishu: we need to make sure this returns a string, try testing with a extra field https://git.drupalcode.org/project/extra_field/-/blob/8.x-1.x/modules/extra_field_example/src/Plugin/ExtraField/Display/ExampleFormattedField.php

drishu (changes requested):

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: I used this from this other test: modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/PatternFormatterTest.php

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Changed

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Changed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: I changed this process

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: I changed this process

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Changed to ensure we always get a string

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Changed to ensure we always get a string

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Changed to the test method

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Removed

modules/oe_theme_helper/src/Element/InLinePaginationElement.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+
+/**
+ * Provides a render element for a html element.
```
nagyad: Provides a render element for an html element.

modules/oe_theme_helper/src/Element/InLinePaginationElement.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @FormElement("field_group_in_page_navigation")
+ */
+class InLinePaginationElement extends RenderElement {
```
nagyad: The component is subject is in-page navigation, so this should be InPageNavigationElement.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Plugin\field_group\FieldGroupFormatter;
+
+use Drupal\field_group\FieldGroupFormatterBase;
+
+/**
+ * Format a field group using the field list pattern.
```
nagyad: Copy-pasta.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * @FieldGroupFormatter(
+ *   id = "oe_theme_helper_in_page_navigation",
+ *   label = @Translation("Field In-page navigation"),
```
nagyad: Field in-page navigation

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+class FieldInPageNavigationFormatter extends FieldGroupFormatterBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function process(&$element, $processed_object) {
+    $element += [
+      '#type' => 'field_group_in_page_navigation',
+    ];
+    // Get children elements in in page navigation group.
```
nagyad: Get children elements from the in-page navigation group.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+    $field_elements = $processed_object['#fieldgroups'][$element['#group_name']]->children;
+    // Add anchors links and anchors to children with labels.
```
nagyad: This comment is not needed.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+    $links = [];
+    $i = 1;
+    foreach ($field_elements as $key => $value) {
+      // Check if we should add the field to page contents.
+      if (!is_array($element[$value]) || $this->excludeField($value, $element[$value], $processed_object['#fieldgroups']) === TRUE) {
+        continue;
+      }
+      // We will add anchors when the element, field or group, has a label.
+      $label = $this->fieldGetLabel($value, $processed_object);
+      // If there is no label, don't add if to the menu.
```
nagyad: Typo, but I think this is not required.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+      if ($label === '') {
+        continue;
+      }
+      // If a group has a default id, use it as anchor link.
+      $type = $this->groupOrField($value, $processed_object['#fieldgroups']);
+      if ($type === 'group' && isset($processed_object['#fieldgroups'][$value]->format_settings['id']) && $processed_object['#fieldgroups'][$value]->format_settings['id'] !== '') {
```
nagyad: empty() checks isset() as well, so could simplify this if to:
`if ($type === 'group' && !empty($processed_object['#fieldgroups'][$value]->format_settings['id'])) {`

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+      if (!is_array($element[$value]) || $this->excludeField($value, $element[$value], $processed_object['#fieldgroups']) === TRUE) {
+        continue;
+      }
+      // We will add anchors when the element, field or group, has a label.
+      $label = $this->fieldGetLabel($value, $processed_object);
+      // If there is no label, don't add if to the menu.
+      if ($label === '') {
+        continue;
+      }
+      // If a group has a default id, use it as anchor link.
```
nagyad, line 46: Move this comment into the if statement.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+      $type = $this->groupOrField($value, $processed_object['#fieldgroups']);
+      if ($type === 'group' && isset($processed_object['#fieldgroups'][$value]->format_settings['id']) && $processed_object['#fieldgroups'][$value]->format_settings['id'] !== '') {
+        $id = $processed_object['#fieldgroups'][$value]->format_settings['id'];
+      }
+      else {
+        $id = "inline-nav-" . $i++;
+      }
```
nagyad, line 52: Empty line after the if would be nice.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+      $links[] = [
+        'href' => "#" . $id,
+        'label' => $label,
+      ];
+      $element[$value]['#id'] = $id;
+      $element[$value]['#attributes']['id'] = $id;
+    }
```
nagyad, line 62: Empty line after the foreach would be nice.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+      if ($label === '') {
+        continue;
+      }
+      // If a group has a default id, use it as anchor link.
+      $type = $this->groupOrField($value, $processed_object['#fieldgroups']);
+      if ($type === 'group' && isset($processed_object['#fieldgroups'][$value]->format_settings['id']) && $processed_object['#fieldgroups'][$value]->format_settings['id'] !== '') {
+        $id = $processed_object['#fieldgroups'][$value]->format_settings['id'];
+      }
+      else {
+        $id = "inline-nav-" . $i++;
```
nagyad: We could ending up not even using the $i index. Try to think about a better way to use a number here without the index if possible.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+   *   Name of the field.
+   * @param array $field
+   *   The field to check.
+   * @param array $fieldgroups
+   *   List of groups.
+   *
+   * @return bool
+   *   Include or not the field.
+   */
+  private function excludeField(string $field_name, array $field, array $fieldgroups) :bool {
```
nagyad: This should be protected.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function preRender(&$element, $rendering_object) {
+    parent::preRender($element, $rendering_object);
+    $this->process($element, $rendering_object);
+  }
+
+  /**
+   * Check if the field should create an in page navigation entry.
```
nagyad: Actually, your function is checking if a field should be excluded from the in-page navigation.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+   *
+   * @param string $field_name
+   *   Name of the field.
+   * @param array $field
+   *   The field to check.
+   * @param array $fieldgroups
```
nagyad: $fieldgroups should be $field_groups

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
nagyad: $fieldgroups should be $field_groups

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+          return TRUE;
+        }
+
+        break;
+
+      case 'group':
+        // Exclude empty groups.
+        if (count($field) < 1) {
+          return TRUE;
+        }
```
nagyad: Empty line would be nice after the if statement.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+        foreach ($field as $value) {
+          // If there is a field with values, don't exclude the group.
+          if (isset($value['#items'])) {
+            return FALSE;
+          }
+        }
```
nagyad: Empty line would be nice after the foreach statement here.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+   *
+   * @param string $field_name
+   *   Name of the field.
+   * @param array $field
+   *   The field to check.
+   * @param array $fieldgroups
+   *   List of groups.
+   *
+   * @return bool
+   *   Include or not the field.
```
nagyad: Exclude or not...

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+   *
+   * @param string $field_name
+   *   Name of the field.
+   * @param array $processed_object
+   *   Entity processed.
+   *
+   * @return string
+   *   Label of the field.
+   */
+  private function fieldGetLabel(string $field_name, array $processed_object): string {
```
nagyad: This should be protected.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+   *
+   * @return bool
+   *   Include or not the field.
+   */
+  private function excludeField(string $field_name, array $field, array $fieldgroups) :bool {
+    $type = $this->groupOrField($field_name, $fieldgroups);
+    switch ($type) {
+      case "field":
+        // Exclude empty fields.
+        if ($type === 'field' && !isset($field['#items'])) {
```
nagyad: You are checking the type in the switch, why are you checking here again the same thing?

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+   *
+   * @param string $field_name
+   *   The name of the field.
+   * @param array $fieldgroups
+   *   List of groups.
+   *
+   * @return string
+   *   The type of field, group or field.
+   */
+  private function groupOrField(string $field_name, array $fieldgroups): string {
```
nagyad: This should be protected.

nagyad: $fieldgroups should be $field_groups

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Get the label of the field.
+   *
+   * @param array $field
+   *   Field info.
+   *
+   * @return string
+   *   The label of the field.
+   */
+  private function getFieldLabel(array $field): string {
```
nagyad: You are using this function only once, could we merge it with fieldGetLabel and rename that function to getFieldLabel?

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+      }
+      if (is_object($title) && method_exists($title, "__toString")) {
+        return $title->__toString();
+      }
+    }
+    return '';
+  }
+
+  /**
+   * Get the label of the field.
```
nagyad: You are getting the label of the group.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,201 @@
...
+   *
+   * @param object $group
+   *   Group info.
+   *
+   * @return string
+   *   The label of the group.
+   */
+  private function getGroupLabel($group): string {
+    // If the label is hidden or empty, don't show it.
+    if ($group->label === '' || isset($group->format_settings['show_label']) && $group->format_settings['show_label'] === FALSE) {
```
nagyad: Why don't you just check: empty($group->label)?

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+namespace Drupal\Tests\oe_theme_helper\Functional\Plugin\field_group;
+
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\field_group\Functional\FieldGroupTestTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Test the in page navigation field group formatter.
```
nagyad: Tests the in-page navigation field group formatter.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+  protected static $modules = [
+    'node',
+    'text',
+    'field_ui',
+    'field_group',
+    'oe_theme_helper',
+  ];
+
+  /**
+   * Fields configuration for testing.
```
nagyad: Field configurations for testing.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
nagyad: Why the assert in the setup method?

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+        'type' => 'text_default',
+        'weight' => $field['weight'],
+      ]);
+    }
+    // Save display + create node.
+    $display->save();
+  }
+
+  /**
+   * Test the in page navigation field group formatter.
```
nagyad: Tests the in-page navigation field group formatter.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+   */
+  public function testFieldInPageNavigationFormmaterTest() {
```
nagyad: Test word is redundant here.

nagyad: Where is the coding style formatting from this function?

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+    $node_values = ['type' => "test"];
+    foreach ($this->fields as $field_name => $field) {
+      if ($field['content'] !== '') {
+        $node_values[$field_name][0]['value'] = $field['content'];
+      }
+    }
+    $node = $this->drupalCreateNode($node_values);
+    $html = $this->drupalGet('node/' . $node->id());
+    $crawler = new Crawler($html);
+    // The content is break into two elements, for navigation and content.
```
nagyad: ... is broken ...

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+      // Link label.
+      $this->assertEquals($value['label'], trim($element->text()));
+      // Content label.
+      $this->assertSession()->elementTextContains('css', $content_selector . " #" . $value['id'] . " > " . $value['selector'], $value['label']);
+    }
+    // If a group has no fields, don't include it in the navigation.
+    $this->assertSession()->pageTextNotContains('Group 5');
+    // If a group has no visible label, don't include it in the navigation.
+    $this->assertSession()->pageTextNotContains('Group 6');
+    // If a field has no visible label, don't include it in the navigation.
```
nagyad: If a field has label but hidden, don't include it in the navigation.

nagyad (changes requested):

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
drishu: because of possible false positives like 0

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: I took this from modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/PatternFormatterTest.php, not needed for testing, should we remove it from modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/PatternFormatterTest.php too?

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Other test methods include "Test" on the comments, for example https://github.com/openeuropa/oe_theme/blob/e98e4a218f3c5fd37015de8cc58d964b4ea96baf/tests/Functional/ContentEventRenderTest.php#L54 
should we include it or exclude it on the methods test we implement?

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Added line

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Added line

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Added line

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Added line

modules/oe_theme_helper/src/Element/InLinePaginationElement.php
enriquelacoma: Updated

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: updated

modules/oe_theme_helper/src/Element/InLinePaginationElement.php
enriquelacoma: updated

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Updated

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Updated

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: updated

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Updated

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
nagyad: My comment is addressing the Test in the end of the method name. testFieldInPageNavigationFormmater**Test**

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Changed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Changed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Changed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Removed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Changed

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Changed

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Changed

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Changed

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Changed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Changed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: I merged getFieldLabel into fieldGetLabel and kept getGroupLabel as a separat method.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: We need and incremental unique id to identify anchors, i changed the creation to this:
`      else {
        $id = "inline-nav-" . $i = isset($i) ? ++$i : 1;
      }`
This is more compact but can be harder to read.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
ademarco: ouch! I'm afraid that's wrong there too... please remove it from both.

modules/oe_theme_helper/src/Element/InPageNavigationElement.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Drupal\oe_theme_helper\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+
+/**
+ * Provides a render element for an html element.
+ *
+ * @FormElement("field_group_in_page_navigation")
+ */
+class InPageNavigationElement extends RenderElement {
```
ademarco, line 14: Why do we need to use a render element for this? Isn't an ordinary theme declaration enough?

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+namespace Drupal\oe_theme_helper\Plugin\field_group\FieldGroupFormatter;
+
+use Drupal\field_group\FieldGroupFormatterBase;
+
+/**
+ * Format a field group to create in-page navigation.
+ *
+ * @FieldGroupFormatter(
+ *   id = "oe_theme_helper_in_page_navigation",
+ *   label = @Translation("Field in-page navigation"),
```
ademarco:
```suggestion
 *   label = @Translation("In-page navigation"),
```

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+ *   description = @Translation("Format field group for ecl-inpage-navigation."),
```
ademarco:
```suggestion
 *   description = @Translation("Display a field group using the ECL in-page navigation component."),
```

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Plugin\field_group\FieldGroupFormatter;
+
+use Drupal\field_group\FieldGroupFormatterBase;
+
+/**
+ * Format a field group to create in-page navigation.
```
ademarco:
```suggestion
 * Display a field group using the ECL in-page navigation component.
```

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+ * )
+ */
+class FieldInPageNavigationFormatter extends FieldGroupFormatterBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function process(&$element, $processed_object) {
+    $element += [
+      '#type' => 'field_group_in_page_navigation',
```
ademarco, line 30: Can't this be a simple theme function? See above.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+    ];
+    // Get children elements from the in-page navigation group.
+    $field_elements = $processed_object['#fieldgroups'][$element['#group_name']]->children;
```
ademarco: Isn't this enough?
```php
    foreach (Element::children($element) as $field_name) {
```

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+    $links = [];
+    foreach ($field_elements as $value) {
+      // Check if we should add the field to page contents.
+      if (!is_array($element[$value]) || $this->excludeField($value, $element[$value], $processed_object['#fieldgroups']) === TRUE) {
```
ademarco: Determining emptiness in a render array is known to be problematic in Drupal, the excludeField method tries to do that by trying to handle all possible cases (namely `field` and `group`), but that's not ideal as it doesn't take care of, for example, extra fields.

We had a similar issue on `oe_theme_helper_field_list_pattern` and, at the end, we chose to render the fields in the template and check for emptiness there:

```twig
{% for _item in items %}
  {% set _body = _item.body|render %}
  {% if _body %}
    {% set _items = _items|merge([{
      term: _item.label,
      definition: _body
    }]) %}
  {% endif %}
```

I suggest to do the same here and to get rid of all the "establish if empty" logic.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+        continue;
+      }
+      // We will add anchors when the element, field or group, has a label.
+      $label = $this->fieldGetLabel($value, $processed_object);
```
ademarco: I like the idea of only displaying links if the field group item has a label or not, but let's just check if `#label_display` is `visible`. Also let's have a method only dedicated to that, call it `hasVisibleLabel(...): bool`.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+        continue;
+      }
+      $type = $this->groupOrField($value, $processed_object['#fieldgroups']);
+      // If a group has a default id, use it as anchor link.
+      if ($type === 'group' && !empty($processed_object['#fieldgroups'][$value]->format_settings['id'])) {
+        $id = $processed_object['#fieldgroups'][$value]->format_settings['id'];
+      }
+      else {
+        $id = "inline-nav-" . $i = isset($i) ? ++$i : 1;
+      }
```
ademarco, line 52: Let's not again discriminate if it's a group or not, etc. let's just use the label to generate the anchor ID. Use `\Drupal\Component\Utility\Html::cleanCssIdentifier()`, which takes care of uniqueness already.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+   *   Name of the field.
+   * @param array $field
+   *   The field to check.
+   * @param array $field_groups
+   *   List of groups.
+   *
+   * @return bool
+   *   Exclude or not the field.
+   */
+  protected function excludeField(string $field_name, array $field, array $field_groups) :bool {
```
ademarco: Remove this method as we will check for emptiness in the template.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @param string $field_name
+   *   Name of the field.
+   * @param array $processed_object
+   *   Entity processed.
+   *
+   * @return string
+   *   Label of the field.
+   */
+  protected function fieldGetLabel(string $field_name, array $processed_object): string {
```
ademarco: Remove this method, as we will have one dedicated to establishing if the item should be included in the navigation or not (see above).

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,188 @@
...
+    switch ($type) {
+      case "field":
+        // Show the label if it is not hidden.
+        if (!in_array($processed_object[$field_name]['#label_display'], ['hidden', 'visually_hidden'])) {
+          $title = $processed_object[$field_name]['#title'];
+          if (is_string($title)) {
+            $label = $processed_object[$field_name]['#title'];
+          }
+          if (is_object($title) && method_exists($title, "__toString")) {
+            $label = $title->__toString();
```
ademarco: You don't need to do this, if objects labels are renderable as strings.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\field_group\Functional\FieldGroupTestTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Test the in-page navigation field group formatter.
+ */
+class FieldInPageNavigationFormatterTest extends BrowserTestBase {
```
ademarco:
```suggestion
class InPageNavigationFormatterTest extends BrowserTestBase {
```

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+    // Enable oe_theme and set it as default.
+    $this->assertTrue($this->container->get('theme_installer')->install(['oe_theme']));
+    $this->container->get('config.factory')
+      ->getEditable('system.theme')
+      ->set('default', 'oe_theme')
+      ->save();
+
+    // Rebuild the ui_pattern definitions to collect the ones provided by
+    // oe_theme itself.
+    $this->container->get('plugin.manager.ui_patterns')->clearCachedDefinitions();
```
ademarco: We don't use patterns in this test, remove this.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
```diff
@@ -0,0 +1,284 @@
...
+      ]);
+    }
+    // Save display + create node.
+    $display->save();
+  }
+
+  /**
+   * Tests the in-page navigation field group formatter.
+   */
+  public function testFieldInPageNavigationFormmater(): void {
```
ademarco:
```suggestion
  public function testFormmater(): void {
```
No need to repeat the whole class name in the method name.

ademarco (changes requested):

ademarco (changes requested):
We should also make sure that the titles receive H2 styles too, automatically:

image

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Moved

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Updated.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: $this->assertTrue removed from both files

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Updated

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Updated

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Updated

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Yes, this works better.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: I moved the check to the template, and removed it from the formatter

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: I implemented hasVisibleLabel(), the way visibility is controlled on fields and groups is different, so we need to control both cases.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Removed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Removed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: Removed

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/FieldInPageNavigationFormatterTest.php
enriquelacoma: Removed

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
enriquelacoma: I used an implementation similar to Drupal\field_group\Plugin\field_group\FieldGroupFormatter\HtmlElement 
When I tried using a theme function, something like:
`
'#theme' => 'oe_theme_helper_in_page_navigation'
`
$element['#children'] is empty

modules/oe_theme_helper/src/Element/InPageNavigationElement.php
drishu: tested, its not enough, plus I could not find a group formatter without one to have such an implementation as example

drishu (approved):

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/FieldInPageNavigationFormatter.php
```diff
@@ -0,0 +1,132 @@
...
+   * @param string $field_name
+   *   The name of the field.
+   * @param array $processed_object
+   *   The object / entity beïng processed.
+   *
+   * @return bool
+   *   Returns if the label is visible or not.
+   */
+  protected function hasVisibleLabel(string $type, string $field_name, array $processed_object): bool {
+    if (
```
yenyasinn, line 111: Such style of conditions isn't used anywhere on the project.

modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
```diff
@@ -0,0 +1,71 @@
...
+  {% else %}
+    {% for item in items %}
+      <div{{ attributes }}>{{ item.content }}</div>
+    {% endfor %}
+  {% endif %}
+{% else %}
+  <div{{ attributes }}>
+    <div{{ title_attributes.addClass(title_classes) }}>{{ label }}</div>
+    {% if multiple %}
+    <div>
```
yenyasinn, line 62: Why do we need empty div here?

modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
```diff
@@ -0,0 +1,71 @@
...
+      {% endfor %}
+    </div>
+  {% else %}
+    {% for item in items %}
+      <div{{ attributes }}>{{ item.content }}</div>
+    {% endfor %}
+  {% endif %}
+{% else %}
+  <div{{ attributes }}>
+    <div{{ title_attributes.addClass(title_classes) }}>{{ label }}</div>
```
yenyasinn, line 60: Based on https://ec.europa.eu/component-library/playground/ec/?path=/story/components-navigation-in-page-navigation--default label has to be wrapped by H2.

modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
```diff
@@ -0,0 +1,71 @@
+{#
+/**
+ * @file
+ * Theme override for a field.
+ *
+ * To override output, copy the "field.html.twig" from the templates directory
```
yenyasinn, line 6: It is copy paste from build/core/themes/stable/templates/field/field.html.twig. Do we need so long description?  Might be ` * @see ./core/themes/stable/templates/field/field.html.twig` is enough?

modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
yenyasinn: It is in build/core/themes/stable/templates/field/field.html.twig but why do we need it here?

modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
drishu: I agree

modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
drishu: yes

modules/oe_theme_helper/templates/field--in-page-navigation.html.twig
```diff
@@ -0,0 +1,71 @@
+{#
```
yenyasinn, line 1: It works for fields only. Extra fields and field groups don't work now.

yenyasinn:
Close it in favour https://github.com/openeuropa/oe_theme/pull/544/files

PR #531, opened 2020-08-05
EWPP-113: Update ECL to version 2.31.0.

nagyad:


modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -258,11 +258,13 @@ public function toEclAttributes($attributes): array {
...
    * @param string|null $icon
    *   The icon to be converted.
+   * @param string $size
+   *   The icon size.
    *
    * @return array
    *   Icon array for ECL components containing icon name, path and rotation.
    */
-  public function toEclIcon(array $context, $icon): array {
+  public function toEclIcon(array $context, $icon, string $size = ''): array {
```
22Alexandra, line 267: https://ecl-twig-js.netlify.app/ec/?path=/story/components-icon--branded I see there are 3 optional elements but only size was added. Should we add also the others or size wouldn't be needed neither?

22Alexandra (commented):
This test is failing Drupal\Tests\oe_theme_helper\Functional\NodeViewRoutesMetadataTest::testNodeRoutes

22Alexandra (dismissed):

imanoleguskiza (dismissed):

22Alexandra:
@ademarco Cannot test the default size for the icon template since we always have the size set in the templates and where we don't set it, ecl would override it.

I found an issue in the featured_item pattern where we use the ecl card template. The ecl card template overrides the icon size we send with 'xs' but only for the first item.

Set the size xl in our featured_item template for the icon in _footer_items array.
size set for all items

First icon gets size 'xs' and the rest keep the size 'xl'.
override only first item

We need to set the size 'xs' in our template, otherwise the default size for icon would be set in this case (size 'm') but this would cause inconsistency (first item having size 'xs' and the rest of the items size 'm').

22Alexandra:
Had few discussions with Davide Ferrante from FRONT team whom managed to reproduce the issue with the card template.

imanoleguskiza:
Regarding the issue @22Alexandra brought up, currently we have it under control because as she mentioned we already pass the size on the text_with_featured_media pattern. Since Davide is already aware of the issue they will fix it in their end and once its fix we can remove the 'xs' on our side. 
About testing the default icon size, we already can see it when we test the icons_with_text pattern: no size is given there and the  "ecl-icon--m" is correctly rendered.

imanoleguskiza (dismissed):

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
```diff
@@ -262,18 +262,20 @@ protected static function getEuropeanUnionLanguageList(): array {
...
+   * @param string|null $size
+   *   The icon size.
    * @param array $expected_icon_array
    *   The icon array to be rendered.
    *
    * @covers ::toEclIcon
    * @dataProvider toEclIconProvider
    */
-  public function testToEclIcon(string $icon_name, array $expected_icon_array) {
+  public function testToEclIcon(string $icon_name, string $size = NULL, array $expected_icon_array) {
```
ademarco: About `string $size = NULL,` let's move optional parameters at the end, after all the mandatory ones. Or, if we want to keep it here, let's remove `= NULL`, it's a test anyway, we control what's passed to this method in `toEclIconProvider`

modules/oe_theme_helper/tests/src/Unit/TwigExtensionTest.php
```diff
@@ -262,18 +262,20 @@ protected static function getEuropeanUnionLanguageList(): array {
...
     $context = ['ecl_icon_path' => '/path/to/theme/resources/icons/'];
     // We join the resulting array from to_ecl_icon() function so that we have
     // a visual representation of the array being returned by the function.
-    $result = $this->twig->render("{{ to_ecl_icon('$icon_name')|join('|') }}", $context);
-    $this->assertEquals(implode('|', $expected_icon_array), $result);
+    $result = $this->twig->render("{{ to_ecl_icon('$icon_name', '$size')|join('|') }}", $context);
+    $this->assertEquals(implode('|', array_filter($expected_icon_array)), $result);
```
ademarco: We don't test when we don't pass any size to the function, like now we test `to_ecl_icon('$icon_name', '')` but never `to_ecl_icon('$icon_name')`, let's add an if and, if `$size === NULL`, then we use `to_ecl_icon('$icon_name')` in our test, like that we can test both cases, when not passed or when passed empty (it could happened if size is a variable coming from somewhere else).

ademarco (dismissed):

imanoleguskiza (approved):

sergepavle (approved):
Ready for merge. Detected issue with Gallery component could be handled in ticket EWPP-180.

PR #534, opened 2020-08-12
EWPP-131: Theme EC and EU footers with latest ECL Twig components.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


ademarco:
Closed as we now need to factor in core/standardised branding too. This PR can be still used as reference.

PR #538, opened 2020-08-18
EWPP-127: Non required Venue and Organiser fields

julien-:
Nothing needs to be changed about Venue and Organiser fields display.
But testing Event content type, **Description** label was displayed even if the field value is empty.

In front the "text" variable equals to :

```
array (size=1)
  '#cache' => 
    array (size=3)
      'contexts' => 
        array (size=0)
          empty
      'tags' => 
        array (size=0)
          empty
      'max-age' => int -1
```

Insofar as the array is not empty, the label is rendered even if no value have to be rendered

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "openeuropa/behat-transformation-context": "~0.1",
         "openeuropa/code-review": "~1.1",
         "openeuropa/drupal-core-require-dev": "^8.7",
-        "openeuropa/oe_content": "~1.8",
+        "openeuropa/oe_content": "dev-EWPP-125",
         "openeuropa/oe_corporate_blocks": "~2.2",
         "openeuropa/oe_corporate_countries": "~1.0.0-beta1",
         "openeuropa/oe_media": "~1.8",


templates/patterns/text_featured_media/pattern-text-featured-media.html.twig
```diff
@@ -11,10 +11,10 @@
   {% set video_ratio = '16-9' %}
 {% endif %}
 
-{% if title %}
-  <h2 class="ecl-u-type-heading-2 ecl-u-mt-2xl ecl-u-mt-md-3xl ecl-u-mb-l">{{ title }}</h2>
-{% endif %}
-{% if image or video or text %}
+{% if image or video or text|render %}
```
drishu: not needed

nagyad:
Where are the changes? Under commit "EWPP-127: Empty Text featured media cached text." I don't see the corresponding changes.

drishu:
> Where are the changes? Under commit "EWPP-127: Empty Text featured media cached text." I don't see the corresponding changes.

Initially we made small changes but it proved they where not needed, so this branch right now is here to just prove that tests are green (no need to merge, can be closed as soon as EWPP-125 is merged)

PR #539, opened 2020-08-19
EWPP-181: Add page header metadata for project content type.

julien-:


tests/features/page-header.feature
```diff
@@ -78,6 +78,16 @@ Feature: Page header block component.
     # The default text format should be applied, converting URLs into links.
     And I should see the link "http://www.example.org" in the "page header intro"
 
+  Scenario: Project content type has custom metadata shown in the page header.
```
drishu: this could have been done also here https://github.com/openeuropa/oe_theme/blob/EWPP-181/tests/Functional/ContentProjectRenderTest.php#L145, but since we have this feature I guess its ok.
@ademarco what do you think?

ademarco: Yep, we need to use functional tests from now on. Lets move it there.

ademarco (changes requested):

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -143,6 +143,10 @@ public function testProjectRendering(): void {
 
     $this->drupalGet($node->toUrl());
 
+    // Assert page header - metadata.
+    $project_metadata = $this->assertSession()->elementExists('css', '.ecl-page-header');
+    $this->assertContains('Summary', $project_metadata->getText());
```
ademarco: Check the specific summary HTML element, not the whole header, also assert "Project" in its metadata HTML element, you can use CSS selector for it.

ademarco (changes requested):

imanoleguskiza (approved):

ademarco (approved):

PR #542, opened 2020-08-19
EWPP-133: Use menu component on navigation-menu pattern.

imanoleguskiza:
### Description

Update the navigation menu pattern to use the new menu ecl component.

### Change log

- Added:
- Changed: Update the navigation menu pattern to use the new menu ecl component.
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/templates/patterns/navigation_menu/navigation_menu.ui_patterns.yml
+++ b/templates/patterns/navigation_menu/navigation_menu.ui_patterns.yml
@@ -11,16 +11,6 @@ navigation_menu:
       label: "Label"
       description: "Label for the menu, only visible on mobile."
       preview: "Navigation"
-    close:
-      type: "text"
-      label: "Close button label"
-      description: "Label for closing the menu popup, only visible on mobile."
-      preview: "Close"
-    back:
-      type: "text"
-      label: "Back button label"
-      description: "Label for moving back on the menu popup, only visible on mobile."
-      preview: "Back"
     items:
       type: "array"
       label: "Links"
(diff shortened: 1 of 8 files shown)


templates/patterns/navigation_menu/navigation_menu.ui_patterns.yml
```diff
@@ -11,6 +11,16 @@ navigation_menu:
       label: "Label"
       description: "Label for the menu, only visible on mobile."
       preview: "Navigation"
+    close:
```
sergepavle: But as I see in the ticket description we should not change the definition of the pattern or create new variant.

imanoleguskiza: I think in this case it should be fine since we are adding things to the pattern, so it will be backwards compatible. What do you think @ademarco ?

ademarco: I haven't looked at this PR yet, but in theory we can add new, non mandatory, fields, as that would be BC.

imanoleguskiza: After further discussion, we decided to remove the close and back files since they are not going to be customizable.

sergepavle (dismissed):

ademarco (approved):

sergepavle (approved):

PR #544, opened 2020-08-21
EWPP-55: Create in-page navigation field group formatter.

yenyasinn:
### Description

Two field groups have been added: "In-page navigation", "In-page navigation item".


modules/oe_theme_helper/templates/oe-theme-helper-in-page-navigation.html.twig
```diff
@@ -0,0 +1,32 @@
...
+
+<div {{ attributes.addClass(classes) }}>
+  <div class="ecl-col-12 ecl-col-lg-3">
+    {% include '@ecl-twig/inpage-navigation' with {
+      'title': title,
+      'links': links
+    } %}
+  </div>
+  <div class="ecl-col-lg-9">
+    <main>
```
yenyasinn: Tag `main` doesn't exist in the component https://ec.europa.eu/component-library/playground/ec/?path=/story/components-navigation-in-page-navigation--default but exists in the page https://ec.europa.eu/component-library/playground/ec/?path=/story/templates-pages--call-for-proposals

modules/oe_theme_helper/src/Element/InPageNavigation.php
```diff
@@ -0,0 +1,33 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+
+/**
+ * Provides a render element for an html element.
+ *
+ * @RenderElement("oe_theme_helper_in_page_navigation")
```
yenyasinn: We should use render element for the field group - it allows to render tree easily. Otherwise children field groups aren't rendered and we need some workarounds to make it work. field_group module use this way to render its groups.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigation.php
```diff
@@ -0,0 +1,97 @@
...
+    $children = Element::children($element);
+    $length = count($children);
+    $i = 0;
+    foreach ($children as $value) {
+      // Mark first and last elements. It is helpful in theming.
+      if ($i === 0) {
+        $element[$value]['#first'] = TRUE;
+      }
+      if ($i === $length - 1) {
+        $element[$value]['#last'] = TRUE;
```
yenyasinn: I use it in modules/oe_theme_helper/templates/oe-theme-helper-in-page-navigation-item.html.twig.

modules/oe_theme_helper/oe_theme_helper.theme.inc
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * @file
+ * Preprocessors and theme functions of OpenEuropa Theme Helper module.
+ */
+
+declare(strict_types = 1);
+
+/**
+ * Prepares variables for in-page navigation..
```
drishu: just one dot at the end

modules/oe_theme_helper/oe_theme_helper.theme.inc
```diff
@@ -0,0 +1,93 @@
...
+ * @param array $variables
+ *   An array of variables.
+ */
+function template_preprocess_oe_theme_helper_in_page_navigation(array &$variables) {
+  $element = $variables['element'];
+
+  $variables['title'] = !empty($element['#title']) ? $element['#title'] : '';
+  $variables['children'] = !empty($element['#children']) ? $element['#children'] : '';
+  $variables['links'] = !empty($element['#links']) ? $element['#links'] : '';
+
```
drishu: this new line is not needed

modules/oe_theme_helper/oe_theme_helper.theme.inc
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Prepares variables for in-page navigation..
```
drishu: just one dot here, for in-page navigation item.

modules/oe_theme_helper/src/Element/InPageNavigation.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+
+/**
+ * Provides a render element for an html element.
```
drishu: for an in-page navigation element.

modules/oe_theme_helper/src/Element/InPageNavigationItem.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+
+/**
+ * Provides a render element for an html element.
```
drishu: for an in-page navigation item element.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/InPageNavigationTest.php
```diff
@@ -0,0 +1,249 @@
...
+    $instance = FieldConfig::create([
+      'field_storage' => $field_storage,
+      'bundle' => $this->type,
+      'label' => "Field label $field_name",
+    ]);
+    $instance->save();
+  }
+
+  /**
+   * Asserts field froup header.
```
drishu: typo, group header.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigation.php
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * Returns if the field is a group or a field.
+   *
+   * @param string $field_name
+   *   The name of the field.
+   * @param array $field_groups
+   *   List of groups.
+   *
+   * @return mixed
```
drishu: this is just a suggestion, but we could go here with
```suggestion
   * @return \stdClass|null
```

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigation.php
```diff
@@ -0,0 +1,97 @@
...
+   *   Object if field group is found, FALSE otherwise.
+   */
+  protected function getInPageNavigationItemGroup(string $field_name, array $field_groups) {
```
drishu: and if we go with above, then
```suggestion
  protected function getInPageNavigationItemGroup(string $field_name, array $field_groups): ?\stdClass {
```

then on line 52
if (empty($field_group_item))
and on line 94
return NULL

but this is just a suggestion, mixed is ok for me, this just goes better with the strict type

modules/oe_theme_helper/templates/oe-theme-helper-in-page-navigation.html.twig
drishu: please remove, `<main>` cannot be a descendent of a `<article>`

drishu (changes requested):

drishu (dismissed):

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -61,6 +61,16 @@ function oe_theme_helper_theme($existing, $type, $theme, $path) {
...
     ],
+    'oe_theme_helper_in_page_navigation' => [
+      'render element' => 'element',
+      'template' => 'oe-theme-helper-in-page-navigation',
+      'file' => 'oe_theme_helper.theme.inc',
+    ],
+    'oe_theme_helper_in_page_navigation_item' => [
+      'render element' => 'element',
+      'template' => 'oe-theme-helper-in-page-navigation-item',
+      'file' => 'oe_theme_helper.theme.inc',
```
ademarco: We never used `'file'` in the past, let's move all those functions in the main `.module` file.

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -61,6 +61,16 @@ function oe_theme_helper_theme($existing, $type, $theme, $path) {
...
       ],
     ],
+    'oe_theme_helper_in_page_navigation' => [
+      'render element' => 'element',
+      'template' => 'oe-theme-helper-in-page-navigation',
+      'file' => 'oe_theme_helper.theme.inc',
+    ],
+    'oe_theme_helper_in_page_navigation_item' => [
+      'render element' => 'element',
+      'template' => 'oe-theme-helper-in-page-navigation-item',
```
ademarco: Do we need to specify the template name here? Isn't it derived from the theme definition array key (i.e. `oe_theme_helper_in_page_navigation_item`)?

modules/oe_theme_helper/oe_theme_helper.theme.inc
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @param array $variables
+ *   An array of variables passed to the theme hook.
+ * @param string $hook
+ *   Hook name.
+ *
+ * @return array
+ *   Theme suggestions.
+ */
+function oe_theme_helper_field_group_theme_suggestions(array $variables, string $hook): array {
```
ademarco:
```suggestion
function _oe_theme_helper_field_group_theme_suggestions(array $variables, string $hook): array {
```

Prefix this helper function with a "_", or it might be mistaken for a hook implementation.

modules/oe_theme_helper/oe_theme_helper.theme.inc
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @param array $variables
+ *   An array of variables.
+ */
+function template_preprocess_oe_theme_helper_in_page_navigation(array &$variables) {
+  $element = $variables['element'];
+
+  $variables['title'] = !empty($element['#title']) ? $element['#title'] : '';
+  $variables['children'] = !empty($element['#children']) ? $element['#children'] : '';
+  $variables['links'] = !empty($element['#links']) ? $element['#links'] : '';
```
ademarco: Shall we pass here the entity object too? It might be useful.

modules/oe_theme_helper/oe_theme_helper.theme.inc
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @param array $variables
+ *   An array of variables.
+ */
+function template_preprocess_oe_theme_helper_in_page_navigation_item(array &$variables) {
+  $element = $variables['element'];
+
+  $variables['title'] = !empty($element['#title']) ? $element['#title'] : '';
+  $variables['children'] = !empty($element['#children']) ? $element['#children'] : '';
+  $variables['title_attributes'] = !empty($element['#title_attributes']) ? $element['#title_attributes'] : '';
```
ademarco: Shall we pass here the entity object too? It might be useful.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigation.php
ademarco: Let's use [Twig lop variable](https://twig.symfony.com/doc/3.x/tags/for.html#the-loop-variable) to determine which item is first or last in a loop, so we can remove all this logic from here.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigation.php
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Returns if the field is a group or a field.
+   *
+   * @param string $field_name
+   *   The name of the field.
+   * @param array $field_groups
+   *   List of groups.
+   *
+   * @return object|null
+   *   Object if field group is found, FALSE otherwise.
```
ademarco:
```suggestion
   *   Object if field group is found, NULL otherwise.
```

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigation.php
```diff
@@ -0,0 +1,97 @@
...
+      // Generate list of links.
+      $label = $field_group_item->label;
+      $id = "inline-nav-" . Html::cleanCssIdentifier($label);
+      $links[] = [
+        'href' => "#" . $id,
+        'label' => $label,
+      ];
+
+      $element[$value]['#title_attributes'] = new Attribute();
+      $element[$value]['#title_attributes']->setAttribute('id', $id);
```
ademarco: As we have a dedicated render element for a field group item, can't we move this logic withing one of its static render callbacks?

modules/oe_theme_helper/src/Element/InPageNavigationItem.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+
+/**
+ * Provides a render element for an in-page navigation item element.
```
ademarco: Let's also document what render array parameters this render element accepts, for example:

```
'#title' => ....
...
```

modules/oe_theme_helper/src/Element/InPageNavigation.php
```diff
@@ -0,0 +1,33 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper\Element;
+
+use Drupal\Core\Render\Element\RenderElement;
+
+/**
+ * Provides a render element for an in-page navigation element.
+ *
```
ademarco: Let's also document what render array parameters this render element accepts.

ademarco (changes requested):

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigation.php
yenyasinn: `#last` is for in-page navigation item to determine its position to use it in the wrapper.
In-page navigation includes already rendered children items so we can't wrap items there.

modules/oe_theme_helper/src/Element/InPageNavigation.php
yenyasinn: It is simple render element without any special input parameters. Standard parameters are used: `#process`, `#pre_render`. Not sure that we should document it.

modules/oe_theme_helper/src/Element/InPageNavigationItem.php
yenyasinn: It is simple render element without any special input parameters. Standard parameters are used: `#process`, `#pre_render`. Not sure that we should document it.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigation.php
ademarco: Well, the fact that something is first or last should be determined at the container level right? Not on the children, so all of this should go in the containing 'oe-theme-helper-in-page-navigation.html.twig'.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/InPageNavigationTest.php
```diff
@@ -0,0 +1,258 @@
...
+    $navigation_list_items = $navigation_list->findAll('css', '.ecl-inpage-navigation__item');
+    $this->assertCount(3, $navigation_list_items);
+    for ($i = 0; $i < 3; $i++) {
+      $navigation_list_item_link = $navigation_list_items[$i]->find('css', '.ecl-link');
+      $this->assertEquals("Field group inpage_nav_item_$i", $navigation_list_item_link->getText());
+      $this->assertEquals("#field-group-inpage-nav-item-$i", $navigation_list_item_link->getAttribute('href'));
+    }
+
+    // Assert content part.
+    $wrapper = $this->assertSession()->elementExists('css', '.ecl-row.ecl-u-mt-l');
```
imanoleguskiza: This wrapper contains both the navigation and the content, we can declare it earlier and use it to assert that both navigation and content are inside it.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/InPageNavigationTest.php
```diff
@@ -0,0 +1,258 @@
...
+    $this->createFieldGroup('main', 'oe_theme_helper_in_page_navigation', $children_group_inpage_nav);
+    $this->drupalGet('node/' . $this->node->id());
+
+    // Assert navigation part.
+    $this->assertSession()->elementTextContains('css', '.ecl-inpage-navigation__title', 'Field group main');
+    $navigation_list = $this->assertSession()->elementExists('css', '.ecl-inpage-navigation__list');
+    $navigation_list_items = $navigation_list->findAll('css', '.ecl-inpage-navigation__item');
+    $this->assertCount(3, $navigation_list_items);
+    for ($i = 0; $i < 3; $i++) {
+      $navigation_list_item_link = $navigation_list_items[$i]->find('css', '.ecl-link');
```
imanoleguskiza: I think its better to use more specific selectors: a.ecl-inpage-navigation__link for example

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/InPageNavigationTest.php
```diff
@@ -0,0 +1,258 @@
...
+        $group_html_element->group_name,
+      ],
+      // It is empty extra field, so field group shouldn't be rendered.
+      [
+        'extra_field_empty_formatted_test',
+      ],
+    ];
+    $groups_inpage_nav_item = [];
+    $children_group_inpage_nav = [];
+    for ($i = 0; $i < 4; $i++) {
```
imanoleguskiza: I'm not a huge fan of this, if we add an extra navigation group to the test in the future we will need to update this. Instead we can do:    foreach ($children_group_inpage_nav_items as $index => $children_group_inpage_nav_item)

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/InPageNavigationTest.php
```diff
@@ -0,0 +1,258 @@
...
+      $this->assertEquals("Field group inpage_nav_item_$i", $navigation_list_item_link->getText());
+      $this->assertEquals("#field-group-inpage-nav-item-$i", $navigation_list_item_link->getAttribute('href'));
+    }
+
+    // Assert content part.
+    $wrapper = $this->assertSession()->elementExists('css', '.ecl-row.ecl-u-mt-l');
+    $content_items = $wrapper->findAll('css', '.ecl-col-lg-9 > div');
+    $this->assertCount(3, $content_items);
+    // Assert headers of field groups.
+    for ($i = 0; $i <= 2; $i++) {
```
imanoleguskiza: Same as before, we already have the $content_items to iterate through.

modules/oe_theme_helper/tests/src/Functional/Plugin/field_group/InPageNavigationTest.php
```diff
@@ -0,0 +1,258 @@
...
+    // Create In-page navigation group.
+    $this->createFieldGroup('main', 'oe_theme_helper_in_page_navigation', $children_group_inpage_nav);
+    $this->drupalGet('node/' . $this->node->id());
+
+    // Assert navigation part.
+    $this->assertSession()->elementTextContains('css', '.ecl-inpage-navigation__title', 'Field group main');
+    $navigation_list = $this->assertSession()->elementExists('css', '.ecl-inpage-navigation__list');
+    $navigation_list_items = $navigation_list->findAll('css', '.ecl-inpage-navigation__item');
+    $this->assertCount(3, $navigation_list_items);
+    for ($i = 0; $i < 3; $i++) {
```
imanoleguskiza: Same as before, we already have the $navigation_list_items to iterate trhough

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/InPageNavigationBase.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+abstract class InPageNavigationBase extends FieldGroupFormatterBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm() {
+    $form = parent::settingsForm();
+    // Set label required, because we use it for links and navigation label.
+    $form['label']['#required'] = TRUE;
```
imanoleguskiza, line 20: If we are only going to use the label, then we should hide the Id and extra classes fields.

imanoleguskiza (changes requested):

imanoleguskiza (approved):

ademarco (approved):

PR #545, opened 2020-08-21
D8TEWPPE-77: Fix page header test.

drishu:
### Description

Fix ecl 2.31.0 with amended patch, because raw filter was added to page header description, this caused the functional test https://github.com/openeuropa/oe_theme/blob/2.x/modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php to fails.

### Change log

- Changed:
  - container for service in page header test
  - removed war from description in page header patch


ademarco:
Closed in favor of #546

PR #551, opened 2020-08-31
EWPP-191: File teaser pattern.

22Alexandra:


tests/Kernel/fixtures/rendering.yml
```diff
@@ -1672,7 +1754,7 @@
       "div.ecl-card__image": 0
       'h1.ecl-card__title': 1
       'div.ecl-card__description': 0
-      'footer.ecl-card__footer': 0
+      'footer.ecl-card__footer': 1
```
ademarco: We shouldn't change this test, if footer is not set then we shouldn't have it displayed. Maybe this was a regression in ecl twing, in which case we need to patch the related component in order to make sure that the original version of this test passes.

ademarco (changes requested):
I've removed the teaser from the file value object, as it wasn't supposed to be part of that, and pushed my change. We still have one thing to fix.

ademarco (dismissed):

imanoleguskiza (dismissed):

imanoleguskiza (dismissed):

composer.json
```diff
@@ -39,7 +39,7 @@
         "guzzlehttp/guzzle": "~6.3",
         "nikic/php-parser": "~3.0",
         "openeuropa/behat-transformation-context": "~0.1",
-        "openeuropa/code-review": "~1.1",
+        "openeuropa/code-review": "~1.5.0",
```
nagyad: Patch number should not be there.

nagyad (dismissed):

imanoleguskiza (approved):

ademarco (approved):

PR #552, opened 2020-09-01
EWPP-55: In-page navigation field group formatter

ademarco:
This PR refactors #544 without the usage of render elements.


modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -76,3 +81,67 @@ function oe_theme_helper_field_widget_oe_paragraphs_variants_form_alter(&$elemen
...
+ *
+ * @param array $variables
+ *   An array of variables passed to the theme hook.
+ * @param string $hook
+ *   Hook name.
+ *
+ * @return array
+ *   Theme suggestions.
+ */
+function _oe_theme_helper_field_group_theme_suggestions(array $variables, string $hook): array {
```
yenyasinn: We don't need separate function anymore - functionality can be moved to oe_theme_helper_theme_suggestions_oe_theme_helper_in_page_navigation.

PR #555, opened 2020-09-02
EWPP-43: Implement teasers using extra fields.

ademarco:


modules/oe_theme_content_event/src/Plugin/ExtraField/Display/TeaserDetailsExtraField.php
```diff
@@ -0,0 +1,112 @@
...
+    $build = [
+      '#theme' => 'oe_theme_content_event_teaser_details',
+      '#items' => [],
+    ];
+
+    $cache = CacheableMetadata::createFromRenderArray($build);
+    if (!$entity->get('oe_event_venue')->isEmpty()) {
+      /** @var \Drupal\oe_content_entity_venue\Entity\VenueInterface $venue */
+      $venue = $this->entityRepository->getTranslationFromContext($entity->get('oe_event_venue')->entity);
+      $venue_access = $venue->access('view', NULL, TRUE);
```
imanoleguskiza, line 90: Are we suppose to bail not add the location if we don't have access to the venue?

modules/oe_theme_content_event/templates/oe-theme-content-event-teaser-details.html.twig
```diff
@@ -0,0 +1,24 @@
+{#
+/**
+ * @file
+ * Provides a "Icon with text"-like pattern for event teaser details.
```
imanoleguskiza: Why not use the pattern?

imanoleguskiza (changes requested):

modules/oe_theme_content_event/templates/oe-theme-content-event-teaser-details.html.twig
ademarco: We have to add an extra class `ecl-u-type-s`, which is not in the pattern, but I guess we could do that or at least reuse the icons with text composition, as it's in our code base.

ademarco: I'm reusing now the composition.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/TeaserDetailsExtraField.php
ademarco: Not really, that's there in case venue is not empty but user has no access, in this way, when access to venue change, cache is invalidated, and venue can be displayed.

ademarco: Mh but we should not print the location anyway if we don't ahve access right?

imanoleguskiza: Indeed, that was my question

modules/oe_theme_helper/tests/src/Kernel/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatterTest.php
```diff
@@ -0,0 +1,131 @@
...
+use Drupal\Tests\media\Traits\MediaTypeCreationTrait;
+use Drupal\Tests\oe_theme\Kernel\AbstractKernelTestBase;
+
+/**
+ * Tests the media thumbnail url field formatter.
+ */
+class MediaThumbnailUrlFormatterTest extends AbstractKernelTestBase {
+
+  use MediaTypeCreationTrait;
+  /**
```
ademarco, line 22: Missing new line.

modules/oe_theme_helper/tests/src/Kernel/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatterTest.php
```diff
@@ -0,0 +1,131 @@
...
+    $this->installEntitySchema('file');
+    $this->installSchema('file', 'file_usage');
+
+    $this->installConfig([
+      'file',
+      'field',
+      'entity_reference',
+      'media',
+    ]);
+
```
ademarco: Remove empty line.

modules/oe_theme_helper/tests/src/Kernel/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatterTest.php
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Test social media links formatting.
```
ademarco: Wrong comment.

modules/oe_theme_helper/tests/src/Kernel/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatterTest.php
```diff
@@ -0,0 +1,131 @@
...
+    ]);
+    $this->assertRendering($this->renderRoot($build), [
+      'contains' => [
+        // Because the media has no thumbnail configured we use the generic
+        // no thumbnail file by default.
+        'files/styles/large/public/media-icons/generic/no-thumbnail.png',
+      ],
+    ]);
+  }
+
```
ademarco, line 143: Add test case when no image style is set.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/TeaserDetailsExtraField.php
```diff
@@ -0,0 +1,114 @@
...
+          'icon' => 'location',
+          'text' => $renderable[0]['locality']['#value'] . ', ' . $renderable[0]['country']['#value'],
+        ];
+      }
+    }
+
+    if (!$entity->get('oe_event_online_type')->isEmpty()) {
+      $build['#fields']['items'][] = [
+        'icon' => 'livestreaming',
+        'text' => t('Live streaming available'),
```
upchuk: $this->t()

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/TimeStatusExtraField.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Drupal\oe_theme_content_event\Plugin\ExtraField\Display;
+
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\oe_content_event\EventNodeWrapper;
+
+/**
+ * Display whereas an event is "past", "ongoing", "future" or "cancelled".
```
upchuk: whether

modules/oe_theme_content_news/oe_theme_content_news.post_update.php
```diff
@@ -17,3 +19,30 @@ function oe_theme_content_news_post_update_00001(array &$sandbox): void {
   $news_date_format->set('pattern', 'j F Y');
   $news_date_format->save();
 }
+
+/**
+ * Enable Extra field module.
+ */
+function oe_theme_content_news_post_update_00002(): void {
+  \Drupal::service('module_installer')->install(['extra_field']);
```
upchuk: We set oe_theme_helper as new dependency so we should install that here as well

modules/oe_theme_content_publication/oe_theme_content_publication.info.yml
```diff
@@ -4,4 +4,5 @@ description: Companion module for the OE Content Publication module
 package: OpenEuropa
 core: 8.x
 dependencies:
+  - extra_field:extra_field
```
upchuk, line 7: Doesn't this require oe_theme_helper like the news one? If we are shipping with extra fields defined in display settings, we should add the dependency here and in the update path. Ensure it's consistent for all.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatter.php
```diff
@@ -0,0 +1,115 @@
...
+
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\image\Plugin\Field\FieldFormatter\ImageFormatter;
+use Drupal\Core\Field\Plugin\Field\FieldType\EntityReferenceItem;
+use Drupal\Core\Field\FieldDefinitionInterface;
+
+/**
+ * Plugin implementation of the 'media_thumbnail_url' formatter.
```
upchuk, line 16: Let's describe a bit what this formatter ends up rendering please.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatter.php
```diff
@@ -0,0 +1,115 @@
...
+
+      if ($media->get('thumbnail')->isEmpty()) {
+        // In case the thumbnail is missing from the media entity, we should
+        // apply the cache metadata of the media to the render array.
+        $cache->applyTo($elements[$delta]);
+        continue;
+      }
+
+      // Get default URL.
+      $uri = $media->get('thumbnail')->entity->getFileUri();
```
upchuk: You need to check if the entity actually exists. isEmpty checks if there is a target_id but the referenced file can be deleted === big crash here.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatter.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    $image_style_setting = $this->getSetting('image_style');
+
+    /** @var \Drupal\media\MediaInterface[] $media_items */
+    foreach ($media_items as $delta => $media) {
+      // Create a new CacheableMetadata instance as it needs to be applied
+      // to each element.
+      $cache = new CacheableMetadata();
+      $cache->addCacheableDependency($media);
```
upchuk, line 76: `$this->getEntitiesToView($items, $langcode);` checks for access and sets an `$item->_accessCacheability` onto the $items so we need to add that cache metadata as well, otherwise we lose it.

tests/PatternAssertions/IconsTextAssert.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme\PatternAssertions;
+
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Assertions for the icons with text pattern.
```
upchuk, line 10: add a `@see` for the pattern file

tests/PatternAssertions/ListItemAssert.php
```diff
@@ -0,0 +1,210 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme\PatternAssertions;
+
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Assertions for the list item pattern.
```
upchuk, line 10: add a `@see` for the pattern file

tests/PatternAssertions/PatternAssertState.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme\PatternAssertions;
+
+/**
+ * Basic pattern assertion state.
```
upchuk, line 8: Let's explain a bit what this is:

```
An assertion state object is used to contain the expected value of a rendered pattern and corresponding pattern assertion class. This is used when asserting patterns embedded in other patterns.
```

upchuk (changes requested):

upchuk (dismissed):

README.md
```diff
@@ -62,6 +62,13 @@ page.
 Note for developers: changing the component library will only load different CSS and JS assets, the actual HTML is the
 same between the two libraries.
 
+### Updating to 2.9.0+
+
+If you are using the oe_content module together with the OpenEuropa theme then updating to 2.9.0 or later will affect your
+existing teaser displays. The 2.9.0 version updates the teaser display of most content types provided by oe_content
+so if you want to keep any customization you have made to your site you will need to redo those modifications and
+override the teaser templates no your own custom theme.
```
hernani: typo

upchuk (approved):

hernani (approved):

imanoleguskiza (approved):

PR #559, opened 2020-09-07
EWPP-227: Update ECL to 2.32.0.

imanoleguskiza:
### Description

Update dependencies and patches for ecl 2.32.0.

### Change log

- Added:
- Changed: Update dependencies and patches for ecl 2.32.0.
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


tests/Kernel/fixtures/rendering.yml
```diff
@@ -515,7 +515,7 @@
       thumbnail:
         src: "https://inno-ecl.s3.amazonaws.com/media/examples/example-image.jpg"
         alt: "thumbnail alt"
-      resource_type: "resource type"
+      resource_type: "RESOURCE TYPE"
```
22Alexandra, line 518: This is supposed to be updated to upper case by the component.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -560,7 +560,8 @@
     equals:
       'div.ecl-file__detail-info div.ecl-file__title': "Example document"
       'div.ecl-file__detail-info div.ecl-file__description': 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer quis lorem tellus. Nullam sollicitudin suscipit diam, ac blandit ipsum tempor consectetur. Duis vitae pulvinar turpis. Donec maximus pharetra ex a ultricies.'
-      'div.ecl-file__detail-info div.ecl-file__detail-meta': 'RESOURCE TYPE | 2020-31-08'
+      'div.ecl-file__detail-info div.ecl-file__detail-meta span.ecl-file__detail-meta-item:nth-child(1)': 'RESOURCE TYPE'
```
22Alexandra, line 563: We could keep these 2 assertions plus the removed one since the component renders them as such.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -162,8 +162,8 @@
       'div.ecl-hero-banner__image[style="background-image:url(http://via.placeholder.com/150x150)"]': 1
       'div.ecl-page-banner--full-width': 0
     equals:
-      'div.ecl-container.ecl-hero-banner__container div.ecl-hero-banner__content h1.ecl-hero-banner__title': Banner title
-      'div.ecl-container.ecl-hero-banner__container div.ecl-hero-banner__content p.ecl-hero-banner__description': Banner description
+      'div.ecl-container div.ecl-hero-banner__container div.ecl-hero-banner__content div.ecl-hero-banner__title': Banner title
```
22Alexandra, line 165: There are new banners variants like with or without text block and with white background that should be tested.
banners variants

tests/Kernel/fixtures/rendering.yml
imanoleguskiza: If we keep the removed one then the text will include html tags (the new span.ecl-file__detail-meta-item elements), that is why I removed it.

tests/Kernel/fixtures/rendering.yml
imanoleguskiza: It's no longer the case so now it will be our responsibility to pass the formatter strings.

imanoleguskiza: Checked it with @22Alexandra and found out that, indeed the first element is made uppercase through CSS.

imanoleguskiza: However, since it is handled through CSS if we leave the string lowercase the test will fail since we test the raw rendered html.

tests/Kernel/fixtures/rendering.yml
22Alexandra: Created a follow up: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-235

22Alexandra (dismissed):

package.json
```diff
@@ -13,19 +13,19 @@
...
+        "@ecl/eu-component-form-help-block": "2.32.0",
         "@ecl/stylelint-config": "0.2.1",
-        "@ecl/ec-component-fact-figures": "2.31.0",
-        "@ecl/ec-component-media-container": "2.31.0",
-        "@ecl-twig/ec-components": "2.31.1",
-        "node-sass": "4.13.1",
+        "@ecl/ec-component-fact-figures": "2.32.0",
+        "@ecl/ec-component-media-container": "2.32.0",
+        "@ecl-twig/ec-components": "2.32.0",
+        "node-sass": "4.12.0",
```
ademarco: This needs to be back to "node-sass": "4.13.1", to suppress GitHub security warnings.

22Alexandra (approved):

ademarco (changes requested):
The breadcrumb behavior mobile is not consistent with ECL's:

Peek 2020-09-09 12-00

While we have this on our theme:

image

ademarco (approved):

22Alexandra:
> The breadcrumb behavior mobile is not consistent with ECL's:
> 
> Peek 2020-09-09 12-00
> 
> While we have this on our theme:
> 
> image

It works if the page is already opened on mobile.
breadcrumb

PR #529, opened 2020-08-03
EWPP-63: Theme "Call for tender" full view

drishu:
### Description

Theme "Call for tender" full view

### Change log

- Added: oe_theme_content_tender module

PR #563, opened 2020-09-08
EWPP-230: Theme project teaser.

ademarco:


modules/oe_theme_helper/tests/src/Kernel/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatterTestBase.php
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Create a media of type "image".
+   *
+   * @param string $filepath
+   *   Full path to image file.
+   *
+   * @return \Drupal\media\Entity\Media
+   *   Media object.
+   */
+  protected function createMediaImage(string $filepath): Media {
```
yenyasinn, line 57: We need methods like: 
* createMediaImage($name, $filepath) (https://github.com/openeuropa/oe_theme/blob/EPIC-project/tests/Functional/ContentProjectRenderTest.php#L539),
* createMediaDocument($name, $filepath) https://github.com/openeuropa/oe_theme/blob/EPIC-project/tests/Functional/ContentProjectRenderTest.php#L569,
* createOrganisationEntity($name, $bundle)  (https://github.com/openeuropa/oe_theme/blob/EPIC-project/tests/Functional/ContentProjectRenderTest.php#L331)
almost in all content type tests.

Also asserts of field formatters are re-used a lot for 
* Media Document Default view mode (https://github.com/openeuropa/oe_theme/pull/529/files#diff-07c9677f4102faefc99f738124eaa57aR85)
* Featured Media field (https://github.com/openeuropa/oe_theme/blob/EPIC-project/tests/Functional/ContentProjectRenderTest.php#L597)

I think we need some one place where all these methods will be stored. It can be base class in oe_theme/src/tests/.../ (but if we have Kernel and Functional tests both not sure it can be implemented) or trait classes in oe_theme/src/tests/ that can be included to the tests or another option.

I see that you have started with assertions based on BasePatternAssert in this PR. My initiative was https://github.com/openeuropa/oe_theme/pull/529/files#diff-2b3a320550cf248ff78b8e0da8355d64R15 that can be re-used for Event/Project/Tender content types but looks like such approach isn't correct as we will have Kernel and Functional tests both. Anyway we also need createEntity() methods to reduce code duplication and simplify developer's work

tests/Kernel/ProjectRenderTest.php
```diff
@@ -0,0 +1,191 @@
...
+    // Set current user to UID 1, so that by default we can access everything.
+    $account = User::load(1);
+    $this->setCurrentUser($account);
+  }
+
+  /**
+   * Test an event being rendered as a teaser.
+   */
+  public function testProjectTeaser(): void {
+    $file = file_save_data(file_get_contents(drupal_get_path('theme', 'oe_theme') . '/tests/fixtures/example_1.jpeg'), 'public://example_1.jpeg');
```
yenyasinn, line 80: See my comment in MediaThumbnailUrlFormatterTestBase.php - this code is duplicated.

tests/Kernel/ProjectRenderTest.php
```diff
@@ -0,0 +1,191 @@
...
+      'bundle' => 'image',
+      'name' => 'test image',
+      'oe_media_image' => [
+        'target_id' => $file->id(),
+        'alt' => 'Alt',
+      ],
+    ]);
+    $media->save();
+
+    $coordinator = Organisation::create([
```
yenyasinn, line 94: See my comment in MediaThumbnailUrlFormatterTestBase.php - this code is duplicated.

modules/oe_theme_helper/tests/src/Kernel/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatterTestBase.php
ademarco: I think the best is to organise all of those into traits, stored in in the module's codebase, so that they can be used on all types of tests, kernel, functional, behat contexts, etc. let's do this gradually, after release we will focus on a full test rewrite on oe_theme and oe_content, so that we consolidate by using the same techniques everywhere.

tests/Kernel/ProjectRenderTest.php
```diff
@@ -0,0 +1,186 @@
...
+    module_load_include('install', 'oe_content');
+    oe_content_install();
+
+    // Set current user to UID 1, so that by default we can access everything.
+    $account = User::load(1);
+    $this->setCurrentUser($account);
+  }
+
+  /**
+   * Test an event being rendered as a teaser.
```
nagyad: This is not testing event.

tests/PatternAssertions/FieldListAssert.php
```diff
@@ -0,0 +1,107 @@
...
+ * Assertions for the field list pattern.
+ *
+ * @see ./templates/patterns/field_list/field_list.ui_patterns.yml
+ */
+class FieldListAssert extends BasePatternAssert {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getAssertions($variant): array {
```
nagyad: Missing parameter type hint.

tests/PatternAssertions/FieldListAssert.php
```diff
@@ -0,0 +1,107 @@
...
+   * Asserts the items of the pattern.
+   *
+   * @param array $expected_items
+   *   The expected item values.
+   * @param string $variant
+   *   The variant of the pattern being checked.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertItems(array $expected_items, string $variant, Crawler $crawler): void {
```
nagyad: Variant parameter is not used in this function.

tests/PatternAssertions/FieldListAssert.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    // Assert all values are correct.
+    $expected_values = array_column($expected_items, 'body');
+    $value_items = $crawler->filter('dd.ecl-description-list__definition');
+    self::assertCount(count($expected_labels), $value_items);
+    foreach ($expected_values as $index => $expected_value) {
+      self::assertEquals($expected_value, trim($value_items->eq($index)->text()));
+    }
+
```
nagyad: Empty line not required.

nagyad (changes requested):

nagyad (dismissed):

imanoleguskiza (approved):

nagyad (approved):

PR #568, opened 2020-09-10
EWPP-192: Add translation toggle label variable to file patterns.

22Alexandra:


templates/patterns/file_teaser/pattern-file-teaser.html.twig
```diff
@@ -48,7 +48,7 @@
   },
   'translation': {
     'toggle': {
-      'label': "Available translations (@count)"|t({ '@count': translations|length}),
+      'label': translation_toggle_label ~ " (@count)"|t({ '@count': translations|length}),
```
ademarco: Let's simply concatenate this string, not passing via t()
```suggestion
      'label': translation_toggle_label ~ " (" ~ translations|length ~ ")",
```

Same below

ademarco (commented):
Only one comment, rest is ok, haven't tested it manually though.

templates/patterns/file_teaser/pattern-file-teaser.html.twig
```diff
@@ -48,7 +48,7 @@
   },
   'translation': {
     'toggle': {
-      'label': "Available translations (@count)"|t({ '@count': translations|length}),
+      'label': translation_toggle_label ~ " (" ~ translations|length ~ ")",
```
ademarco: In order to make this backward compatible, i.e. when people are using patterns that don't have the new field set, let's do this:
```suggestion
      'label': translation_toggle_label|default("Available translations"|t) ~ " (" ~ translations|length ~ ")",
```

templates/patterns/file_translation/pattern-file-translation.html.twig
```diff
@@ -45,7 +45,7 @@
   },
   'translation': {
     'toggle': {
-      'label': "Available translations (@count)"|t({ '@count': translations|length}),
+      'label': translation_toggle_label ~ " (" ~ translations|length ~ ")",
```
ademarco: same remark as above.

ademarco (changes requested):

imanoleguskiza (approved):

ademarco (approved):

PR #571, opened 2020-09-14
EWPP-253: Add oe_theme_content_organisation companion module.

drishu:
### Description

Add oe_theme_content_organisation companion module.

### Change log

- Added: oe_theme_content_organisation companion module.


tests/Functional/ContentOrganisationRenderTest.php
```diff
@@ -0,0 +1,209 @@
...
+      'oe_teaser' => 'The teaser text',
+      'oe_content_content_owner' => 'http://publications.europa.eu/resource/authority/corporate-body/COMMU',
+      'uid' => 0,
+      'status' => 1,
+    ]);
+    $node->save();
+    $this->drupalGet($node->toUrl());
+
+    // Assert page header - metadata.
+    $this->assertSession()->elementTextContains('css', '.ecl-page-header-core .ecl-page-header-core__meta', 'My acronym');
```
yenyasinn: PatternAssert can be created for the pattern "page-header-core", see templates/patterns/page_header/pattern-page-header.html.twig.

modules/oe_theme_content_organisation/oe_theme_content_organisation.module
```diff
@@ -0,0 +1,58 @@
...
+  // If its an image media, render it and assign it to the image variable.
+  if (!$source instanceof MediaAvPortalPhotoSource && !$source instanceof Image) {
+   return;
+  }
+
+  $thumbnail = $media->get('thumbnail')->first();
+  $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_ratio_3_2_medium');
+
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
+  $cacheability->addCacheableDependency($media);
```
yenyasinn, line 51: I would also suggest to add file entity to the cache dependency because file can be changed externally.

modules/oe_theme_content_organisation/src/Plugin/PageHeaderMetadata/OrganisationContentType.php
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getMetadata(): array {
+    $metadata = parent::getMetadata();
+    $node = $this->getNode();
+
+    if (!($node->get('oe_summary')->isEmpty())) {
```
yenyasinn: From my point of view we should implement a method like setIntroductionMetaData($field_name = 'oe_summary') in the NodeViewRoutesBase because this code is re-used in all content types.

tests/Functional/ContentOrganisationRenderTest.php
```diff
@@ -0,0 +1,209 @@
...
+    $this->assertSession()->elementTextContains('css', '.ecl-page-header-core h1.ecl-page-header-core__title', 'My node title');
+    $this->assertSession()->elementTextContains('css', '.ecl-page-header-core .ecl-page-header-core__description', 'My introduction');
+
+    // Assert navigation part.
+    $wrapper = $this->assertSession()->elementExists('css', '.ecl-row.ecl-u-mt-l');
+    $navigation = $this->assertSession()->elementExists('css', 'nav.ecl-inpage-navigation', $wrapper);
```
yenyasinn: Assert class can be created for In-page navigation template modules/oe_theme_helper/templates/oe-theme-helper-in-page-navigation.html.twig, but it isn't pattern, it is template. Anyway I think the same approach as with PatternAsserts can be implemented here.

modules/oe_theme_content_organisation/src/Plugin/PageHeaderMetadata/OrganisationContentType.php
yenyasinn: If NodeViewRoutesBase has been updated then this method should be adjusted accordingly.

tests/Functional/ContentOrganisationRenderTest.php
drishu: done

modules/oe_theme_content_organisation/oe_theme_content_organisation.module
drishu: I've tested this, does not seem to be needed, lets discuss

EDIT: we agreed to add in in case change is made to file entity through api

modules/oe_theme_content_organisation/src/Plugin/PageHeaderMetadata/OrganisationContentType.php
drishu: done

tests/Functional/ContentOrganisationRenderTest.php
```diff
@@ -0,0 +1,205 @@
...
+      ->grantPermission('view published oe_contact')
+      ->save();
+  }
+
+  /**
+   * Tests that the Organisation page renders correctly.
+   */
+  public function testOrganisationRendering(): void {
+    $file = $this->createFile(drupal_get_path('theme', 'oe_theme') . '/tests/fixtures/example_1.jpeg');
+    $media = $this->createImage(['name' => 'test image', 'alt' => 'Alt'], (int) $file->id());
```
drishu: for now, I'm doing this only here till we get feedback on this

tests/Functional/ContentOrganisationRenderTest.php
drishu: If its not a pattern I would not put it in the same location though, lets discuss

drishu: done

nagyad:
We should have consistent epic branch names across the components. Like: EPIC-Branchname. Not EPIC-brancname or Epic-Branchname

composer.json
```diff
@@ -42,10 +42,10 @@
         "openeuropa/behat-transformation-context": "~0.1",
         "openeuropa/code-review": "~1.5",
         "openeuropa/drupal-core-require-dev": "^8.7",
-        "openeuropa/oe_content": "dev-master",
+        "openeuropa/oe_content": "dev-EPIC-Organisation",
```
nagyad, line 45: This is a correct epic branch name.

modules/oe_theme_content_organisation/README.md
```diff
@@ -0,0 +1,22 @@
+# OpenEuropa Theme Organisation Content module
+
+This module is a theming companion module to the [OpenEuropa Organisation Content](https://github.com/openeuropa/oe_content/blob/EPIC-Organisation/modules/oe_content_organisation) module.
```
nagyad: The URL should not point to the epic branch.

modules/oe_theme_content_organisation/README.md
```diff
@@ -0,0 +1,22 @@
...
+It provides the logic needed to theme the Organisation content type.
+
+## Installation
+
+Make sure you have read the OpenEuropa Content Organisation module's [README.md](https://github.com/openeuropa/oe_content/blob/EPIC-Organisation/modules/oe_content_organisation/README.md)
```
nagyad: Same here.

modules/oe_theme_content_organisation/README.md
nagyad: We need to fix this in the project module as well.

modules/oe_theme_content_organisation/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+After enabling this module make sure you assign the following permissions to the anonymous user role, so visitors can
+correctly access all organisation information.
+
+- `Contact: View any published entity`
+
+## Overridden configuration
+
+Installing this module will override the default organisation content type view mode, shipped by the
+[OpenEuropa Organisation Content](https://github.com/openeuropa/oe_content/blob/EPIC-Organisation/modules/oe_content_organisation)
```
nagyad: No link to epic branch.

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
```diff
@@ -89,4 +94,53 @@ protected function getNode(): ?NodeInterface {
...
+  /**
+   * Gets introduction data from the node.
+   *
+   * @param string $field_name
+   *   Node's field name where data is taken from.
+   *
+   * @return array|null
+   *   Inline template if data exists or NULL otherwise.
+   */
+  protected function getIntroduction($field_name = 'oe_summary'): ?array {
```
nagyad: The type hint is missing for the argument.

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
```diff
@@ -89,4 +94,53 @@ protected function getNode(): ?NodeInterface {
...
+    $node = $this->getNode();
+    if (!$node->hasField($field_name) || $node->get($field_name)->isEmpty()) {
+      return NULL;
```
nagyad: We could return just an empty array Instead of null.

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
```diff
@@ -89,4 +94,53 @@ protected function getNode(): ?NodeInterface {
...
+   *   Text in the template.
+   * @param string $format
+   *   Text format.
+   * @param string $langcode
+   *   Language code.
+   *
+   * @return array
+   *   Template array.
+   */
+  protected function getInlineTemplate($text, $format, $langcode): array {
```
nagyad: Type hints for the arguments are missing.

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
nagyad: I think we should be more specific with this function name.

tests/Functional/ContentOrganisationRenderTest.php
```diff
@@ -0,0 +1,203 @@
...
+      'bundle' => 'oe_general',
+      'name' => 'General contact',
+      'oe_email' => 'example@test.com',
+      'oe_phone' => '0123456789',
+      'oe_address' => [
+        'country_code' => 'BE',
+        'locality' => 'Brussels',
+        'address_line1' => 'My address',
+        'postal_code' => '1001',
+      ],
```
nagyad, line 77: Let's fill all the fields here, Faz, Mobile, Organisation, Office, Body, Image, Press contact url. Just like on the screenshot.

tests/Functional/ContentOrganisationRenderTest.php
```diff
@@ -0,0 +1,203 @@
...
+   * Asserts field group header.
+   *
+   * @param \Behat\Mink\Element\NodeElement $element
+   *   Field group content.
+   * @param string $title
+   *   Expected title.
+   * @param string $id
+   *   Expected id.
+   */
+  protected function assertContentHeader(NodeElement $element, string $title, string $id): void {
```
nagyad, line 164: This could be abstracted from here, since field group headers are the same across content types. Up to you.

nagyad (changes requested):

modules/oe_theme_content_organisation/README.md
drishu: done

modules/oe_theme_content_organisation/README.md
drishu: done

modules/oe_theme_content_organisation/README.md
drishu: done

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
drishu: fixed

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
drishu: done

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
drishu: I refactored in the end, made it simpler

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
drishu: renamed

tests/Functional/ContentOrganisationRenderTest.php
drishu: I moved it into its own method like on project test

tests/Functional/ContentOrganisationRenderTest.php
drishu: The work on that is already started in tender epic, so I wont do it here, but we should have a follow up after both tender and this are merged to sync the tests.

tests/Functional/ContentOrganisationRenderTest.php
```diff
@@ -187,6 +159,129 @@ protected function assertContentHeader(NodeElement $element, string $title, stri
...
+   *   Entity name. Is used as a parameter for test data.
+   * @param string $bundle
+   *   Entity bundle.
+   * @param int $status
+   *   Entity status.
+   *
+   * @return \Drupal\oe_content_entity_contact\Entity\ContactInterface
+   *   Contact entity.
+   */
+  protected function createContactEntity(string $name, string $bundle, int $status): ContactInterface {
```
nagyad, line 183: Let's create a ContentCreationTrait where we move the following functions:

- getStorage()
- createMediaImage()
- createContactEntity()
- createStakeholderOrganisationEntity()

Make use of the new trait in project, and contact tests.

nagyad (changes requested):

tests/Functional/ContentOrganisationRenderTest.php
yenyasinn: getStorage() - we don't need it in trait.
createMediaImage() - already exist in oe_media
createContactEntity() - need to be created
createStakeholderOrganisationEntity() -> createOrganisationEntity and pass bundle as a parameter.

ademarco: As discussed with @drishu we will do this in a separate ticket, since we need to refactor tests anyway.

tests/Functional/ContentOrganisationRenderTest.php
```diff
@@ -0,0 +1,320 @@
...
+    $file = $this->createFile(drupal_get_path('theme', 'oe_theme') . '/tests/fixtures/placeholder.png');
+    $media = $this->createMediaImage($file, [
```
ademarco: Same holds for this then. We will have a dedicated ticket to refactor and better organize tests. Let's roll back to the previous, yet verbose, method.

modules/oe_theme_content_organisation/oe_theme_content_organisation.install
```diff
@@ -0,0 +1,37 @@
...
+  $config_manager = \Drupal::service('config.manager');
+  $entity_type = $config_manager->getEntityTypeIdByName($config_name);
+  $config_entity_storage = $config_manager
+    ->getEntityTypeManager()
+    ->getStorage($entity_type);
+  $config_entity = $config_entity_storage->load($config_values['id']);
+
+  // Update configuration.
+  $updated_config_entity = $config_entity_storage->updateFromStorageRecord($config_entity, $config_values);
+  $updated_config_entity->save();
```
ademarco: As we will need to provide a teaser for this content type too, let's use the same method we are using on all other content types, like on the event content type case below:

```php
function oe_theme_content_event_install() {
  // If we are installing from config, we bail out.
  if (\Drupal::isConfigSyncing() === TRUE) {
    return;
  }

  // Override entity view displays.
  $storage = new FileStorage(drupal_get_path('module', 'oe_theme_content_event') . '/config/overrides');
  $displays = [
    'core.entity_view_display.node.oe_event.default',
    'core.entity_view_display.node.oe_event.teaser',
  ];

  foreach ($displays as $display) {
    $values = $storage->read($display);
    $config = EntityViewDisplay::load($values['id']);
    if (!$config) {
      $config = EntityViewDisplay::create($values);
      $config->save();
      continue;
    }

    foreach ($values as $key => $value) {
      $config->set($key, $value);
    }

    $config->save();
  }
}
```

modules/oe_theme_content_organisation/oe_theme_content_organisation.module
```diff
@@ -0,0 +1,65 @@
...
+    $cacheability->applyTo($variables);
+    return;
+  }
+
+  $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_ratio_3_2_medium');
+
+  $cacheability->addCacheableDependency($thumbnail->entity);
+  $cacheability->addCacheableDependency(\Drupal::entityTypeManager()->getStorage('media_type')->load($media->bundle()));
+  $cacheability->addCacheableDependency(EntityViewDisplay::collectRenderDisplay($media, 'oe_theme_ratio_3_2_medium'));
+  $cacheability->applyTo($variables);
```
ademarco, line 64: I think it would be useful to create an image formatter that would format an image using a media container, given its image style settings.

@drishu I'll create a follow up, so then we can refactor this using that formatter, at which point here we can just do:

```php
    $variables['image'] = $view_builder->viewField($entity->get('field_test'), [
      'type' => 'oe_theme_helper_media_container_formatter',
      'label' => 'hidden',
      'settings' => [
        'image_style' => 'oe_theme_ratio_3_2_medium',
      ],
    ]);
```
And in the template:

```twig
{% block navigation %}
<div class="ecl-col-12 ecl-col-lg-3">
  {% if image %}
    {{ image|field_value }}
  {% endif %}
  {% include '@ecl-twig/inpage-navigation' with {
    'title': title,
    'links': _items,
    'icon_path': ecl_icon_path
  } %}
</div>
{% endblock %}

```
```

File: modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php [hunk 19]
```diff
@@ -89,4 +94,36 @@ protected function getNode(): ?NodeInterface {
...
+  /**
+   * Gets introduction data from the node.
+   *
+   * @param string $field_name
+   *   Node's field name where data is taken from.
+   *
+   * @return array
+   *   Inline template if data exists or NULL otherwise.
+   */
+  protected function getPageHeaderIntroduction(string $field_name = 'oe_summary'): array {
```
ademarco:
```suggestion
  protected function getIntroductionMetadata(string $field_name): array {
```
I propose these two changes:
- Not including PageHeader in the name as it's already in the class name
- removing  this default parameter value, it will be clearer if the extending class explicitly request the field it wants to render.

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
```diff
@@ -66,6 +66,11 @@ public function getMetadata(): array {
 
     $metadata = [];
 
+    $introduction = $this->getPageHeaderIntroduction();
+    if ($introduction) {
```
ademarco: We can remove this `if`, when empty the template won't render anything.

ademarco (changes requested):

ademarco (changes requested):
Can we wrap the image with a div with class `ecl-u-mb-l` so it gets this look?

image

modules/oe_theme_content_organisation/src/Plugin/PageHeaderMetadata/OrganisationContentType.php
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function getMetadata(): array {
+    $metadata = parent::getMetadata();
+    $node = $this->getNode();
+
+    if (!$node->get('oe_organisation_acronym')->isEmpty()) {
+      $metadata['metas'] = [
+        $node->get('oe_organisation_acronym')->value,
```
ademarco, line 81: We need to get translation from context here, as the field can be set as translatable. Check other page header metadata implementations.

ademarco (changes requested):

modules/oe_theme_content_organisation/src/Plugin/PageHeaderMetadata/OrganisationContentType.php
drishu: done, but I don't think its needed

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
drishu: done

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
drishu: so I'm sending the parameter
`$metadata = [
      'introduction' => $this->getIntroductionMetadata('oe_summary'),
    ];`

modules/oe_theme_content_organisation/oe_theme_content_organisation.install
drishu: done

tests/Functional/ContentOrganisationRenderTest.php
drishu: done

modules/oe_theme_content_organisation/src/Plugin/PageHeaderMetadata/OrganisationContentType.php
ademarco: You are right, sorry, my bad! Somehow I was considering that as a reference field, but it's just a simple field.

ademarco (approved):

nagyad (approved):

PR #581, opened 2020-09-17
EWPP-278: Update ECL components to 2.33.

imanoleguskiza:
### Description

Update ECL components to 2.33.

### Change log

- Added:
- Changed: Update ECL components to 2.33.
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


package.json
```diff
@@ -14,20 +14,20 @@
   },
   "dependencies": {
     "@ecl-twig/ec-components": "2.32.0",
```
ademarco: We need to upgrade ecl-twig too.

ademarco (changes requested):

package.json
ademarco: We will actually update ECL only now and open a follow up for ecl twig. We spoke with ECL maintainer and the changes in ECL are actually only on CSS and JS and they block an important ticket (search widget).

@imanoleguskiza will open a follow up for ecl twig.

ademarco (dismissed):

22Alexandra (approved):

ademarco (approved):

PR #583, opened 2020-09-20
EWPP-272: Add News type field to News metadata.

22Alexandra:


modules/oe_theme_content_news/config/overrides/core.entity_view_display.node.oe_news.teaser.yml
```diff
@@ -31,15 +33,23 @@ content:
...
     settings:
       image_style: oe_theme_ratio_3_2_medium
     third_party_settings: {  }
+  oe_news_types:
+    type: skos_concept_entity_reference_label
+    weight: 1
+    region: content
+    label: hidden
+    settings:
+      link: true
```
ademarco: Set this to false

modules/oe_theme_content_news/config/post_updates/00004_update_teaser_view_display/core.entity_view_display.node.oe_news.teaser.yml
```diff
@@ -0,0 +1,78 @@
...
+    settings:
+      image_style: oe_theme_ratio_3_2_medium
+    third_party_settings: {  }
+  oe_news_types:
+    type: skos_concept_entity_reference_label
+    weight: 1
+    region: content
+    label: hidden
+    settings:
+      link: true
```
ademarco: Set this to false

modules/oe_theme_content_news/oe_theme_content_news.module
```diff
@@ -7,25 +7,39 @@
...
+  $news_type = $node->get('oe_news_types')->entity;
+  if ($news_type instanceof ConceptInterface) {
+    $cacheability = new CacheableMetadata();
+    $cacheability->addCacheableDependency($news_type);
+    $news_type = \Drupal::service('entity.repository')
+      ->getTranslationFromContext($news_type);
+    $variables['content']['news_types'] = $news_type->label();
   }
+  $timestamp = $node->get('oe_publication_date')->date->getTimestamp();
+  $variables['content']['date'] = \Drupal::service('date.formatter')->format($timestamp, 'oe_theme_news_date');
```
ademarco: Restore this as it was before, we can simply use drupal fields now, we export the teaser view mode for this reason.

modules/oe_theme_content_news/src/Plugin/PageHeaderMetadata/NewsContentType.php
```diff
@@ -98,9 +100,16 @@ public function getMetadata(): array {
...
+    $news_type = $node->get('oe_news_types')->entity;
+    if ($news_type instanceof ConceptInterface) {
+      $cacheability = new CacheableMetadata();
+      $cacheability->addCacheableDependency($news_type);
+      $news_type = \Drupal::service('entity.repository')->getTranslationFromContext($news_type);
+      $news_type_label = $news_type->label();
+    }
     $timestamp = $node->get('oe_publication_date')->date->getTimestamp();
     $metadata['metas'] = [
-      $this->t('News'),
```
ademarco, line 103: Is this not required anymore? Let's maybe keep it as a fallback in case label is empty, which it will be the case for already existing content.

templates/content/node--oe-news--teaser.html.twig
```diff
@@ -10,7 +10,8 @@
   'url': url,
   'title': content.extra_field_oe_theme_helper_short_title_with_fallback|field_value,
   'meta': [
-    content.oe_publication_date|field_value
+    content.news_types|upper,
+    content.date|upper
```
ademarco: Remove `|upper`, we will fix this in CSS later on. Also here use the actual fields.

templates/content/node--oe-news--teaser.html.twig
```diff
@@ -10,7 +10,8 @@
   'url': url,
   'title': content.extra_field_oe_theme_helper_short_title_with_fallback|field_value,
   'meta': [
-    content.oe_publication_date|field_value
```
ademarco, line 13: Why you removed this? This was ok, we need to move to use drupal fields now that we export the teaser view mode.

ademarco (changes requested):

tests/features/news.feature
```diff
@@ -7,14 +7,15 @@ Feature: News content type.
   Scenario: News information is shown in teasers.
```
ademarco: I think it's time to move this behat into a kernel test. We are slowly doing that for all content type. Steps to do this are:

- Create a NewsRenderTest class, check ProjectRenderTest.php for reference 
- Move news testNews() method from here \Drupal\Tests\oe_theme\Kernel\LegacyContentRenderTest into that class and call it testFullView().
- Create a new method to test teaser in NewsRenderTest and use the new pattern assertions to do that, as we did in ProjectRenderTest, check with @imanoleguskiza if you need help
- Test all cases, such as if there is no label that it falls back to news, etc.
- Remove this behat scenario

ademarco (changes requested):

tests/features/news.feature
yenyasinn: @ademarco does it mean that I have to remove functional test ContentNewsRenderTest for full News page https://github.com/openeuropa/oe_theme/pull/584/files#diff-01cf51fb99613a3614ce350a61c37c8aR16 ?

ademarco: yeah here we have a bit of legacy to deal with. @yenyasinn I'd say keep it as you are doing it for now, then depending from which one we merge first we remove the kernel test for the full page.

modules/oe_theme_content_news/src/Plugin/PageHeaderMetadata/NewsContentType.php
```diff
@@ -98,9 +100,16 @@ public function getMetadata(): array {
       ];
     }
 
+    $news_type = $node->get('oe_news_types')->entity;
+    if ($news_type instanceof ConceptInterface) {
+      $cacheability = new CacheableMetadata();
+      $cacheability->addCacheableDependency($news_type);
+      $news_type = \Drupal::service('entity.repository')->getTranslationFromContext($news_type);
```
ademarco: This service needs to be injected, and we need to print all values, not only the first one, as the field is multiple. We should also test that, on both teaser and full page.

ademarco (changes requested):

ademarco (changes requested):
I've added logic to print multiple types on the teaser, needs to be covered by a test. Also check my other remark.

ademarco (changes requested):
We also need to make these fields visible on the page header, in case of multiple fields print all of their values.

image

ademarco:
After talking to @kugta  we will use commas to separate same fields values on metas and pipes to separate different fields

tests/Kernel/NewsRenderTest.php
```diff
@@ -0,0 +1,204 @@
...
+    $this->nodeViewBuilder->resetCache();
+
+    $build = $this->nodeViewBuilder->view($node, 'teaser');
+    $html = $this->renderRoot($build);
+
+    $expected_values = [
+      'title' => 'Test news node',
+      'url' => '/node/1',
+      'detail' => 'Teaser',
+      'meta' => "Press release\n | 2 April 2019",
```
imanoleguskiza: There shouldn't be any break lines between the meta items.

tests/Kernel/NewsRenderTest.php
```diff
@@ -0,0 +1,204 @@
...
+    $assert->assertVariant('thumbnail_primary', $html);
+
+    // Set news type.
+    $node->set('oe_news_types', 'http://publications.europa.eu/resource/authority/resource-type/PRESS_REL')->save();
+    $this->nodeViewBuilder->resetCache();
+
+    $build = $this->nodeViewBuilder->view($node, 'teaser');
+    $html = $this->renderRoot($build);
+
+    $expected_values = [
```
imanoleguskiza: When setting the expected values, the only one that changes is "meta", so you can just do     $expected_values['meta'] = "Press release\n | 2 April 2019";

templates/content/node--oe-news--teaser.html.twig
```diff
@@ -4,13 +4,16 @@
...
+{# Default to "News" if type field is empty. #}
+{# This can happen since the field was introduced at a later stage. #}
 {% set image_url = content.oe_news_featured_media|field_value|render %}
 {{ pattern('list_item', {
   'variant': 'thumbnail_primary',
   'url': url,
   'title': content.extra_field_oe_theme_helper_short_title_with_fallback|field_value,
   'meta': [
-    content.oe_publication_date|field_value
+    not node.oe_news_types.isEmpty() ? content.oe_news_types : 'News'|t,
```
imanoleguskiza: We don't need the default value since the field already has one.

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

ademarco (approved):

imanoleguskiza (approved):

PR #590, opened 2020-09-24
EWPP-49: Theming the list pages.

upchuk:


templates/content/node--oe-list-page--full.html.twig
```diff
@@ -0,0 +1,55 @@
...
+
+  {{ title_prefix }}
+  {% if not page %}
+    <h2{{ title_attributes }}>
+      <a href="{{ url }}" rel="bookmark">{{ label }}</a>
+    </h2>
+  {% endif %}
+  {{ title_suffix }}
+
+  {% if display_submitted %}
```
imanoleguskiza: Not sure we need this chunk since we always set "display_submitted" to false anyways. We could probably just remove it.

modules/oe_theme_list_pages/oe_theme_list_pages.info.yml
```diff
@@ -0,0 +1,9 @@
+name: OpenEuropa Theme List Pages
```
imanoleguskiza, line 1: Usually the name of the module is "oe_theme_content_list_pages".

imanoleguskiza (changes requested):

modules/oe_theme_list_pages/oe_theme_list_pages.info.yml
upchuk: If it's from oe_content...this is not from oe_content..

imanoleguskiza (approved):

hernani (approved):

PR #596, opened 2020-09-25
EWPP-309: Fix project styling issues.

22Alexandra:


templates/content/field-group-html-element--node--oe-project--group-result.html.twig
```diff
@@ -11,6 +11,9 @@
     {% if title %}
       <h2 class="ecl-u-type-heading-2">{{ title }}</h2>
     {% endif %}
-    {{ children }}
+    {{ element.oe_project_results }}
   </div>
 </div>
+<div id="project-result-files">
+  {{ element.oe_project_result_files }}
+</div>
```
ademarco, line 19: Try this:

```twig
<div id="project-results">
  <div class="ecl-row">
    <div class="ecl-col-6">
      {% if title %}
        <h2 class="ecl-u-type-heading-2">{{ title }}</h2>
      {% endif %}
      {{ element.oe_project_results }}
    </div>
  </div>
  {{ element.oe_project_result_files }}
</div>
```

It will make it easier to port tests both here and on EWCMS.

sass/components/_content_project.scss
```diff
@@ -10,3 +10,6 @@
     margin-left: 0;
   }
 }
+div#project-details div.ecl-col-12.ecl-col-md-6 div + div.ecl-editor {
+  margin-top: $ecl-spacing-2-xs;
+}
```
ademarco: Instead of doing this can't we add margin bottom classes the each of this field templates?

- featured media
- body

image

ademarco (changes requested):

ademarco (dismissed):

ademarco (approved):

sergepavle (approved):

PR #602, opened 2020-09-29
EWPP-323: Update logo in organisation content type

drishu:
### Description

This fixes display of larger landscape and transparent logos.

### Change log

- Changed:
  - oe_theme_medium_no_crop  style for logo
  - transparent background for logo
- Fixed: default in-page naviagation


modules/oe_theme_content_organisation/templates/oe-theme-helper-in-page-navigation--node--oe-organisation.html.twig
```diff
@@ -12,7 +12,7 @@
 {% block navigation %}
 <div class="ecl-col-12 ecl-col-lg-3">
   {% if image %}
-  <div class="ecl-u-mb-l">
+  <div class="ecl-u-mb-l figure-transparent">
```
ademarco: Instead of doing this try to pass the property `extra_classes` (string) to the `@ecl-twig/media-container` component and add an ECL utility to either make the background transparent or, if non available, white.

sass/components/_content_organisation.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * @file
+ * This file is used to style the Organisation content.
+ */
+
+.figure-transparent {
+  figure.ecl-media-container {
+    background-color: transparent !important;
+  }
+}
```
ademarco: See above, let's try to remove this if possilbe

tests/Functional/ContentOrganisationRenderTest.php
```diff
@@ -30,6 +31,8 @@ class ContentOrganisationRenderTest extends BrowserTestBase {
     'node',
     'address',
     'oe_media',
+    'media_avportal',
+    'oe_media_avportal',
```
ademarco: One or both of these two modules should be removed from this test and listed as dependencies of the oe_theme_content_org module. Also update the readme.

tests/Kernel/OrganisationRenderTest.php
```diff
@@ -33,6 +33,7 @@ class OrganisationRenderTest extends ContentRenderTestBase {
     'link',
     'options',
     'image',
+    'media_avportal',
```
ademarco: Same here as above.

ademarco (changes requested):

modules/oe_theme_content_organisation/templates/oe-theme-helper-in-page-navigation--node--oe-organisation.html.twig
drishu: white background was the only one available

sass/components/_content_organisation.scss
drishu: removed

tests/Functional/ContentOrganisationRenderTest.php
drishu: dependency added in oe_content https://github.com/openeuropa/oe_content/blob/EWPP-323/modules/oe_content_organisation/oe_content_organisation.info.yml#L14

tests/Kernel/OrganisationRenderTest.php
drishu: done

modules/oe_theme_content_organisation/oe_theme_content_organisation.module
```diff
@@ -56,10 +56,10 @@ function oe_theme_content_organisation_preprocess_oe_theme_helper_in_page_naviga
     return;
   }
 
-  $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_ratio_3_2_medium');
+  $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_medium_no_crop');
 
   $cacheability->addCacheableDependency($thumbnail->entity);
   $cacheability->addCacheableDependency(\Drupal::entityTypeManager()->getStorage('media_type')->load($media->bundle()));
-  $cacheability->addCacheableDependency(EntityViewDisplay::collectRenderDisplay($media, 'oe_theme_ratio_3_2_medium'));
+  $cacheability->addCacheableDependency(EntityViewDisplay::collectRenderDisplay($media, 'oe_theme_medium_no_crop'));
```
upchuk: This line doesn't make much sense. First of all, there is no media view mode called `oe_theme_medium_no_crop`. Nor was there one called `oe_theme_ratio_3_2_medium` so was wrong from before. This in fact will return the default entity view display of the media entity.

Second of all, why do we need to add this cache dependency? You are not using anything from the media entity display in building this image here.

upchuk (changes requested):

upchuk (approved):

ademarco (approved):

PR #606, opened 2020-09-30
EWPP-328: Theme date and datepicker using ecl.

drishu:
### Description

Theme date and datepicker using ecl.

### Change log

- Added:
   - templates for date & time
   - theme suggestion for time input
   - datepicker ecl component

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -442,7 +442,7 @@ function oe_theme_preprocess_input__checkbox(array &$variables): void {
  * Implements template_preprocess_input__date().
  */
 function oe_theme_preprocess_input__date(array &$variables): void {
-  $php_date_format = $variables['#attributes']['data-drupal-date-format'] ?? 'Y-m-d';
+  $php_date_format = !empty($variables['attributes']['data-drupal-date-format'] && is_array($variables['attributes']['data-drupal-date-format'])) ? reset($variables['attributes']['data-drupal-date-format']) : 'Y-m-d';
   $format_map = [
     // @todo find a cleverer way of mapping PHP date formats to the expected
     // ECL ones.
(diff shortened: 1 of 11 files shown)


oe_theme.theme
```diff
@@ -425,6 +438,22 @@ function oe_theme_preprocess_input__checkbox(array &$variables): void {
   _oe_theme_preprocess_input_label_wrapper($variables);
 }
 
+/**
+ * Implements template_preprocess_input__date().
+ */
+function oe_theme_preprocess_input__date(array &$variables): void {
+  $php_date_format = $variables['#attributes']['data-drupal-date-format'] ?? 'Y-m-d';
```
imanoleguskiza: There is no key #attributes, just attributes, and the value of data-drupal-date-format seems to be an array, so we always end up using the default Y-m-d

templates/form/input--date.html.twig
```diff
@@ -0,0 +1,24 @@
...
+#}
+{% include '@ecl-twig/ec-component-datepicker/ecl-datepicker.html.twig' with {
+  id: attributes.id,
+  name: attributes.name,
+  default_value: attributes.value,
+  required: element['#required'],
+  invalid: element['#errors'] is not empty,
+  disabled: attributes.disabled is defined,
+  icon_path: ecl_icon_path,
+  placeholder: 'DD-MM-YYYY'
```
imanoleguskiza: This value should be the same as the value we send to the datepicker js.

imanoleguskiza (changes requested):

templates/form/input--date.html.twig
upchuk: @imanoleguskiza preprocess this please so it takes it also dynamically like the component for the format.

imanoleguskiza (dismissed):

22Alexandra (approved):

imanoleguskiza (approved):

PR #614, opened 2020-10-04
EWPP-366: Add spacing between embedded media and embedded document in wysiwyg.

22Alexandra:


sass/components/_wysiwyg.scss
```diff
@@ -5,4 +5,7 @@
   .ecl-file + article {
     margin-top: $ecl-spacing-2-xs;
   }
+  article + .ecl-file {
+    margin-top: $ecl-spacing-2-xs;
+  }
```
ademarco, line 6: Can we try with this below and see how it looks on all combinations?
```scss
.ecl-editor {
  article,
  .ecl-file {
    margin-top: $ecl-spacing-l;
    margin-bottom: $ecl-spacing-l;
  }
}
```

ademarco (changes requested):

sass/components/_wysiwyg.scss
22Alexandra: The spacing works fine like this but previously when we fixed spacing in wysiwyg we used $ecl-spacing-2-xs as agreed with Tamas. I think $ecl-spacing-l is too much since it would be in one field and not between fields. It looks like this with $ecl-spacing-l:
spacing

Which one should I keep?

22Alexandra: I'm using $ecl-spacing-2-xs for now.

ademarco: Ok sure, if agreed with @kugta  is ok for course!

ademarco (approved):

imanoleguskiza (approved):

PR #616, opened 2020-10-05
EWPP-337: Update pattern list item image background styles.

22Alexandra:


nagyad (dismissed):

imanoleguskiza (dismissed):

ademarco (changes requested):
Not sure this will work guys, we get a space here if image format is not complying with the container ratio. Maybe we should also center the backgound?

image

ademarco (approved):
This now looks like that, approving.

image

upchuk (approved):

PR #621, opened 2020-10-09
EWPP-384: Add margin to blocks if they are part of the content region.

imanoleguskiza:
### Description

Add margin to blocks if they are part of the content region.

### Change log

- Added: Add margin to blocks if they are part of the content region.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -1234,16 +1234,20 @@ function oe_theme_preprocess_responsive_image_formatter(&$variables) {
 }
 
 /**
- * Implements template_preprocess_preprocess_block().
- *
- * For preprocessing default mode of document media type.
+ * Implements template_preprocess_block().
  */
 function oe_theme_preprocess_block(&$variables) {
+  // Style block titles using ECL.
   $variables['title_attributes'] = array_merge_recursive($variables['title_attributes'], [
     'class' => [
       'ecl-u-type-heading-2',
     ],
   ]);
+  // If the block is in the content region add a margin on top.
+  $block = \Drupal::entityTypeManager()->getStorage('block')->load($variables['elements']['#id']);
+  if ($block->getRegion() == 'content') {
+    $variables['attributes']['class'][] = 'ecl-u-mt-s';
+  }
 }
 
 /**
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -1243,11 +1243,6 @@ function oe_theme_preprocess_block(&$variables) {
       'ecl-u-type-heading-2',
     ],
   ]);
-  // If the block is in the content region add a margin on top.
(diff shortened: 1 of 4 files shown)


oe_theme.theme
```diff
@@ -1632,3 +1631,14 @@ function oe_theme_preprocess_field__media__oe_media_iframe__video_iframe__oe_the
...
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ *
+ * Add block region as a suggestion.
+ */
+function oe_theme_theme_suggestions_block_alter(array &$suggestions, array $variables) {
+  /** @var \Drupal\block\Entity\Block $block */
+  $block = \Drupal::entityTypeManager()->getStorage('block')->load($variables['elements']['#id']);
+  $suggestions[] = 'block__' . $block->getRegion();
```
ademarco: Let's prefix this with `region_` like 
```
   $suggestions[] = 'block__region_' . $block->getRegion();
```

oe_theme.theme
```diff
@@ -1632,3 +1631,14 @@ function oe_theme_preprocess_field__media__oe_media_iframe__video_iframe__oe_the
...
 }
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ *
+ * Add block region as a suggestion.
+ */
+function oe_theme_theme_suggestions_block_alter(array &$suggestions, array $variables) {
+  /** @var \Drupal\block\Entity\Block $block */
+  $block = \Drupal::entityTypeManager()->getStorage('block')->load($variables['elements']['#id']);
```
ademarco, line 1642: Don't we have the block object in $variables already?

templates/blocks/block--content.html.twig
```diff
@@ -0,0 +1,18 @@
...
+#}
+<div{{ attributes.addClass('ecl-u-mt-l').addClass('ecl-u-mb-l') }}>
+  {{ title_prefix }}
+  {% if label %}
+    <h2{{ title_attributes }}>{{ label }}</h2>
+  {% endif %}
+  {{ title_suffix }}
+  {% block content %}
+    {{ content }}
+  {% endblock %}
```
ademarco: You can extend and use the {{ parent() }} twig directive. Look for parent() in our templates, we did that somewhere else.

ademarco (changes requested):

oe_theme.theme
imanoleguskiza: Sadly no, the configuration of the block is there but the region information is nowhere to be found.

nagyad (approved):

ademarco (approved):

PR #623, opened 2020-10-09
EWPP-394: Project content type has wrong styling

enriquelacoma:
### Description

Include missing components/description_list in the sass style files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "openeuropa/behat-transformation-context": "~0.1",
         "openeuropa/code-review": "~1.5",
         "openeuropa/drupal-core-require-dev": "^8.7",
-        "openeuropa/oe_content": "~1.11",
+        "openeuropa/oe_content": "dev-EWPP-0000",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -43,7 +43,7 @@
         "openeuropa/behat-transformation-context": "~0.1",
         "openeuropa/code-review": "~1.5",
         "openeuropa/drupal-core-require-dev": "^8.7",
-        "openeuropa/oe_content": "~1.11",
+        "openeuropa/oe_content": "dev-EWPP-0000",
```
nagyad: This can be changed to 1.11 and add composer patch to master.

ademarco: mh no let's use dev-master or else it will apply the patch on ewcms.

nagyad: ok

enriquelacoma: Updated

sergepavle (approved):

ademarco (approved):

PR #629, opened 2020-10-15
EWPP-402: Rework theme content type tests to use pattern assertion and drop behat

yenyasinn:
### Description

Test refactoring.


composer.json
```diff
@@ -102,6 +102,9 @@
             },
             "drupal/address": {
                 "https://www.drupal.org/project/address/issues/3144823": "https://www.drupal.org/files/issues/2020-06-02/address_content_language-3144823-3.patch"
+            },
+            "symfony/dom-crawler": {
+                "Fix from version 4.4 https://github.com/symfony/symfony/pull/32440/files": "./patches/symfony/dom-crawler/trim_new_line_characters.patch"
```
yenyasinn: This is a code from the symfony/dom-crawler v.4.4. It's quite hard update package to that version because dependency "symfony/dom-crawler": "~3.4" exists in lots of projects including drupal/core-dev https://github.com/drupal/core-dev/blob/8.9.x/composer.json. Drupal 9 has dependency "symfony/dom-crawler": "^4.4 !=4.4.5"  https://github.com/drupal/core-dev/blob/9.0.x/composer.json. So we need this patch until migration to Drupal 9.
Why do we need this patch for? This patch removes new line characters that exist in lots of templates. Without it I used strings for asserts `"09199 Ages Burgos, Spain\n\n  Munich, Germany"`. With patch I can define trivial `'09199 Ages Burgos, Spain Munich, Germany'`. See https://github.com/openeuropa/oe_theme/pull/629/commits/74bcc10208e4de5fa36fdd3654f651eb8f9e7679.

symfony/dom-crawler v.4.4. already sanitise it, so we will need to rewrite test asserts in any case.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/RegistrationButtonExtraField.php
```diff
@@ -101,6 +101,9 @@ public function viewElements(ContentEntityInterface $entity) {
       $datetime_start = $event->getRegistrationStartDate();
       $datetime_end = $event->getRegistrationEndDate();
       $request_datetime = DrupalDateTime::createFromTimestamp($this->requestTime);
+      // Field value is stored in UTC timezone. So set proper timezone for the
```
yenyasinn: It fixes a bug (found during writing test) when user is in Australia timezone, for example, and start or end date is today.

composer.json
ademarco: Good catch guys, only I'm a bit uneasy to ship with a patch, as this might break builds down the line (patches from components are applied). Let's do like this:

- We keep this aside from now, and we ship with `\n` for the test
- We open an issue where we track these findings, we can create a Drupal 9 epic already (if not already there), so we fix this when we pass to Drupal 9

ademarco (changes requested):

composer.json
yenyasinn: @ademarco  Without this patch we also sometimes need to use different expected values in different PHP versions - https://github.com/openeuropa/oe_theme/pull/634/files#diff-45e52785ef57fd3b53dcc3e8fbb0c31c7ec218e12361a503ef37acca4b9e7bcdR240.
Explanation is here https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-437?focusedCommentId=4402494&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4402494

tests/features/page-header.feature
```diff
@@ -56,17 +56,6 @@ Feature: Page header block component.
     # The default text format should be applied, converting URLs into links.
     And I should see the link "http://www.example.org" in the "page header intro"
 
-  Scenario: News content type has custom metadata shown in the page header.
```
yenyasinn, line 59: It was moved to ContentNewsRenderTest.

tests/features/event.feature
```diff
@@ -1,375 +0,0 @@
-@api @event @datetime_testing
```
yenyasinn, line 1: It was moved to ContentEventRenderTest.

templates/field/field--node--oe-social-media-links--oe-event--full.html.twig
```diff
@@ -6,8 +6,4 @@
  * @see ./core/themes/stable/templates/field/field.html.twig
  */
 #}
-<div class="ecl-u-mt-l">
```
yenyasinn, line 9: Class `ecl-u-mt-l` exists in modules/oe_theme_helper/templates/oe-theme-helper-social-media-links.html.twig

tests/Kernel/ContactRenderTest.php
```diff
@@ -0,0 +1,245 @@
+<?php
```
yenyasinn: Since we need to test entity view step by step I decided to implement Contact full view in the separate test.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,510 @@ public function testEventRegistrationDateDescription(): void {
...
+    ];
+    $field_list_assert->assertPattern($field_list_expected_values, $practical_list_content->getOuterHtml());
+
+    $icons_text_expected_values['items'][3] = [
+      'icon' => 'livestreaming',
+      'text' => 'Live streaming available',
+    ];
+    $icons_text_assert->assertPattern($icons_text_expected_values, $details_list_content->getOuterHtml());
+
+    // Assert changing type of "Online type".
```
sergepavle, line 350: Why we asserting change of exactly 'Online type'.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,510 @@ public function testEventRegistrationDateDescription(): void {
...
+   *   Parent element.
    */
-  protected function setTimezone(string $timezone): void {
-    \Drupal::configFactory()->getEditable('system.date')
-      ->set('timezone.default', $timezone)
-      ->save();
+  protected function assertRegisterButtonEnabled(NodeElement $parent_element): void {
+    $rendered_button = $this->assertSession()->elementExists('css', 'a.ecl-u-mt-2xl.ecl-link.ecl-link--cta', $parent_element);
+    $this->assertEquals('Register here', $rendered_button->getText());
+    $this->assertEquals('http://www.example.com/registation', $rendered_button->getAttribute('href'));
```
sergepavle: hardcoded url.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,510 @@ public function testEventRegistrationDateDescription(): void {
...
+        ], [
+          'label' => 'When',
+          'body' => "Wednesday 19 February 2020, 01:00\n to Friday 28 February 2020, 01:00",
+        ], [
+          'label' => 'Languages',
+          'body' => 'Estonian, French',
+        ],
+      ],
+    ];
+    if (version_compare(PHP_VERSION, '7.3') < 0) {
```
sergepavle: As I understand that change based on that decision: https://github.com/openeuropa/oe_theme/pull/629#discussion_r531058337
Should we add here a comment for describing the problem and a link to a follow-up ticket for better tracking?

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -157,35 +149,47 @@ public function testProjectRendering(): void {
...
+        ], [
+          'label' => 'Project duration',
+          'body' => "10.05.2020\n - 15.05.2025",
+        ], [
+          'label' => 'Project locations',
+          'body' => "09199 Ages Burgos, Spain\n\n  Munich, Germany",
+        ],
+      ],
+    ];
+    if (version_compare(PHP_VERSION, '7.3') < 0) {
```
sergepavle: As I understand that change based on that decision: https://github.com/openeuropa/oe_theme/pull/629#discussion_r531058337
Should we add here a comment for describing the problem and a link to a follow-up ticket for better tracking?

tests/Kernel/ContactRenderTest.php
```diff
@@ -0,0 +1,245 @@
...
+    $field_list_assert->assertPattern($expected_values, $this->renderRoot($build));
+
+    // Add and assert Email field.
+    $contact->set('oe_email', "$name@example.com")->save();
+    $build = $this->contactViewBuilder->view($contact, 'full');
+    $expected_values['items'][] = [
+      'label' => 'Email',
+      'body' => "$name@example.com",
+    ];
+    $field_list_assert->assertPattern($expected_values, $this->renderRoot($build));
```
sergepavle, line 101: Again why we can't add all fields (rest) and then do single rendering and assertion?
For instance, in 'Call for tenders' content type it is reasonable to do that because we doing testing statuses (upcoming/past, etc) but it would be nice to understand why are we doing that here?
First of because rendering has some influence on performance.

tests/Functional/ContentEventRenderTest.php
yenyasinn: Yes, based on the decision above and root of cause is explained here https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-437?focusedCommentId=4402494&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4402494

tests/Functional/ContentEventRenderTest.php
yenyasinn: 'Online type' type is a label of a field. I assert behaviour of changing value of this field.

tests/Functional/ContentEventRenderTest.php
yenyasinn: Yes, this url is used in the current test.

tests/Functional/ContentProjectRenderTest.php
yenyasinn: Answered above.

tests/Functional/ContentEventRenderTest.php
yenyasinn: There are assertRegisterButtonEnabled and assertRegisterButtonDisabled methods that test state of the button. URL isn't changed. So I think in this particular case url can be hardcoded.

tests/Functional/ContentEventRenderTest.php
yenyasinn: Todos have been added. Also it is described in the https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-3356.

sergepavle (dismissed):

drishu (dismissed):

composer.json
drishu: I agree with @yenyasinn, the ewcms pr shows that there is no conflict with the other components, this would also bring the code base a bit closer to D9, so for me this is ok. Up to you @ademarco

yenyasinn: @drishu I have already rewritten code without patch.

templates/field/field--bare.html.twig
```diff
@@ -6,6 +6,4 @@
  * @see ./core/themes/stable/templates/field/field.html.twig
  */
 #}
-{% for item in items %}
-  {{ item.content }}
-{% endfor %}
+{% for item in items %}{{ item.content }}{% endfor %}
```
ademarco: Can we use `spaceless` instead?
```twig
{% spaceless %}
{% for item in items %}
  {{ item.content }}
{% endfor %}
{% endspaceless %}
```

templates/field/field--node--oe-project-budget-eu--oe-project---custom.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
+/**
+ * @file
+ * Theme override for the Overall budget field template.
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+{% for item in items %}{{ item.content }}{% endfor %}
```
ademarco: Same as above: can we use `spaceless`?

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+
+    // Assert "Online type", "Online link", "Online time" fields.
+    $online_start_date = (clone $static_time)->modify('+ 1 months');
+    $online_end_date = (clone $static_time)->modify('+ 2 months');
+    $node->set('oe_event_online_type', 'facebook');
+    $node->set('oe_event_online_link', [
+      'uri' => 'http://www.example.com/online_link',
+      'title' => 'Link to online event',
+    ]);
+    // @todo: "Online description" isn't shown on the page. Why do we need it?
```
ademarco: I've opened an issue about this, could you please change this comment into:

```
// The "Online description" field is not currently displayed.
// @todo: Assert its visibility as soon as the issue below will be fixed.
// @see https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-1063

File: tests/Functional/ContentEventRenderTest.php [hunk 15]
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+          'text' => 'Live streaming available',
+        ],
+      ],
+    ];
+    $icons_text_assert->assertPattern($icons_text_expected_values, $details_list_content->getOuterHtml());
   }
 
   /**
-   * Sets site's timezone.
+   * Assets enabled register button.
```
ademarco:
```suggestion
   * Asserts enabled registration button.
```

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
-   * @param string $timezone
-   *   Timezone.
+   * @param \Behat\Mink\Element\NodeElement $parent_element
+   *   Parent element.
    */
-  protected function setTimezone(string $timezone): void {
-    \Drupal::configFactory()->getEditable('system.date')
-      ->set('timezone.default', $timezone)
-      ->save();
+  protected function assertRegisterButtonEnabled(NodeElement $parent_element): void {
```
ademarco:
```suggestion
  protected function assertRegistrationButtonEnabled(NodeElement $parent_element): void {
```

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+    $rendered_button = $this->assertSession()->elementExists('css', 'a.ecl-u-mt-2xl.ecl-link.ecl-link--cta', $parent_element);
+    $this->assertEquals('Register here', $rendered_button->getText());
+    $this->assertEquals('http://www.example.com/registation', $rendered_button->getAttribute('href'));
+  }
+
+  /**
+   * Assets disabled register button.
```
ademarco:
```suggestion
   * Asserts disabled registration button.
```

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+   *
+   * @param \Behat\Mink\Element\NodeElement $element
+   *   Parent element.
+   */
+  protected function assertRegisterButtonDisabled(NodeElement $element): void {
```
ademarco:
```suggestion
  protected function assertRegistrationButtonDisabled(NodeElement $element): void {
```

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+   *   Parent element.
    */
-  protected function setTimezone(string $timezone): void {
-    \Drupal::configFactory()->getEditable('system.date')
-      ->set('timezone.default', $timezone)
-      ->save();
+  protected function assertRegisterButtonEnabled(NodeElement $parent_element): void {
+    $rendered_button = $this->assertSession()->elementExists('css', 'a.ecl-u-mt-2xl.ecl-link.ecl-link--cta', $parent_element);
+    $this->assertEquals('Register here', $rendered_button->getText());
+    $this->assertEquals('http://www.example.com/registation', $rendered_button->getAttribute('href'));
```
ademarco: Better to never hardcode assertion values, even if the assertion method is used only once.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -196,40 +201,34 @@ public function testProjectRendering(): void {
...
-    $file_title = $file_row->find('css', '.ecl-file__title');
-    $this->assertContains('Test document project_document', $file_title->getText());
-    $file_info_language = $file_row->find('css', '.ecl-file__info div.ecl-file__language');
-    $this->assertContains('English', $file_info_language->getText());
-    $file_info_properties = $file_row->find('css', '.ecl-file__info div.ecl-file__meta');
-    $this->assertContains('(2.96 KB - PDF)', $file_info_properties->getText());
-    $file_download_link = $file_row->find('css', '.ecl-file__download');
-    $this->assertContains('/sample_project_document.pdf', $file_download_link->getAttribute('href'));
-    $this->assertContains('Download', $file_download_link->getText());
+    $this->assertMediaDocumentDefaultRender($file_wrapper, 'project_document');
```
ademarco: We shouldn't hardcode assertion values below in the assertion method `assertMediaDocumentDefaultRender(...)`, let's instead pass them as method arguments, so they can be set by the caller.

```
    $this->assertContains('English', $file_info_language->getText());
    $this->assertContains('(2.96 KB - PDF)', $file_info_properties->getText());
```

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -239,47 +238,29 @@ public function testProjectRendering(): void {
...
-    $file_title = $file_row->find('css', '.ecl-file__title');
-    $this->assertContains('Test document project_result', $file_title->getText());
-    $file_info_language = $file_row->find('css', '.ecl-file__info div.ecl-file__language');
-    $this->assertContains('English', $file_info_language->getText());
-    $file_info_properties = $file_row->find('css', '.ecl-file__info div.ecl-file__meta');
-    $this->assertContains('KB - PDF)', $file_info_properties->getText());
-    $file_download_link = $file_row->find('css', '.ecl-file__download');
-    $this->assertContains('/sample_project_result.pdf', $file_download_link->getAttribute('href'));
-    $this->assertContains('Download', $file_download_link->getText());
+    $this->assertMediaDocumentDefaultRender($file_wrapper, 'project_result');
```
ademarco: Same as above, we need to pass the language and document size parts as arguments in the assert method `assertMediaDocumentDefaultRender(...)`

ademarco (changes requested):

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+      'uri' => 'http://www.example.com/online_link',
+      'title' => 'Link to online event',
+    ]);
+    // @todo: "Online description" isn't shown on the page. Why do we need it?
+    $node->set('oe_event_online_description', 'Online event description');
+    $node->set('oe_event_online_dates', [
+      'value' => $online_start_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+      'end_value' => $online_end_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+    ])->save();
+    $this->drupalGet($node->toUrl());
```
ademarco, line 328: When testing all the different messages, that depends on the current time, we shouldn't touch the node at all, we only have to modify the time, via the datetime testing, and run cron, as we do on Behat:

image

This is important since, if we modify the node and reload the page, then cache is invalidated by modifying the actual node, so we are not testing that time related cache tags (set in the extra fields) are working correctly.

Everything related to time testing should never involve modifying the node dates, but it should always consist of the following steps:

- Change frozen time to desired one
- Run cron (this will invalidate the cache)
- Reload the page
- Run assertions

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+    $registration_content = $this->assertSession()->elementExists('css', '#event-registration-block');
+    $this->assertRegisterButtonEnabled($registration_content);
+
+    // Assert "Registration date" field in the past.
+    $registration_start_date = (clone $static_time)->modify('- 2 months');
+    $registration_end_date = (clone $static_time)->modify('- 1 months');
+    $node->set('oe_event_registration_dates', [
+      'value' => $registration_start_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+      'end_value' => $registration_end_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+    ])->save();
```
ademarco: See above, when testing time-related features we must never change the actual event dates, we should change time, run cron and reload the page, instead.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
 
-    $this->setTimezone('America/New_York');
-    \Drupal::entityTypeManager()->getViewBuilder('node')->resetCache([$node]);
+    // Assert "Registration date" field when registration is in progress.
+    $registration_start_date = (clone $static_time)->modify('- 10 days');
+    $registration_end_date = (clone $static_time)->modify('+ 10 days');
+    $node->set('oe_event_registration_dates', [
+      'value' => $registration_start_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+      'end_value' => $registration_end_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+    ])->save();
```
ademarco: See above, when testing time-related features we must never change the actual event dates, we should change time, run cron and reload the page, instead.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+    $this->assertRegisterButtonEnabled($registration_content);
+    $this->assertEquals('Book your seat, 1 week left to register, registration will end on 28 February 2020, 01:00', $registration_info_content->getText());
+
+    // Assert "Registration date" field when registration will finish today.
+    $registration_start_date = (clone $static_time)->modify('- 10 days');
+    $registration_end_date = (clone $static_time)->modify('+ 1 hours');
     $node->set('oe_event_registration_dates', [
       'value' => $registration_start_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
       'end_value' => $registration_end_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
     ])->save();
```
ademarco: See above, when testing time-related features we must never change the actual event dates, we should change time, run cron and reload the page, instead.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+    $this->assertEquals('Book your seat, the registration will end today, 18 February 2020, 02:00', $registration_info_content->getText());
+
+    // Assert "Registration date" field when registration will start today.
+    $registration_start_date = (clone $static_time)->modify('+ 1 hour');
+    $registration_end_date = (clone $static_time)->modify('+ 10 days');
+    $node->set('oe_event_registration_dates', [
+      'value' => $registration_start_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+      'end_value' => $registration_end_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+    ])->save();
+    $this->drupalGet($node->toUrl());
```
ademarco: See above, when testing time-related features we must never change the actual event dates, we should change time, run cron and reload the page, instead.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+    $this->assertRegisterButtonDisabled($registration_content);
+    $this->assertEquals('Registration will open today, 18 February 2020, 02:00.', $registration_info_content->getText());
+
+    // Assert "Registration date" field when registration will start in future.
+    $registration_start_date = (clone $static_time)->modify('+ 1 day');
+    $registration_end_date = (clone $static_time)->modify('+ 10 days');
+    $node->set('oe_event_registration_dates', [
+      'value' => $registration_start_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+      'end_value' => $registration_end_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+    ])->save();
```
ademarco: See above, when testing time-related features we must never change the actual event dates, we should change time, run cron and reload the page, instead.

tests/Functional/ContentEventRenderTest.php
```diff
@@ -183,42 +167,511 @@ public function testEventRegistrationDateDescription(): void {
...
+
+    // Assert "Report text" and "Summary for report" fields when event finished.
+    $start_date = (clone $static_time)->modify('- 20 days');
+    $end_date = (clone $static_time)->modify('- 10 days');
+    $node->set('oe_event_report_summary', 'Event report summary');
+    $node->set('oe_event_report_text', 'Event report text');
+    $node->set('oe_event_dates', [
+      'value' => $start_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+      'end_value' => $end_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT),
+    ])->save();
```
ademarco, line 439: See above, when testing time-related features we must never change the actual event dates, we should change time, run cron and reload the page, instead.

ademarco (changes requested):
It seems we didn't port this scenario:

```gherkin
    # Assert event rendering after the event has ended.
    When the time is frozen at "21 June 2020 2pm"
    And I run cron
    And I reload the page
    Then I should see the heading "Report"
    And I should see the text "Report summary"
    And I should see the text "Report text"
    But I should not see the heading "Description"
    And I should not see the text "Event description summary"
    And I should not see the text "Event description"
    And I should not see the registration block
```

templates/field/field--bare.html.twig
yenyasinn: Since we don't support PHP 7.2 anymore we can leave it as is.

tests/Functional/ContentEventRenderTest.php
yenyasinn: Done

tests/Functional/ContentEventRenderTest.php
yenyasinn: Done

tests/Functional/ContentRenderTestBase.php
```diff
@@ -87,20 +95,28 @@ protected function getStorage(string $entity_type_id): EntityStorageInterface {
...
+   *   Document language.
+   * @param string $meta
+   *   Size and format of the file.
+   * @param string $link
+   *   Link to the file.
+   * @param string $button_label
+   *   Text on the button.
    */
-  protected function assertMediaDocumentDefaultRender(NodeElement $element, string $name): void {
+  protected function assertMediaDocumentDefaultRender(NodeElement $element, string $name, string $language, string $meta, string $link, string $button_label): void {
```
ademarco, line 108: The button label is hardcoded in the media template, we should remove it from the assertion method, as it's something that cannot be changed by the application, so we don't need to test it. Let's only test the button URL.

image

tests/Functional/ContentCallForTendersRenderTest.php
```diff
@@ -79,163 +65,240 @@ public function testTenderRendering(): void {
...
+        'body' => '21 February 2020, 01:00 (AEDT)',
+      ],
+    ];
+    $field_list_assert->assertPattern($details_expected_values, $content_items[0]->getHtml());
+    $this->assertDeadlineDateStrike($content);
+
+    // Assert "Closed" status.
+    $deadline_date = (clone $static_time)->modify('- 2 days');
+    $node->set('oe_call_tenders_deadline', $deadline_date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT))->save();
+    $this->drupalGet($node->toUrl());
```
ademarco, line 179: Same as for events, we don't have to change the date and refresh the page, we need to move the time, run cron and assert the change, or else we are not testing our time based tags.

Please also change this on ContentCallForProposalRenderTest.php

tests/Functional/ContentCallForTendersRenderTest.php
```diff
@@ -79,163 +65,240 @@ public function testTenderRendering(): void {
...
+      ],
+    ];
+    $field_list_assert->assertPattern($details_expected_values, $content_items[0]->getHtml());
+    $this->assertDeadlineDateStrike($content);
 
-    // Assert Responsible department field label.
+    // Assert "Upcoming" status.
+    $opening_date = (clone $static_time)->modify('+ 2 days');
+    $node->set('oe_call_tenders_opening_date', ['value' => $opening_date->format('Y-m-d')])->save();
+    $this->drupalGet($node->toUrl());
```
ademarco, line 152: Same as per event, we need to move the time, run cron and refresh the page, not change the node, or else we are not testing our time based tags, since resaving the node will invalidate the tags anyway.

ademarco (changes requested):

tests/Functional/ContentRenderTestBase.php
yenyasinn: "Download" is translatable string and it can be different on different languages.
Also in some another template it can be changed

ademarco: Ok, let's leave it.

ademarco (approved):

imanoleguskiza (approved):

PR #632, opened 2020-10-27
EWPP-441: Add assertions for text_featured_media pattern.

enriquelacoma:
### Description

Create assertion  for text_featured_media pattern

### Change log

- Added: TextFeaturedMediaAssert.php
- Changed: assertImage() has been moved to BasePatternAssert

### Example

```sh
$assert = new TextFeaturedMediaAssert();
$expected_values = [
  'title' => 'Lorem ipsum',
  'image' => [
    'alt' => 'Alternative text event_featured_media',
    'src' => 'http://localhost:8080/build/sites/simpletest/59151335/files/placeholder_event_featured_media.png',
  ],
  'caption' => 'Event featured media legend',
  'text' => 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Non enim iam stirpis bonum quaeret, sed animalis.',
  //'video' => 'Rendering string as a video',
  //'video_ratio' => '16-9',
];
```


tests/PatternAssertions/FeaturedMediaAssert.php
```diff
@@ -0,0 +1,90 @@
...
+namespace Drupal\Tests\oe_theme\PatternAssertions;
+
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Assertions for text_featured_media pattern.
+ *
+ *  @see ./templates/patterns/text_featured_media/text_featured_media.ui_patterns.yml
+ */
+class FeaturedMediaAssert extends BasePatternAssert {
```
drishu: we should call this TextFeaturedMediaAssert

tests/PatternAssertions/FeaturedMediaAssert.php
```diff
@@ -0,0 +1,90 @@
...
+   * Asserts the image of the pattern.
+   *
+   * @param array|null $expected_image
+   *   The expected image values.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertImage($expected_image, Crawler $crawler): void {
+    $image_div = $crawler->filter('div.ecl-row figure.ecl-media-container img');
+    self::assertEquals($expected_image['alt'], $image_div->attr('alt'));
```
drishu: we could use assertElementAttribute here

tests/PatternAssertions/FeaturedMediaAssert.php
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Asserts the video url of the pattern.
+   *
+   * @param string|null $expected_video
+   *   The expected video url.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertVideo($expected_video, Crawler $crawler): void {
+    $video = $crawler->filter('div .ecl-row figure.ecl-media-container iframe');
```
drishu: assertElementAttribute

tests/PatternAssertions/FeaturedMediaAssert.php
```diff
@@ -0,0 +1,90 @@
...
+      'video_ratio' => [
+        [$this, 'assertRatio'],
+      ],
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function assertBaseElements(string $html, string $variant): void {
```
drishu: this seems silly, lets suggest to not make this a abstract method in the parent class and remove this

tests/PatternAssertions/FeaturedMediaAssert.php
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Asserts the video ratio of the pattern.
+   *
+   * @param string|null $expected_ratio_class
+   *   The expected class with the video ratio.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertRatio($expected_ratio_class, Crawler $crawler): void {
+    self::assertElementExists('div .ecl-row figure.ecl-media-container div.' . $expected_ratio_class, $crawler);
```
drishu: $expected_ratio_class would need to be something like 'ecl-media-container__media--ratio-16-9', lets make it so we send only 16-9, so add the rest of the class here

```suggestion
    self::assertElementExists('div .ecl-row figure.ecl-media-container div.ecl-media-container__media--ratio-' . $expected_ratio, $crawler);
```

drishu (changes requested):

tests/PatternAssertions/BasePatternAssert.php
```diff
@@ -168,4 +168,23 @@ protected function assertElementNotExists(string $selector, Crawler $crawler): v
     ));
   }
 
+  /**
+   * Asserts the image of the pattern.
+   *
+   * @param array|null $expected_image
```
yenyasinn, line 174: I have not understood why type is set as `array|null` or `string|null` for all expected values. If we expect value then it shouldn't be NULL, from my point of view.

tests/PatternAssertions/TextFeaturedMediaAssert.php
```diff
@@ -0,0 +1,76 @@
...
+        [$this, 'assertVideoRatio'],
+        'div.ecl-row figure.ecl-media-container div.ecl-media-container__media',
+      ],
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function assertBaseElements(string $html, string $variant): void {}
```
yenyasinn, line 52: Looks like sometimes we just need empty method.

tests/PatternAssertions/FeaturedMediaAssert.php
yenyasinn: let's leave it empty.  will see what maintainer say

tests/PatternAssertions/TextFeaturedMediaAssert.php
```diff
@@ -0,0 +1,76 @@
...
+      'caption' => [
+        [$this, 'assertElementText'],
+        'div.ecl-row figure figcaption.ecl-media-container__caption',
+      ],
+      'image' => [
+        [$this, 'assertImage'],
+        'div.ecl-row figure.ecl-media-container img',
+      ],
+      'video' => [
+        [$this, 'assertElementText'],
```
drishu, line 39: adding iframe to the selector and then assertElementExists (or even assertElementAttribute for the src) would have made sense, but its ok for me like this since the pattern definition for video is basically free text

drishu (approved):

tests/PatternAssertions/TextFeaturedMediaAssert.php
yenyasinn: Would be better avoid hardcode iframe here since video can use `<video>` tag.

yenyasinn:
Closed in favour of https://github.com/openeuropa/oe_theme/pull/686/files

PR #633, opened 2020-10-27
EWPP-440: Create assertion for Social media links pattern.

yenyasinn:
### Description

1. Merged Social media links horizontal and vertical to one pattern as described in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-2489.

2. SocialMediaLinksAssert has been added to test Social media links pattern.

## Example

```
    $html = $this->assertSession()->elementExists('css', '.ecl-social-media-follow')->getOuterHtml();
    $assert = new SocialMediaLinksAssert();
    $expected_values = [
      'title' => 'Social media',
      'links' => [
        [
          'service' => 'facebook',
          'label' => 'Social media general_contact',
          'url' => 'http://www.example.com/social_media_general_contact',
        ]
      ]
    ];
    $assert->assertPattern($expected_values, $html);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,13 +83,13 @@ To learn more about EC/EU families and ECL branding visit the [ECL website](http
 
 #### Social media links pattern
 
-In 2.13.0 we introduced new pattern ["Social media links"](./templates/patterns/social_media_links) with two variants:
+In 2.13.0 we introduced a new pattern ["Social media links"](./templates/patterns/social_media_links) with two variants:
 
 - `horizontal`: social media links will be arranged horizontally.
 - `vertical`: social media links will be arranged vertically.
 
-Therefore patterns "Social media links: horizontal" and "Social media links: vertical" are deprecated. Use "Social media
-links" pattern with appropriate variant instead.
+Therefore patterns "Social media links: horizontal" and "Social media links: vertical" are now deprecated. Use the "Social media
+links" pattern with an appropriate variant instead.
 
 ### Upgrade to 2.10.0
 
--- a/README.md
+++ b/README.md
@@ -79,6 +79,18 @@ users can access to the main navigation menu only on sites using standardised EC
 
 To learn more about EC/EU families and ECL branding visit the [ECL website](https://ec.europa.eu/component-library).
 
+### Upgrade to 2.13.0
+
+#### Social media links pattern
+
+In 2.13.0 we introduced new pattern ["Social media links"](./templates/patterns/social_media_links) with two variants:
+
+- `horizontal`: social media links will be arranged horizontally.
+- `vertical`: social media links will be arranged vertically.
+
+Therefore patterns "Social media links: horizontal" and "Social media links: vertical" are deprecated. Use "Social media
+links" pattern with appropriate variant instead.
+
 ### Upgrade to 2.10.0
 
 #### ECL page header
--- a/README.md
+++ b/README.md
@@ -79,11 +79,11 @@ users can access to the main navigation menu only on sites using standardised EC
 
 To learn more about EC/EU families and ECL branding visit the [ECL website](https://ec.europa.eu/component-library).
 
-### Upgrade to 2.13.0
+### Upgrade to 2.15.0
 
 #### Social media links pattern
 
-In 2.13.0 we introduced a new pattern ["Social media links"](./templates/patterns/social_media_links) with two variants:
+In 2.15.0 we introduced a new pattern ["Social media links"](./templates/patterns/social_media_links) with two variants:
 
 - `horizontal`: social media links will be arranged horizontally.
 - `vertical`: social media links will be arranged vertically.
--- a/templates/patterns/social_media_links/social_media_links.ui_patterns.yml
+++ b/templates/patterns/social_media_links/social_media_links.ui_patterns.yml
@@ -28,3 +28,5 @@ social_media_links:
         - service: "instagram"
           label: "Instagram"
           url: "#"
+        - label: "See more"
+          url: "#"
--- a/templates/patterns/social_media_links/social_media_links.ui_patterns.yml
+++ b/templates/patterns/social_media_links/social_media_links.ui_patterns.yml
@@ -1,13 +1,13 @@
 social_media_links:
   label: "Social media links"
-  description: "Social network icon with link of label."
+  description: "Showcases a list of social media links with their respective icons."
   variants:
     horizontal:
       label: "Horizontal"
-      description: "It will align the field items horizontally."
+      description: "Social media links will be arranged horizontally."
     vertical:
       label: "Vertical"
-      description: "It will align the field items vertically."
+      description: "Social media links will be arranged vertically."
   fields:
     title:
       type: "text"
--- a/templates/patterns/social_media_links/social_media_links.ui_patterns.yml
+++ b/templates/patterns/social_media_links/social_media_links.ui_patterns.yml
@@ -1,6 +1,13 @@
-social_media_links_vertical:
-  label: "Social media links: vertical (Deprecated)"
+social_media_links:
+  label: "Social media links"
   description: "Social network icon with link of label."
+  variants:
+    horizontal:
+      label: "Horizontal"
+      description: "It will align the field items horizontally."
+    vertical:
+      label: "Vertical"
+      description: "It will align the field items vertically."
   fields:
     title:
       type: "text"
@@ -21,6 +28,3 @@ social_media_links_vertical:
         - service: "instagram"
           label: "Instagram"
           url: "#"
-  additional:
-    style: "max-width: 400px"
-    hidden: true
--- a/tests/PatternAssertions/SocialMediaLinksAssert.php
+++ b/tests/PatternAssertions/SocialMediaLinksAssert.php
@@ -68,11 +68,11 @@ class SocialMediaLinksAssert extends BasePatternAssert {
     if (in_array('ecl-social-media-follow--vertical', $existing_classes)) {
       return 'vertical';
     }
-    throw new Exception('Variant of social media links pattern is not identified.');
+    throw new Exception('Variant of social media links pattern could not be identified.');
   }
 
   /**
-   * Asserts the links of the pattern.
+   * Asserts the links of social media follow.
    *
    * @param array $expected_items
    *   The expected item values.
--- a/tests/PatternAssertions/SocialMediaLinksAssert.php
+++ b/tests/PatternAssertions/SocialMediaLinksAssert.php
@@ -1,0 +1,108 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme\PatternAssertions;
+
+use Symfony\Component\DomCrawler\Crawler;
+use PHPUnit\Framework\Exception;
+
+/**
+ * Assertions for the field list pattern.
+ *
+ * @package Drupal\Tests\oe_theme\PatternAssertions
+ */
+class SocialMediaLinksAssert extends BasePatternAssert {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getAssertions(string $variant): array {
+    return [
+      'title' => [
+        [$this, 'assertElementText'],
+        'div' . $this->getBaseItemClass($variant) . ' p.ecl-social-media-follow__description',
+      ],
+      'links' => [
+        [$this, 'assertLinks'],
+      ],
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function assertBaseElements(string $html, string $variant): void {
+    $crawler = new Crawler($html);
+    $field_list_container = $crawler->filter('div' . $this->getBaseItemClass($variant));
+    self::assertCount(1, $field_list_container);
+  }
+
+  /**
+   * Returns the base CSS selector for a field item depending on the variant.
+   *
+   * @param string $variant
+   *   The variant being checked.
+   *
+   * @return string
+   *   The base selector for the variant.
+   */
+  protected function getBaseItemClass(string $variant): string {
+    if ($variant === 'horizontal') {
+      return '.ecl-social-media-follow.ecl-social-media-follow--horizontal';
+    }
+    return '.ecl-social-media-follow.ecl-social-media-follow--vertical';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getPatternVariant(string $html): string {
+    $crawler = new Crawler($html);
+    $field_list_container = $crawler->filter('div.ecl-social-media-follow');
+    $existing_classes = $field_list_container->attr('class');
+    $existing_classes = explode(' ', $existing_classes);
+    if (in_array('ecl-social-media-follow--horizontal', $existing_classes)) {
+      return 'horizontal';
+    }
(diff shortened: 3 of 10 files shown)


tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,108 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme\PatternAssertions;
+
+use Symfony\Component\DomCrawler\Crawler;
+use PHPUnit\Framework\Exception;
+
+/**
+ * Assertions for the field list pattern.
```
drishu: for the social media ...

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,108 @@
...
+   * {@inheritdoc}
+   */
+  protected function assertBaseElements(string $html, string $variant): void {
+    $crawler = new Crawler($html);
+    $field_list_container = $crawler->filter('div' . $this->getBaseItemClass($variant));
+    self::assertCount(1, $field_list_container);
+  }
+
+  /**
+   * Returns the base CSS selector for a field item depending on the variant.
```
drishu: .. for a social media item depen..

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,108 @@
...
+        // Service icon is absent.
+        $link_element = $label_element = $li_item->filter('a.ecl-link.ecl-link--standalone.ecl-social-media-follow__link');
+      }
+      else {
+        // Service icon exists.
+        $link_element = $li_item->filter('a.ecl-link.ecl-link--standalone.ecl-link--icon.ecl-link--icon-before.ecl-social-media-follow__link');
+        $label_element = $link_element->filter('span.ecl-link__label');
+        $icon_social_file = 'icons-social.svg#';
+        $svg = $link_element->filter('svg.ecl-icon.ecl-icon--xl.ecl-link__icon.ecl-social-media-follow__icon use');
+        self::assertContains($icon_social_file . $expected_item['service'], $svg->attr('xlink:href'));
```
drishu: concatenation happens anyway but the value does not change, so I dont see the point in adding this to a variable

```suggestion
        self::assertContains('icons-social.svg#'. $expected_item['service'], $svg->attr('xlink:href'));
```

same below

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,108 @@
...
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertLinks(array $expected_items, Crawler $crawler): void {
+    $li_items = $crawler->filter('li.ecl-social-media-follow__item');
+    self::assertCount(count($expected_items), $li_items);
+
+    foreach ($expected_items as $index => $expected_item) {
+      $li_item = $li_items->eq($index);
+      if (empty($expected_item['service'])) {
```
drishu, line 88: this `if` actually just adds cyclomatic complexity, we can do without, lets set some healthy default and check for the `service` part only

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,108 @@
...
+        // Service icon is absent.
+        $link_element = $label_element = $li_item->filter('a.ecl-link.ecl-link--standalone.ecl-social-media-follow__link');
+      }
+      else {
+        // Service icon exists.
+        $link_element = $li_item->filter('a.ecl-link.ecl-link--standalone.ecl-link--icon.ecl-link--icon-before.ecl-social-media-follow__link');
```
drishu, line 94: `.ecl-link--icon.ecl-link--icon-before` is actually not needed to find the correct element, so since the selector can be the same lets set this outside the if, do the same with the label and just overwrite the label selector if the `service` is set

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,108 @@
...
+        $label_element = $link_element->filter('span.ecl-link__label');
+        $icon_social_file = 'icons-social.svg#';
```
drishu: this actually does not change, so either put it outside the loop or remove it all together and just leave static string inside selector

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,108 @@
...
+        $svg = $link_element->filter('svg.ecl-icon.ecl-icon--xl.ecl-link__icon.ecl-social-media-follow__icon use');
+        self::assertContains($icon_social_file . $expected_item['service'], $svg->attr('xlink:href'));
+        $svg_hover = $link_element->filter('svg.ecl-icon.ecl-icon--xl.ecl-link__icon.ecl-social-media-follow__icon-hover use');
+        self::assertContains($icon_social_file . $expected_item['service'] . '_hover', $svg_hover->attr('xlink:href'));
+      }
+
```
drishu, line 101: with above refactoring, we could just set css selectors above and fetch the elements just once here

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,108 @@
...
+      self::assertEquals($expected_item['label'], trim($label_element->text()));
+      self::assertEquals($expected_item['url'], $link_element->attr('href'));
```
drishu, line 103: above suggestions would improve readability and reduce complexity, I imagine it something like:

`
foreach ($expected_items as $index => $expected_item) {
      $li_item = $li_items->eq($index);
      $label_selector = $link_selector = 'a.ecl-link.ecl-link--standalone.ecl-social-media-follow__link';

      if (!empty($expected_item['service'])) {
        $label_selector = 'span.ecl-link__label';
        $svg = $li_item->filter('svg.ecl-icon.ecl-icon--xl.ecl-link__icon.ecl-social-media-follow__icon use');
        self::assertContains('icons-social.svg#' . $expected_item['service'], $svg->attr('xlink:href'));
        $svg_hover = $li_item->filter('svg.ecl-icon.ecl-icon--xl.ecl-link__icon.ecl-social-media-follow__icon-hover use');
        self::assertContains('icons-social.svg#' . $expected_item['service'] . '_hover', $svg_hover->attr('xlink:href'));
      }

      $link_element = $li_item->filter($link_selector);
      $label_element = $li_item->filter($label_selector);

      self::assertEquals($expected_item['label'], trim($label_element->text()));
      self::assertEquals($expected_item['url'], $link_element->attr('href'));
    }
`

drishu (changes requested):

tests/PatternAssertions/SocialMediaLinksAssert.php
yenyasinn: Such approach doesn't test nesting `svg` inside `a` that is worth, I suppose.

This implementation can be rewritten and support nesting but it requires additional useless invocation of `filter` method that, so I would propose leave current implementation as is.

drishu (dismissed):

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,107 @@
...
+    $field_list_container = $crawler->filter('div.ecl-social-media-follow');
+    $existing_classes = $field_list_container->attr('class');
+    $existing_classes = explode(' ', $existing_classes);
+    if (in_array('ecl-social-media-follow--horizontal', $existing_classes)) {
+      return 'horizontal';
+    }
+    if (in_array('ecl-social-media-follow--vertical', $existing_classes)) {
+      return 'vertical';
+    }
+    throw new Exception('Variant of social media links pattern is not identified.');
```
22Alexandra:
```suggestion
    throw new Exception('Variant of social media links pattern could not be identified.');
```

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Asserts the links of the pattern.
```
22Alexandra:
```suggestion
   * Asserts the links of social media follow.
```

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,107 @@
...
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertLinks(array $expected_items, Crawler $crawler): void {
+    $li_items = $crawler->filter('li.ecl-social-media-follow__item');
+    self::assertCount(count($expected_items), $li_items);
+
+    foreach ($expected_items as $index => $expected_item) {
+      $li_item = $li_items->eq($index);
+      if (empty($expected_item['service'])) {
```
22Alexandra, line 88: Should we print the link if there is no service?
It's mandatory in ecl : https://ec.europa.eu/component-library/ec/components/social-media-follow/usage/
There is no validation in twig, but if there is no service maybe we shouldn't print the link at all.
This removed part from the deprecated tpl was building a link only if there was a service we should keep the same logic:
https://github.com/openeuropa/oe_theme/pull/633/files#diff-41b2bab11b31265a931f90314e29dd9656b15c21edc1383988bf29c834dd5e2bL14-L31
Which is in fact updated but if a link doesn't have a service is still printed:
link without service

22Alexandra (changes requested):

tests/PatternAssertions/SocialMediaLinksAssert.php
yenyasinn: We don't need "service" if simple link should be printed, like "See more". See templates/patterns/social_media_links/social_media_links.ui_patterns.yml.

Screenshot 2020-12-11 at 16 49 55

> This removed part from the deprecated tpl was building a link only if there was a service we should keep the same logic

Might be I don't understand something but in the deprecated tpl link is added in any case

Screenshot 2020-12-11 at 16 53 20

Also please have a look on the implementation oe_theme_preprocess_paragraph__oe_social_media_follow in oe_theme.theme - there is link without service.

tests/PatternAssertions/SocialMediaLinksAssert.php
yenyasinn: Yes, `.ecl-link--icon.ecl-link--icon-before` can be skipped, and also all other classes except `.ecl-social-media-follow__link` but it is required for css to set icon properly.

tests/PatternAssertions/SocialMediaLinksAssert.php
yenyasinn: See answer below.

tests/PatternAssertions/SocialMediaLinksAssert.php
yenyasinn: See discussion below

22Alexandra (dismissed):

nagyad (dismissed):

templates/patterns/social_media_links/social_media_links.ui_patterns.yml
```diff
@@ -0,0 +1,32 @@
...
+        - service: "facebook"
+          label: "Facebook"
+          url: "#"
+        - service: "twitter"
+          label: "Twitter"
+          url: "#"
+        - service: "instagram"
+          label: "Instagram"
+          url: "#"
+        - label: "See more"
```
imanoleguskiza, line 31: I don't think this link is needed, it implies some functionality that is no there on the pattern.

templates/patterns/social_media_links/social_media_links.ui_patterns.yml
```diff
@@ -0,0 +1,32 @@
+social_media_links:
+  label: "Social media links"
+  description: "Social network icon with link of label."
```
imanoleguskiza: This really means nothing, lets put something more meaningful now that we can: "Showcases a list of social media links with their respective icons."

templates/patterns/social_media_links/social_media_links.ui_patterns.yml
```diff
@@ -0,0 +1,32 @@
...
+  variants:
+    horizontal:
+      label: "Horizontal"
+      description: "It will align the field items horizontally."
```
imanoleguskiza: There are no field items here, lets say "Social media links will be arranged horizontally."

templates/patterns/social_media_links/social_media_links.ui_patterns.yml
```diff
@@ -0,0 +1,32 @@
...
+    vertical:
+      label: "Vertical"
+      description: "It will align the field items vertically."
```
imanoleguskiza: Update to follow above.

tests/PatternAssertions/SocialMediaLinksAssert.php
```diff
@@ -0,0 +1,107 @@
...
+      return 'horizontal';
+    }
+    if (in_array('ecl-social-media-follow--vertical', $existing_classes)) {
+      return 'vertical';
+    }
+    throw new Exception('Variant of social media links pattern could not be identified.');
+  }
+
+  /**
+   * Asserts the links of social media follow.
```
imanoleguskiza: Asserts the links of the social media links pattern.

imanoleguskiza (changes requested):

templates/patterns/social_media_links/social_media_links.ui_patterns.yml
yenyasinn: It exists here https://ec.europa.eu/component-library/playground/ec/?path=/story/components-socialmediafollow--horizontal, so I think we should support it.

imanoleguskiza (dismissed):

ademarco (changes requested):
It's ok to deprecate patterns, especially as those two where done when UI Patterns wasn't yet supporting variants, only we should document that in the theme's readme. We are using [this format](https://github.com/openeuropa/oe_theme/blob/2.x/README.md#upgrade-to-2100) for now, we could add a bullet point there about this deprecation.

README.md
```diff
@@ -79,6 +79,18 @@ users can access to the main navigation menu only on sites using standardised EC
 
 To learn more about EC/EU families and ECL branding visit the [ECL website](https://ec.europa.eu/component-library).
 
+### Upgrade to 2.13.0
```
ademarco:
```suggestion
### Upgrade to 2.15.0
```

README.md
```diff
@@ -79,6 +79,18 @@ users can access to the main navigation menu only on sites using standardised EC
...
+
+#### Social media links pattern
+
+In 2.13.0 we introduced a new pattern ["Social media links"](./templates/patterns/social_media_links) with two variants:
```
ademarco:
```suggestion
In 2.15.0 we introduced a new pattern ["Social media links"](./templates/patterns/social_media_links) with two variants:
```

ademarco (changes requested):

ademarco (approved):

22Alexandra (approved):

PR #634, opened 2020-11-03
EWPP-437: Theme full view for "Call for proposals" content type

yenyasinn:


modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
```diff
@@ -0,0 +1,82 @@
...
+      ],
+    ],
+  ];
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_view_alter().
+ */
+function oe_theme_content_call_proposals_node_view_alter(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display) {
+  // Run only for "Call for tenders" content type.
```
drishu:
```suggestion
  // Run only for "Call for proposals" content type.
```

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
```diff
@@ -0,0 +1,82 @@
...
+  _oe_theme_content_call_proposals_strike_deadline($variables);
+}
+
+/**
+ * Add strike through to deadline date when call for proposals is close.
+ *
+ * @param array $variables
+ *   Element variables from preprocess.
+ */
+function _oe_theme_content_call_proposals_strike_deadline(&$variables) {
```
drishu:
```suggestion
function _oe_theme_content_call_proposals_strike_deadline(&$variables): void {
```

modules/oe_theme_content_call_proposals/src/Plugin/ExtraField/Display/CallForProposalsStatusExtraField.php
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Time based cache tag generator service.
+   *
+   * @var \Drupal\oe_time_caching\Cache\TimeBasedCacheTagGeneratorInterface
+   */
+  protected $cacheTagGenerator;
+
+  /**
+   * CallForTendersStatusExtraField constructor.
```
drishu:
```suggestion
   * CallForProposalsStatusExtraField constructor.
```

modules/oe_theme_content_call_proposals/src/Plugin/ExtraField/Display/CallForProposalsStatusExtraField.php
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(ContentEntityInterface $entity) {
+    $entity = CallForProposalsNodeWrapper::getInstance($entity);
+    $cacheable = CacheableMetadata::createFromRenderArray(['#cache' => ['contexts' => ['timezone']]]);
+
+    $status = $entity->getStatus();
+    // Set cache tags based on date.
+    if ($status === CallForProposalsNodeWrapper::STATUS_UPCOMING) {
```
drishu:
```suggestion
    if ($status === CallForProposalsNodeWrapperInterface::STATUS_UPCOMING) {
```

modules/oe_theme_content_call_proposals/src/Plugin/ExtraField/Display/CallForProposalsStatusExtraField.php
```diff
@@ -0,0 +1,100 @@
...
+      $cacheable->addCacheTags($this->cacheTagGenerator->generateTags($entity->getOpeningDate()->getPhpDateTime()));
+    }
+    if ($status === CallForProposalsNodeWrapper::STATUS_OPEN) {
```
drishu:
```suggestion
    if ($status === CallForProposalsNodeWrapperInterface::STATUS_OPEN) {
```

modules/oe_theme_content_call_proposals/src/Plugin/PageHeaderMetadata/CallForProposalsContentType.php
```diff
@@ -0,0 +1,49 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_content_call_proposals\Plugin\PageHeaderMetadata;
+
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\oe_theme_helper\Plugin\PageHeaderMetadata\NodeViewRoutesBase;
+use Drupal\oe_content_call_proposals\CallForProposalsNodeWrapper;
+
+/**
+ * Page header metadata for the OpenEuropa "Call for tenders" content type.
```
drishu:
```suggestion
 * Page header metadata for the OpenEuropa "Call for proposals" content type.
```

modules/oe_theme_content_call_tenders/src/Plugin/ExtraField/Display/CallForTendersStatusExtraField.php
```diff
@@ -89,8 +89,9 @@ public function viewElements(ContentEntityInterface $entity) {
       $cacheable->addCacheTags($this->cacheTagGenerator->generateTags($entity->getDeadlineDate()->getPhpDateTime()));
     }
     $build = [
-      '#theme' => 'oe_theme_content_call_tenders_status',
+      '#theme' => 'oe_theme_helper_call_status',
```
drishu, line 92: nice

modules/oe_theme_helper/templates/oe-theme-helper-call-status.html.twig
```diff
@@ -6,4 +6,4 @@
  * @see \Drupal\oe_theme_content_call_tenders\Plugin\ExtraField\Display\CallForTendersStatusExtraField::viewElements()
```
drishu: maybe we should rather refer to the preprocess function in the theme helper

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
```diff
@@ -0,0 +1,82 @@
...
+    if (!$entity->get($name)->isEmpty() && isset($build[$name]['#title'])) {
+      $build[$name]['#title'] = $label_override;
+    }
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_content_call_proposals_preprocess_field__node__oe_call_proposals_deadline__oe_call_proposals__full(&$variables, $hook) {
```
drishu: I see this also in ternders naming.. since the field machine name already contains the content type prefix, I don't understand why we need `oe_call_proposals` at the end

```suggestion
function oe_theme_content_call_proposals_preprocess_field__node__oe_call_proposals_deadline__full(&$variables, $hook) {
```

I know its not wrong, I just don't see the point of it.

drishu (changes requested):
Let's also add a Readme.md

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
yenyasinn: It was implemented so for consistency. `field-node--[field_name]--[bundle]--[view_mode]` is used in another CTs

modules/oe_theme_content_call_proposals/templates/oe_theme_content_call_proposals_publication_details.html.twig
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @file
+ * Theme for the publication details extra field on Call for proposals content type.
+ *
+ * @see \Drupal\oe_theme_content_call_proposals\Plugin\ExtraField\Display\CallForProposalsPublicationDetailsExtraField::viewElements()
+ */
+#}
+{{ date }}
+{% if link is not empty %}
+  {{ 'in'|t }}
```
yenyasinn: It's quite tricky to use ecl-twig/link and avoid escaping it using t() so it is implemented so.

modules/oe_theme_content_call_proposals/config/install/core.date_format.oe_call_proposals_date_long.yml
```diff
@@ -0,0 +1,7 @@
+langcode: en
+status: true
+dependencies: {  }
+id: oe_call_proposals_date_long
+label: 'Call for proposals long date'
+locked: false
+pattern: 'd F Y'
```
nagyad, line 7: Instead of repeating, why don't we add the formats to the main module?

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.info.yml
```diff
@@ -0,0 +1,13 @@
...
+package: OpenEuropa
+core: 8.x
+dependencies:
+  - extra_field:extra_field
+  - field_group:field_group
+  - oe_content:oe_content_call_proposals
+  - oe_theme:oe_theme_helper
+  - oe_theme:oe_theme_content_entity_contact
+  - oe_time_caching:oe_time_caching
+  - ui_patterns_field_group
```
nagyad: This is not good.

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
```diff
@@ -0,0 +1,82 @@
...
+  _oe_theme_content_call_proposals_strike_deadline($variables);
+}
+
+/**
+ * Add strike through to deadline date when call for proposals is close.
+ *
+ * @param array $variables
+ *   Element variables from preprocess.
+ */
+function _oe_theme_content_call_proposals_strike_deadline(&$variables): void {
```
nagyad: Missing parameter type.

modules/oe_theme_content_call_proposals/src/Plugin/ExtraField/Display/CallForProposalsStatusExtraField.php
```diff
@@ -0,0 +1,102 @@
...
+    $entity = CallForProposalsNodeWrapper::getInstance($entity);
+    $cacheable = CacheableMetadata::createFromRenderArray(['#cache' => ['contexts' => ['timezone']]]);
+
+    $status = $entity->getStatus();
+    // Set cache tags based on date.
+    if ($status === CallEntityWrapperInterface::STATUS_UPCOMING) {
+      $cacheable->addCacheTags($this->cacheTagGenerator->generateTags($entity->getOpeningDate()->getPhpDateTime()));
+    }
+    if ($status === CallEntityWrapperInterface::STATUS_OPEN
+      && !empty($entity->getDeadlineDate())) {
```
nagyad: Move this to single line please.

modules/oe_theme_content_call_proposals/src/Plugin/ExtraField/Display/CallForProposalsPublicationDetailsExtraField.php
```diff
@@ -0,0 +1,107 @@
...
+ * @ExtraFieldDisplay(
+ *   id = "oe_call_proposals_publication_details",
+ *   label = @Translation("Publication details"),
+ *   bundles = {
+ *     "node.oe_call_proposals",
+ *   },
+ *   visible = true
+ * )
+ */
+class CallForProposalsPublicationDetailsExtraField extends ExtraFieldDisplayFormattedBase implements ContainerFactoryPluginInterface {
```
nagyad: The class name gives me the impression this takes care of the whole Details area rendering but it's not.

modules/oe_theme_content_call_tenders/src/Plugin/PageHeaderMetadata/CallForTendersContentType.php
```diff
@@ -43,26 +43,6 @@ public function getMetadata(): array {
...
-      '#template' => '{{ summary|render|striptags("<strong><a><em>")|raw }}',
-      '#context' => [
-        'summary' => [
-          '#type' => 'processed_text',
-          '#text' => $summary->value,
-          '#format' => $summary->format,
-          '#langcode' => $summary->getLangcode(),
-        ],
-      ],
-    ];
```
nagyad, line 64: Why do we have Call for tenders changes here? Out of scope.

modules/oe_theme_content_call_tenders/src/Plugin/ExtraField/Display/CallForTendersStatusExtraField.php
```diff
@@ -82,15 +82,16 @@ public function viewElements(ContentEntityInterface $entity) {
...
+    if ($status === CallEntityWrapperInterface::STATUS_UPCOMING) {
       $cacheable->addCacheTags($this->cacheTagGenerator->generateTags($entity->getOpeningDate()->getPhpDateTime()));
     }
-    if ($status === CallForTendersNodeWrapperInterface::STATUS_OPEN) {
+    if ($status === CallEntityWrapperInterface::STATUS_OPEN) {
       $cacheable->addCacheTags($this->cacheTagGenerator->generateTags($entity->getDeadlineDate()->getPhpDateTime()));
     }
     $build = [
-      '#theme' => 'oe_theme_content_call_tenders_status',
+      '#theme' => 'oe_theme_helper_call_status',
```
nagyad, line 92: Why do we have Call for tenders changes here? Out of scope.

modules/oe_theme_content_call_tenders/oe_theme_content_call_tenders.module
```diff
@@ -16,12 +16,6 @@ use Drupal\oe_content_call_tenders\CallForTendersNodeWrapper;
  */
 function oe_theme_content_call_tenders_theme() {
   return [
-    'oe_theme_content_call_tenders_status' => [
-      'variables' => [
-        'label' => '',
-        'name' => '',
-      ],
-    ],
```
nagyad, line 24: Why do we have Call for tenders changes here? Out of scope.

modules/oe_theme_helper/translations/oe_theme_helper-bg.po
```diff
@@ -143,15 +145,9 @@ msgstr "Данни"
...
 
-msgid "Contact organisation"
-msgstr "Връзка с организацията"
-
 msgid "Publication"
 msgstr "Публикация"
 
-msgid "Publication date"
-msgstr "Дата на публикуване"
-
```
nagyad, line 154: Why are these removed?

tests/Functional/ContentRenderTestBase.php
```diff
@@ -205,4 +206,118 @@ protected function createContactEntity(string $name, string $bundle, int $status
...
 
+  /**
+   * Asserts Contact entity field rendering.
+   *
+   * @param \Behat\Mink\Element\NodeElement $element
+   *   Rendered element.
+   * @param string $name
+   *   Name of the Contact entity.
+   */
+  protected function assertContactEntityDefaultDisplay(NodeElement $element, string $name):void {
```
nagyad: Missing space at $name):void

modules/oe_theme_content_call_proposals/config/install/core.entity_view_display.node.oe_call_proposals.full.yml
```diff
@@ -0,0 +1,242 @@
...
+      children:
+        - oe_call_proposals_contact
+      parent_name: group_in_page_navigation
+      weight: 23
+      format_type: oe_theme_helper_in_page_navigation_item
+      region: content
+      format_settings:
+        id: ''
+        classes: ''
+      label: Contact
```
nagyad, line 90: This should be Contacts, it expects more items. You do it already with Documents.

tests/Functional/ContentRenderTestBase.php
```diff
@@ -205,4 +206,118 @@ protected function createContactEntity(string $name, string $bundle, int $status
...
+    $contacts_html = $element->getHtml();
+    $field_list_assert = new FieldListAssert();
+    $contact_expected_values = [
+      'items' => [
+        [
+          'label' => 'Organisation',
+          'body' => "Organisation $name",
+        ], [
+          'label' => 'Website',
+          'body' => "http://www.example.com/website_$name",
```
nagyad, line 235: I can add node as well and this is what I get:
image

tests/Functional/ContentRenderTestBase.php
```diff
@@ -205,4 +206,118 @@ protected function createContactEntity(string $name, string $bundle, int $status
...
+   *   Name of the Contact entity.
+   */
+  protected function assertContactEntityDefaultDisplay(NodeElement $element, string $name):void {
+    $contact_name = $element->findAll('css', 'h3');
+    $this->assertCount(1, $contact_name);
+    $this->assertEquals($name, $contact_name[0]->getText());
+
+    $contact_body = $element->findAll('css', '.ecl-editor');
+    $this->assertCount(1, $contact_body);
+    $this->assertEquals("Body text $name", $contact_body[0]->getText());
```
nagyad, line 224: If there is no image the text of the body should be corresponding to the width of the body, right?
image

nagyad (changes requested):

modules/oe_theme_content_call_proposals/config/install/core.date_format.oe_call_proposals_date_long.yml
yenyasinn: I don't mind. 

We can call it `core.date_format.date_long` and move to oe_theme_helper. It also can be used in oe_theme_content_call_tenders (instead of core.date_format.oe_call_tenders_date_long) and post_update in oe_theme_content_call_tenders should be provided to update core.entity_view_display.node.oe_call_tenders.full/teaser.

In this case we should move core.date_format.oe_call_tenders_date_full_timezone to core.date_format.date_full_timezone and use it for Call for tenders/Call for proposals both.

I only have a concern about configurations of core.date_format.oe_call_tenders_date_long/core.date_format.oe_call_tenders_date_full_timezone on the existing sites if they were changed. Should we remove already existing core.date_format.oe_call_tenders_date_long/core.date_format.oe_call_tenders_date_full_timezone or not?

@ademarco @nagyad do you have thoughts about it?

modules/oe_theme_content_call_proposals/config/install/core.entity_view_display.node.oe_call_proposals.full.yml
yenyasinn: Field has label "Contacts" modules/oe_content_call_proposals/config/install/field.field.node.oe_call_proposals.oe_call_proposals_contact.yml but front-end label is "Contact" - it is determined in spec https://webgate.ec.europa.eu/fpfis/wikis/display/EWPP/Call+for+Proposals+Content+Type+Specification+-+Version+1.0.0

modules/oe_theme_content_call_tenders/src/Plugin/ExtraField/Display/CallForTendersStatusExtraField.php
yenyasinn: It was discussed with @ademarco to provide common way for Call for tenders/Call for proposals to print status.

modules/oe_theme_helper/translations/oe_theme_helper-bg.po
yenyasinn: These strings are duplicated.

tests/Functional/ContentRenderTestBase.php
yenyasinn: Based on the implementation of Contact entity full view I would say no. See https://github.com/openeuropa/oe_theme/blob/2.x/templates/oe_contact/oe-contact--full.html.twig - width of column is the same despite of the image or text fields.

If it is wrong behaviour I think separate bug should be created.

tests/Functional/ContentRenderTestBase.php
yenyasinn: Yes, it is correct behaviour. Website can be external or internal link.

Spec https://webgate.ec.europa.eu/fpfis/wikis/display/EWPP/Contacts+-+Version+1.0.0 doesn't determine that website is external link only. If it is wrong let's create a separate bug.

modules/oe_theme_content_call_tenders/src/Plugin/PageHeaderMetadata/CallForTendersContentType.php
yenyasinn: We don't have ticket "To remove useless code from Page header in Call for tenders"...

modules/oe_theme_content_call_proposals/config/install/core.date_format.oe_call_proposals_date_long.yml
yenyasinn: BTW, format "j F Y" is used in Event, News, Publication. So we need some common way how to work with date formats - we will create separate date formats in the modules or need to update date formats in all content types.

modules/oe_theme_content_call_proposals/config/install/core.entity_view_display.node.oe_call_proposals.full.yml
yenyasinn: The same case for Publication - https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?pageId=404786614.

modules/oe_theme_content_call_proposals/config/install/core.date_format.oe_call_proposals_date_long.yml
nagyad: Let's do it in a followup then. I will create a ticket for this.

nagyad (dismissed):

drishu (dismissed):

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.info.yml
```diff
@@ -0,0 +1,13 @@
+name: OpenEuropa Theme Call for proposals Content
+type: module
+description: Companion module for the OE Call for proposals Content module
+package: OpenEuropa
+core: 8.x
+dependencies:
```
imanoleguskiza, line 6: I noticed we are missing the definition of all the configuration we ship in most of the submodules. Lets add it here already and add a follow up to add it to the rest.

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
```diff
@@ -0,0 +1,82 @@
...
+  _oe_theme_content_call_proposals_strike_deadline($variables);
+}
+
+/**
+ * Add strike through to deadline date when call for proposals is close.
+ *
+ * @param array $variables
+ *   Element variables from preprocess.
+ */
+function _oe_theme_content_call_proposals_strike_deadline(array &$variables): void {
```
imanoleguskiza: This function is exactly the same as _oe_theme_content_call_tenders_strike_deadline. We should aim to reduce duplicating all this code. Lets open a follow up to assess what we can abstract from both submodules.

modules/oe_theme_content_call_proposals/src/Plugin/ExtraField/Display/CallForProposalsStatusExtraField.php
```diff
@@ -0,0 +1,101 @@
...
+ * @ExtraFieldDisplay(
+ *   id = "oe_call_proposals_status",
+ *   label = @Translation("Status"),
+ *   bundles = {
+ *     "node.oe_call_proposals",
+ *   },
+ *   visible = true
+ * )
+ */
+class CallForProposalsStatusExtraField extends ExtraFieldDisplayFormattedBase implements ContainerFactoryPluginInterface {
```
imanoleguskiza, line 29: This class is another candidate to be abstracted as it is exactly the same as CallForTendersStatusExtraField. Lets include it in the follow up to assess duplication mentioned before.

tests/Functional/ContentRenderTestBase.php
```diff
@@ -205,4 +206,118 @@ protected function createContactEntity(string $name, string $bundle, int $status
...
 
+  /**
+   * Asserts Contact entity field rendering.
+   *
+   * @param \Behat\Mink\Element\NodeElement $element
+   *   Rendered element.
+   * @param string $name
+   *   Name of the Contact entity.
+   */
+  protected function assertContactEntityDefaultDisplay(NodeElement $element, string $name): void {
```
imanoleguskiza, line 217: ContentProjectRenderTest and OrganisationRenderTest have methods to asser the rendering of contact items. Could we check if we can abstract this methods and have a unified way of asserting this? If so, open a follow up.

tests/Functional/ContentRenderTestBase.php
```diff
@@ -205,4 +206,118 @@ protected function createContactEntity(string $name, string $bundle, int $status
...
+   *
+   * @param \Behat\Mink\Element\NodeElement $element
+   *   Field group content.
+   * @param string $title
+   *   Expected title.
+   * @param string $id
+   *   Expected id.
+   */
+  protected function assertContentHeader(NodeElement $element, string $title, string $id = ''): void {
+    $header = $element->find('css', 'h2.ecl-u-type-heading-2');
```
imanoleguskiza, line 282: A similar function exists on ContentOrganisationRenderTest, lets create a follow up to reduce code duplication (this follow up can include the comment above as well).

tests/Functional/ContentRenderTestBase.php
```diff
@@ -205,4 +206,118 @@ protected function createContactEntity(string $name, string $bundle, int $status
...
+   * @param \Behat\Mink\Element\NodeElement $element
+   *   Rendered element.
+   * @param string $title
+   *   Link title.
+   * @param string $href
+   *   Link URL.
+   * @param bool $is_external
+   *   Defines whether it is extrernal link or internal.
+   */
+  protected function assertLinkIcon(NodeElement $element, string $title, string $href, bool $is_external = TRUE): void {
```
imanoleguskiza, line 301: This to me looks like a pattern assertion. Although we don't have a link with icon pattern as such, the final rendered result is very standarized so we may benefit from having it as a pattern assertion. I leave it as nice to have (maybe in a follow up)

tests/Functional/ContentCallForProposalsRenderTest.php
```diff
@@ -0,0 +1,379 @@
...
+      'list' => [
+        ['label' => 'Details', 'href' => '#details'],
+      ],
+    ];
+    $inpage_nav_assert->assertPattern($inpage_nav_expected_values, $navigation->getOuterHtml());
+
+    // Assert content part.
+    $wrapper = $this->assertSession()->elementExists('css', '.ecl-row.ecl-u-mt-l');
+    $content = $this->assertSession()->elementExists('css', '.ecl-col-lg-9', $wrapper);
+    $this->assertSession()->elementsCount('css', '.ecl-col-lg-9', 1);
```
imanoleguskiza: Do we need this assertion here? And if there is only one, why do we need the $wrapper to get the $content?

tests/Functional/ContentCallForProposalsRenderTest.php
```diff
@@ -0,0 +1,379 @@
...
+    $field_list_assert->assertPattern($details_expected_values, $content_items[0]->getHtml());
+
+    // Assert Deadline model and Deadline dates fields.
+    $deadline_date1 = (clone $static_time)->modify('- 3 days');
+    $node->set('oe_call_proposals_model', 'single_stage');
+    $node->set('oe_call_proposals_deadline', $deadline_date1->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT));
+    $node->save();
+    $this->drupalGet($node->toUrl());
+
+    $deadline_date1->setTimeZone(new \DateTimeZone('Australia/Sydney'));
```
imanoleguskiza: Why do we need to set the timezone here?

tests/Functional/ContentCallForProposalsRenderTest.php
```diff
@@ -0,0 +1,379 @@
...
+      'label' => 'Status',
+      'body' => 'Open',
+    ];
+    $details_expected_values['items'][4] = [
+      'label' => 'Deadline model',
+      'body' => 'Two-stage',
+    ];
+    $expected_deadline_dates = $deadline_date1->format('d F Y, H:i (T)') . $deadline_date2->format('d F Y, H:i (T)');
+    if (version_compare(PHP_VERSION, '7.3') < 0) {
+      $expected_deadline_dates = $deadline_date1->format('d F Y, H:i (T)') . "\n" . $deadline_date2->format('d F Y, H:i (T)');
```
imanoleguskiza: This is quite strange, can we explain why this happens?

imanoleguskiza (changes requested):

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.info.yml
yenyasinn: Fixed. Follow up https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-527 has been created.

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
yenyasinn: Follow up https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-528 has been created.

modules/oe_theme_content_call_proposals/src/Plugin/ExtraField/Display/CallForProposalsStatusExtraField.php
yenyasinn: Added to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-528

tests/Functional/ContentCallForProposalsRenderTest.php
yenyasinn: By default Australia/Sydney timezone is used for testing. See install.core.inc.
```
  if ($install_state['interactive'] && drupal_valid_test_ua()) {
    // Set the default timezone. While this doesn't cause any tests to fail, PHP
    // complains if 'date.timezone' is not set in php.ini. The Australia/Sydney
    // timezone is chosen so all tests are run using an edge case scenario
    // (UTC+10  and DST). This choice is made to prevent timezone related
    // regressions and reduce the fragility of the testing system in general.
    date_default_timezone_set('Australia/Sydney');
  }
```
Without setting timezone `$deadline_date1->format('d F Y, H:i (T)')` will print date in UTC timezone but fields are rendered in Australia/Sydney timezone.

tests/Functional/ContentRenderTestBase.php
yenyasinn: Implementation of assertContactRendering() in ContentProjectRenderTest and OrganisationRenderTest is very similar to assertContactEntityDefaultDisplay() but doesn't use PatternAssertions. So assertContactEntityDefaultDisplay() that is implemented in base class is a unified way to assert contact in functional tests. There is https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-402 for tests refactoring.

tests/Functional/ContentRenderTestBase.php
yenyasinn: It is part of ticket https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-402 and it was implemented there https://github.com/openeuropa/oe_theme/pull/629/files#diff-9710e2dff97ec2899daab0082c8735d318016ab06c05b025829a19cd2988f3f3.

tests/Functional/ContentRenderTestBase.php
yenyasinn: It's good question about can we create asserts based on BasePatternAssert for the code that isn't a pattern?
We have talked about creation of classes that can be used in Kernel and Functional tests both to handle specific pieces of rendering, like this one or Contact rendering. We have stopped on the idea of using traits for it. This work has been stopped in favour of pattern asserts.

tests/Functional/ContentCallForProposalsRenderTest.php
yenyasinn: I've faced with this issue a few times. As a generic fix https://github.com/openeuropa/oe_theme/pull/629/commits/504bac8087d10fe7811c0e4c21f62788dda63c4e can be used, but it is in EWPP-402.

tests/Functional/ContentRenderTestBase.php
imanoleguskiza: Since we already have 402 to refactor the tests we can take care of it there.

tests/Functional/ContentCallForProposalsRenderTest.php
imanoleguskiza: I see, the same would happen for the other dates we are checking but this is the only place where we check the hour and timezone so hence the solution. However, there is no need to actually format any date to test this, we know what the result needs to be (we froze the time for a reason) so instead of formatting any date we should simply pass the string as the expected result (we do it like this consistently across all the other tests

tests/Functional/ContentCallForProposalsRenderTest.php
imanoleguskiza: But the solution you send is a QOL improvement so that normalizes whitespace when using the text() method on a crawler. That doesn't explain why we need different assertions for different PHP versions. I don't see this issue anywhere else, not even in the examples on your generic fix. 
As it is, this could be a problem with the symfony crawler, but it could be we are rendering different things on different php versions, and that is something we should fix.

tests/Functional/ContentCallForProposalsRenderTest.php
yenyasinn: Hardcoded dates are used now.

tests/Functional/ContentCallForProposalsRenderTest.php
yenyasinn: Explanation is in the ticket EWPP-437. 

In the EWPP-402 in some cases issue was resolved by updating templates - new line characters have been removed for templates in oe_theme, in some cases patch helped.

I would prefer to keep current solution now or please suggest another option.

imanoleguskiza (dismissed):

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
```diff
@@ -0,0 +1,82 @@
...
+      $build[$name]['#title'] = $label_override;
+    }
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_content_call_proposals_preprocess_field__node__oe_call_proposals_deadline__oe_call_proposals__full(&$variables, $hook) {
+  _oe_theme_content_call_proposals_strike_deadline($variables);
```
nagyad: Let's remove this. See https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-438?focusedCommentId=4394530&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4394530

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.module
nagyad: This can be removed.

tests/Functional/ContentCallForProposalsRenderTest.php
```diff
@@ -0,0 +1,374 @@
...
+      'label' => 'Deadline model',
+      'body' => 'Single-stage',
+    ];
+    $details_expected_values['items'][5] = [
+      'label' => 'Deadline date',
+      'body' => '15 February 2020, 01:00 (AEDT)',
+    ];
+    $field_list_assert->assertPattern($details_expected_values, $content_items[0]->getHtml());
+
+    $deadline_selector = '//*[text() = "Deadline date"]/following-sibling::dd[1]/div[@class="ecl-u-type-strike"]';
```
nagyad: We can remove the strike from here.

nagyad (changes requested):

nagyad (approved):

imanoleguskiza (approved):

PR #642, opened 2020-11-10
EWPP-131: Theme EC and EU footers with latest ECL Twig components

drishu:
### Description

Theme EC and EU footers with latest ECL Twig components

### Change log

- Added:
   - ec-component-footer-core and ec-component-footer-standardised ecl components
  - templates and preprocess for each footer variant
  - patch to fix icon size in footer-standardized
- Removed:
  - old footer templates


oe_theme.theme
```diff
@@ -210,153 +210,352 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+          'path' => $variables['processed']['site_path'],
+        ]
       ],
-    ];
-  }
-
-  $variables['custom_footer'] = _oe_theme_preprocess_oe_corporate_blocks_site_specific_footer($variables);
+      'description' => $variables['processed']['site_owner'],
+      'section_id' => 1,
+      'type' => 'site_name',
```
imanoleguskiza: The type variable does nothing, we can simlpy ignore it (unless there is a reason I am not aware of)

oe_theme.theme
```diff
@@ -210,153 +210,352 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+
+/**
+ * Process corporate_footer data to match ecl component expected format.
+ *
+ * @param array $variables
+ *   Set of available variables.
+ *
+ * @see hook_preprocess_oe_corporate_blocks_eu_footer().
+ */
+function _oe_theme_preprocess_oe_corporate_blocks_eu_footer__standardised(array &$variables): void {
```
imanoleguskiza: Same comments as the one on _oe_theme_preprocess_oe_corporate_blocks_ec_footer__standardised: type is not used on the standarised template.

oe_theme.theme
```diff
@@ -210,153 +210,352 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+  $variables['processed'] = [];
   $corporate_footer = $variables['corporate_footer'];
-
-  // Create the section for 'About EC'.
-  $about_ec = $corporate_footer['about_ec'];
-  $sections_about_ec = [
-    'title' => $about_ec['title'],
-    'links' => [],
+  $site_name = $corporate_footer['site_name'];
+  $site_name_link = [
```
imanoleguskiza: This whole section is very confusing, we have a site_name_link but also a site_name and a site_path :( Maybe we could rename site_name_link to something link: ec_site_link or something

tests/Behat/OeThemeTestContext.php
```diff
@@ -343,23 +343,23 @@ public function setCorporateLibrary(string $component_library): void {
...
+    $this->assertSession()->elementExists('css', 'footer.ecl-footer-core');
 
-    // Get the actual first footer title.
     $page = $this->getSession()->getPage();
-    $actual_title = trim($page->find('css', 'h1.ecl-footer__section-title')->getText());
+    $logo = $page->findAll('css', 'footer.ecl-footer-core img.ecl-footer-core__logo-image-desktop');
 
     // Assert presence or absence of given footer block.
-    $title_found = $actual_title === $expected_title[$component_library];
+    $title_found = count($logo) === $expected[$component_library];
```
imanoleguskiza: $logo_found

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
```diff
@@ -5,316 +5,264 @@
...
 
-    $actual = $first_column->filter('ul li:nth-child(1) > a');
-    $this->assertEquals($test_data['about_ec_links'][0]['href'], $actual->attr('href'));
-    $this->assertEquals($test_data['about_ec_links'][0]['label'], $actual->text());
+    $section = $crawler->filter('footer.ecl-footer-core section.ecl-footer-core__section2');
 
-    $actual = $first_column->filter('ul li:nth-child(2) > a');
-    $this->assertEquals($test_data['about_ec_links'][1]['href'], $actual->attr('href'));
-    $this->assertEquals($test_data['about_ec_links'][1]['label'], $actual->text());
+    $actual = $section->filter('ul li:nth-child(1) > a');
```
imanoleguskiza: All link lists are tested the same way but we repeat the same code over and over again, we could use iterations or helper methods to make it less cluttered.

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
```diff
@@ -5,316 +5,264 @@
...
+    $this->assertEquals($test_data['legal_navigation'][1]['label'], $actual->text());
+  }
 
-    $html = (string) $this->container->get('renderer')->renderRoot($render);
+  /**
+   * Test European Commission footer standardised block rendering.
+   */
+  public function testEcFooterStandardisedBlockRendering(): void {
+    $this->configFactory->getEditable('oe_theme.settings')->set('branding', 'standardised')->save();
+    $test_data = [];
```
imanoleguskiza: Why not use the test_data on the assertions? In similar way, there is a lot of repeated code blocks in this tests.

imanoleguskiza (changes requested):

patches/@ecl-twig/ec-component-footer-standardised+2.33.2.patch
```diff
@@ -0,0 +1,11 @@
+patch-package
+--- a/node_modules/@ecl-twig/ec-component-footer-standardised/ecl-footer-standardised.html.twig
++++ b/node_modules/@ecl-twig/ec-component-footer-standardised/ecl-footer-standardised.html.twig
+@@ -135,6 +135,7 @@
+           <li class="ecl-footer-standardised__list-item">
+              {% include '@ecl-twig/ec-component-link/ecl-link.html.twig' with _item|merge({
+                 link: _item.link|default({})|merge({ type: 'standalone' }),
++                icon: _item.icon|default({})|merge({ size: 'xs' }),
```
sergepavle, line 8: According to the normal practice we have to create ticket on FRONT project board for considering or applying this patch in ECL-TWIG project.

oe_theme.theme
```diff
@@ -210,153 +210,300 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+  $variables['processed'] = [];
   $corporate_footer = $variables['corporate_footer'];
-
-  // Create the section for 'About EC'.
-  $about_ec = $corporate_footer['about_ec'];
-  $sections_about_ec = [
-    'title' => $about_ec['title'],
-    'links' => [],
+  $site_name = $corporate_footer['site_name'];
+  $ec_site_link = [
```
sergepavle: We are using this variable only once. Looks like we can set 'ec_site_link' value directly.

oe_theme.theme
```diff
@@ -210,153 +210,300 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
           'type' => 'ui',
           'name' => 'external',
         ],
       ];
     }
+    $service_navigation[] = $service_navigation_link;
   }
+  $variables['processed']['service_navigation'] = $service_navigation;
+
+  $ecl_branding = theme_get_setting('branding') ?? 'core';
```
sergepavle: I think we can use $variables['ecl_branding'] . Value should be available.

oe_theme.theme
```diff
@@ -210,153 +210,300 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
  * Implements hook_preprocess_oe_corporate_blocks_ec_footer().
  */
 function oe_theme_preprocess_oe_corporate_blocks_ec_footer(array &$variables): void {
+  $variables['processed'] = [];
```
sergepavle: Why do we need an additional array element? Why we cannot just override elements (class navigation, legal navigation etc.).

oe_theme.theme
```diff
@@ -210,153 +210,300 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+  $preprocess($variables);
+}
 
-  // Create the section for 'About EU'.
-  $about_eu = $corporate_footer['about_eu'];
-  $sections_about_eu = [
-    'title' => $about_eu['title'],
-    'links' => [],
+/**
+ * Process EC footer data to match the expected ECL Core format.
```
sergepavle: core branding

oe_theme.theme
```diff
@@ -210,153 +210,300 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
 
-  $variables['custom_footer'] = _oe_theme_preprocess_oe_corporate_blocks_site_specific_footer($variables);
+  $ecl_branding = theme_get_setting('branding') ?? 'core';
+  $preprocess = '_oe_theme_preprocess_oe_corporate_blocks_eu_footer__' . $ecl_branding;
+  $preprocess($variables);
 }
 
 /**
- * Implements hook_preprocess_oe_corporate_blocks_eu_footer().
+ * Process EU footer data to match the expected ECL Core format.
```
sergepavle: core branding

oe_theme.theme
```diff
@@ -210,153 +210,300 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
-      ],
-    ];
-  }
+  $variables['processed']['contacts'] = _oe_theme_preprocess_footer_links($corporate_footer['contact']);
+  $variables['processed']['social_media'] = _oe_theme_preprocess_footer_links($corporate_footer['social_media']);
+  $variables['processed']['institution_links'] = _oe_theme_preprocess_footer_links($corporate_footer['institution_links']);
+  $variables['processed']['legal_links'] = _oe_theme_preprocess_footer_links($corporate_footer['legal_links']);
 
-  $variables['custom_footer'] = _oe_theme_preprocess_oe_corporate_blocks_site_specific_footer($variables);
+  $ecl_branding = theme_get_setting('branding') ?? 'core';
```
sergepavle: I think we can use $variables['ecl_branding'] . Value should be available.

oe_theme.theme
```diff
@@ -210,153 +210,300 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+    [
+      'title' => $variables['corporate_footer']['institution_links_title'],
+      'links' => $variables['processed']['institution_links'],
+    ],
+  ];
 }
 
 /**
- * Prepare data for site specific footer preprocess.
+ * Process EU footer data to match the expected ECL Standardised format.
```
sergepavle: branding

sergepavle (changes requested):

patches/@ecl-twig/ec-component-footer-standardised+2.33.2.patch
drishu: I will look into it but without this the icon is way to big

oe_theme.theme
drishu: fixed

oe_theme.theme
drishu: changed, but I have a concern, (the reason why I did it like this in the first place) I tested the upgrade from 2.x branch to this one, and it could not find the right template.
Granted the code changed quite a bit since then, but I was not able to retest this scenario, I tried but updb was with errors:
`root@dbe60b6d5f48:/var/www/html# ./vendor/bin/drush updb -y
 [error]   (Currently using Missing or invalid module Array) `
`[notice] Module file_link has an entry in the system.schema key/value storage, but is missing from your site. <a href=":url">More information about this error</a>.
>  [notice] Update started: oe_corporate_blocks_post_update_30001
>  [error]  The module file_link does not exist. 
>  [error]  Update failed: oe_corporate_blocks_post_update_30001 
 [error]  Update aborted by: oe_corporate_blocks_post_update_30001 
 [error]  Finished performing updates. `

maybe I did something wrong with my upgrade procedure, **please confirm the upgrade scenario works for you.**

oe_theme.theme
drishu: reworked

oe_theme.theme
drishu: done

oe_theme.theme
drishu: done

oe_theme.theme
drishu: done but see https://github.com/openeuropa/oe_theme/pull/642#discussion_r524418251

oe_theme.theme
drishu: done

patches/@ecl-twig/ec-component-footer-standardised+2.33.2.patch
drishu: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-1789

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+}
 
-  foreach ($about_eu['items'] as $item) {
-    $sections_about_eu['links'][] = [
-      'link' => [
-        'label' => $item['label'],
-        'path' => $item['href'],
-        'icon_position' => 'after',
+/**
+ * Process EC footer data to match the expected ECL Standardised format.
```
sergepavle: Branding

sergepavle (dismissed):

ademarco (changes requested):
@sergepavle have you tested an upgrade path scenario? I get this:

image

Also I'm forced to clear the cache in order for `oe_corporate_site_info.site_information` service to be discovered, maybe this is to be fixed in oe_corporate_blocks

tests/Kernel/CorporateBlocks/CorporateBlocksTestBase.php
```diff
@@ -5,19 +5,161 @@
...
+   */
+  protected $configFactory;
+
   /**
    * Modules to enable.
    *
    * @var array
    */
   protected static $modules = [
+    'oe_content',
```
ademarco: This shouldn't be a dependency to test the footer.

tests/Kernel/CorporateBlocks/CorporateBlocksTestBase.php
```diff
@@ -5,19 +5,161 @@
...
+    ]);
+
+    $this->installEntitySchema('rdf_entity');
+    $this->installEntitySchema('skos_concept');
+    $this->installEntitySchema('skos_concept_scheme');
+
+    $graphs = [
+      'corporate_body' => 'http://publications.europa.eu/resource/authority/corporate-body',
+    ];
+    $this->container->get('rdf_skos.skos_graph_configurator')->addGraphs($graphs);
```
ademarco: Why do we need this here? If we need it it should be done by the application being tested, not by the test, right?

tests/Kernel/CorporateBlocks/CorporateBlocksTestBase.php
```diff
@@ -5,19 +5,161 @@
...
+   *
+   * @param string $type
+   *   The type of block, ec or eu.
+   * @param array $test_data
+   *   The test data for config and assertion.
+   *
+   * @return string
+   *   The rendered HTML.
+   */
+  protected function renderCorporateBlocksFooter(string $type, array &$test_data): string {
```
ademarco: This method render the block and fetches the test data, isn't it more clear for one method to do one thing only? Let's simply:

```php
    $test_data = $this->getFixtureContent("{$type}_footer.yml");
    $html = $this->renderCorporateBlocksFooter($type, $test_data);
```

ecl-builder.config.js
```diff
@@ -43,6 +43,8 @@ module.exports = {
     { from: path.resolve(nodeModules, '@ecl-twig'), patterns: 'ec-component-fact-figures/*.twig', to: path.resolve(outputFolder, 'components') },
     { from: path.resolve(nodeModules, '@ecl-twig'), patterns: 'ec-component-file/*.twig', to: path.resolve(outputFolder, 'components') },
     { from: path.resolve(nodeModules, '@ecl-twig'), patterns: 'ec-component-footer/*.twig', to: path.resolve(outputFolder, 'components') },
```
ademarco: Let's remove this if we don't use it anymore.

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
 function oe_theme_preprocess_oe_corporate_blocks_ec_footer(array &$variables): void {
```
ademarco: We should move all this in the Twig templates, we can easily do everything we need using Twig's filters, Drupal core Twig functions and macros. I'll comment below to highlight each scenario.

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
   $corporate_footer = $variables['corporate_footer'];
 
-  // Create the section for 'About EC'.
-  $about_ec = $corporate_footer['about_ec'];
-  $sections_about_ec = [
-    'title' => $about_ec['title'],
-    'links' => [],
+  $corporate_footer['instance_name'] = \Drupal::config('system.site')->get('name');
+  $corporate_footer['instance_path'] = Url::fromRoute('<front>', [], ['absolute' => TRUE])->toString();
```
ademarco: Drupal core provides a Twig function for this `{{ url(...) }}`. Check `\Drupal\Core\Template\TwigExtension::getFunctions()`. Also, the Twig function will make sure cache bubbles up correctly.

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
 function oe_theme_preprocess_oe_corporate_blocks_ec_footer(array &$variables): void {
   $corporate_footer = $variables['corporate_footer'];
 
-  // Create the section for 'About EC'.
-  $about_ec = $corporate_footer['about_ec'];
-  $sections_about_ec = [
-    'title' => $about_ec['title'],
-    'links' => [],
+  $corporate_footer['instance_name'] = \Drupal::config('system.site')->get('name');
```
ademarco: I understand that `site_name` is already taken by the configuration:

image

Let's change it there, so we can call this `site_name`. Let's call that one `corporate_site_link` in both EC and EU, since that's what it indicates.

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+/**
+ * Processes a series of footer links to make it compatible with ECL formatting.
+ *
+ * @param array $links
+ *   Set of links to be processed.
+ *
+ * @return array
+ *  Set of processed links.
+ */
+function _oe_theme_preprocess_footer_links(array $links): array {
```
ademarco: We can turn this into a Twig macro and use it directly in the Twig templates.

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
-    'links' => [],
+/**
+ * Process EC footer data to match the expected ECL core branding.
+ *
+ * @param array $variables
+ *   Set of available variables.
+ *
+ * @see hook_preprocess_oe_corporate_blocks_ec_footer().
+ */
+function _oe_theme_preprocess_oe_corporate_blocks_ec_footer__core(array &$variables): void {
```
ademarco: Let's move this to the Twig template.

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
-        'icon_position' => 'after',
+/**
+ * Process EC footer data to match the expected ECL Standardised format.
+ *
+ * @param array $variables
+ *   Set of available variables.
+ *
+ * @see hook_preprocess_oe_corporate_blocks_ec_footer().
+ */
+function _oe_theme_preprocess_oe_corporate_blocks_ec_footer__standardised(array &$variables): void {
```
ademarco: Let's move this to the Twig template. We could also resume the idea of having template suggestions based on branding value, if it makes things easier.

oe_theme.theme
ademarco: We also need to make sure we add the system.site as a cacheable dependency, so that the cache can be invalidated when the site name changes.

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
-    $sections_social_media,
-    $sections_about_eu,
+/**
+ * Implements hook_preprocess_oe_corporate_blocks_eu_footer().
+ */
+function oe_theme_preprocess_oe_corporate_blocks_eu_footer(array &$variables): void {
+  $corporate_footer = $variables['corporate_footer'];
+  $corporate_footer['instance_name'] = \Drupal::config('system.site')->get('name');
+  $corporate_footer['instance_path'] = Url::fromRoute('<front>', [], ['absolute' => TRUE])->toString();
+  $corporate_footer['site_owner'] = t('This site is managed by the @name', ['@name' => $variables['site_owner']]);
```
ademarco: This string is provided by the oe_corporate_blocks module, let's just reuse that one, so when we change it there and update its translations, we don't need to change it in here too.

It's this bit here:

```twig
<span class="footer__content-owner-details">{{ 'This site is managed by the @name'|t({'@name': site_owner}) }}</span>
```

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
-    'links' => [],
+  $corporate_footer['instance_name'] = \Drupal::config('system.site')->get('name');
+  $corporate_footer['instance_path'] = Url::fromRoute('<front>', [], ['absolute' => TRUE])->toString();
+  $corporate_footer['ec_site_link'] = [
+    'link' => [
+      'label' => $corporate_footer['site_name']['label'],
+      'path' => $corporate_footer['site_name']['href'],
+    ],
   ];
+  $corporate_footer['site_owner'] = t('This site is managed by the @name', ['@name' => $variables['site_owner']]);
```
ademarco: This string is provided by the oe_corporate_blocks module, let's just reuse that one, so when we change it there and update its translations, we don't need to change it in here too.

It's this bit here:

```twig
<span class="footer__content-owner-details">{{ 'This site is managed by the @name'|t({'@name': site_owner}) }}</span>
```

oe_theme.theme
```diff
@@ -212,151 +212,294 @@ function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void
...
+ * Process EU footer data to match the expected ECL core branding.
+ *
+ * @param array $variables
+ *   Set of available variables.
+ *
+ * @see hook_preprocess_oe_corporate_blocks_eu_footer().
  */
-function oe_theme_preprocess_oe_corporate_blocks_eu_footer(array &$variables): void {
-  $variables['custom_footer'] = _oe_theme_preprocess_oe_corporate_blocks_site_specific_footer($variables);
+function _oe_theme_preprocess_oe_corporate_blocks_eu_footer__core(array &$variables): void {
```
ademarco: As per EC, let's move this in Twig templates.

tests/Kernel/CorporateBlocks/CorporateBlocksTestBase.php
```diff
@@ -5,19 +5,161 @@
...
+  protected function assertLinkList(Crawler $section, string $branding, array $expected): void {
+    $actual = $section->filter('ul li:nth-child(1) > a');
+    $this->assertEquals($expected['0']['href'], $actual->attr('href'));
+    $this->assertEquals($expected['0']['label'], $actual->text());
+    $this->assertEquals("ecl-link ecl-link--standalone ecl-footer-{$branding}__link", $actual->attr('class'));
+
+    $actual = $section->filter('ul li:nth-child(2) > a');
+    $this->assertEquals($expected['1']['href'], $actual->attr('href'));
+    $this->assertEquals($expected['1']['label'], $actual->text());
+    $this->assertEquals("ecl-link ecl-link--standalone ecl-footer-{$branding}__link", $actual->attr('class'));
```
ademarco: This function only works if we assign to each section two links, so what if we want to test that, if we remove a link from config, cache is invalidated and we now show 1 link only?

Each method/function shouldn't be aware of anything regarding its caller, let's keep this as a general principle. If you want to have a link assertion method, make sure that it asserts only one link, so that the caller can have full control.

tests/Kernel/CorporateBlocks/CorporateBlocksTestBase.php
```diff
@@ -5,19 +5,161 @@
...
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   Crawler containing the DOM nodes.
+   * @param string $branding
+   *   Ecl branding, core/standardised.
+   * @param int $expected_section_count
+   *   The number of expected sections.
+   */
+  protected function assertFooterPresence(Crawler $crawler, string $branding, int $expected_section_count): void {
+    $actual = $crawler->filter("footer.ecl-footer-{$branding}");
+    $this->assertCount(1, $actual);
```
ademarco: When we are creating assertion methods, let's always add a message that explains why something failed. If we keep the above we will get "0 expected to be 1". Something like:

```php
     $this->assertCount(1, $actual, "Footer with {$branding} branding not found.");
```

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
```diff
@@ -5,316 +5,182 @@
...
-    $this->assertEquals($custom_footer_data['oe_corporate_blocks.footer_link.general.legal_notice']['url'], $other_link2->attr('href'));
-    $this->assertEquals($custom_footer_data['oe_corporate_blocks.footer_link.general.legal_notice']['label'], preg_replace('/[^[:print:]]/', '', trim($other_link2->text())));
-  }
+    $actual = $section->filter('.ecl-footer-standardised__title');
+    $this->assertEquals('Institution links title', $actual->text());
 
-  /**
-   * Test data for the custom footer.
-   */
-  protected function getTestCustomFooterConfigsData(): array {
```
ademarco, line 267: We are removing this as it's out of scope, maybe let's drop a comment on the issue where we will style specific footer links with a link to this line, so we make sure we look at this to implement it.

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
```diff
@@ -5,316 +5,182 @@
...
-    $this->assertEquals('https://europa.eu/other_links1', $actual->attr('href'));
-    $this->assertEquals('Other link 1', trim($actual->text()));
+    $section = $crawler->filter('footer.ecl-footer-standardised section.ecl-footer-standardised__section1');
 
-    $actual = $common->filter('a:nth-child(2)');
-    $this->assertEquals('https://europa.eu/other_links2', $actual->attr('href'));
-    $this->assertEquals('Other link 2', trim($actual->text()));
+    $actual = $section->filter('a.ecl-footer-standardised__title');
+    $this->assertEquals('OpenEuropa', $actual->text());
+    $expected = \Drupal::urlGenerator()->generateFromRoute('<front>', [], ['absolute' => TRUE]);
```
ademarco: Better to always assert against a scalar value, in this case the full URL.

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
```diff
@@ -5,316 +5,182 @@
...
+  }
 
-    $actual = $second_column->filter('ul.ecl-footer__section-list li.ecl-footer__section-item:nth-child(2) > a');
-    $this->assertEquals('https://europa.eu/institution_links2', $actual->attr('href'));
-    $this->assertEquals('Institution link 2', trim($actual->text()));
+  /**
+   * Test European Union footer standardised block rendering.
+   */
+  public function testEuFooterStandardisedBlockRendering(): void {
+    $this->configFactory->getEditable('oe_theme.settings')->set('branding', 'standardised')->save();
```
ademarco: If we don't provide a template suggestion per branding then we need to make sure cache is invalidated correctly when this configuration changes.

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
```diff
@@ -5,316 +5,182 @@
...
+    $test_data = [];
+    $html = $this->renderCorporateBlocksFooter('eu', $test_data);
```
ademarco: This method render the block and fetches the test data, isn't it more clear for one method to do one thing only? Let's simply:

```php
    $test_data = $this->getFixtureContent("{$type}_footer.yml");
    $html = $this->renderCorporateBlocksFooter($type, $test_data);
```

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
```diff
@@ -5,316 +5,182 @@
...
-      'provider' => 'oe_corporate_blocks',
-      'label_display' => '0',
-    ];
-
-    $render = $this->buildBlock('oe_corporate_blocks_ec_footer', $config);
-
-    $html = (string) $this->container->get('renderer')->renderRoot($render);
+  public function testEcFooterCoreBlockRendering(): void {
+    $test_data = [];
+    $html = $this->renderCorporateBlocksFooter('ec', $test_data);
```
ademarco: This method render the block and fetches the test data, isn't it more clear for one method to do one thing only? Let's simply:

```php
    $test_data = $this->getFixtureContent("{$type}_footer.yml");
    $html = $this->renderCorporateBlocksFooter($type, $test_data);
```

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
```diff
@@ -5,316 +5,182 @@
...
-    $actual = $second_column->filter('ul li:nth-child(2) > a');
-    $this->assertEquals($test_data['social_media_links'][1]['link']['href'], $actual->attr('href'));
-    $this->assertEquals($test_data['social_media_links'][1]['link']['label'] . ' ', $actual->text());
+  /**
+   * Test European Commission footer standardised block rendering.
+   */
+  public function testEcFooterStandardisedBlockRendering(): void {
+    $this->configFactory->getEditable('oe_theme.settings')->set('branding', 'standardised')->save();
+    $test_data = [];
+    $html = $this->renderCorporateBlocksFooter('ec', $test_data);
```
ademarco: This method render the block and fetches the test data, isn't it more clear for one method to do one thing only? Let's simply:

```php
    $test_data = $this->getFixtureContent("{$type}_footer.yml");
    $html = $this->renderCorporateBlocksFooter($type, $test_data);
```

tests/Kernel/CorporateBlocks/CorporateBlocksTestBase.php
```diff
@@ -5,19 +5,161 @@
...
+    ];
+    $this->container->get('rdf_skos.skos_graph_configurator')->addGraphs($graphs);
+
+    $this->configFactory = $this->container->get('config.factory');
+    $corporate_site_info = $this->configFactory->getEditable('oe_corporate_site_info.settings');
+    $corporate_site_info->setData([
+      'site_owner' => 'http://publications.europa.eu/resource/authority/corporate-body/ACP-EU_JA',
+      'content_owners' => ['http://publications.europa.eu/resource/authority/corporate-body/COMMU'],
+    ]);
+    $corporate_site_info->save();
```
ademarco: We shouldn't do this here, we should test with empty values as well, as these values will be initially empty. We should also make sure that cache is invalidated correctly.

ademarco (changes requested):

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
drishu: I've provided template suggestion

oe_theme.theme
drishu: I'm using site_identity now, see
https://github.com/openeuropa/oe_corporate_blocks/commit/1c0de346aae6217d8b14b5d12eaaf1f559401350
the cache dependency is already done here
https://github.com/openeuropa/oe_corporate_blocks/blob/3.x/src/Plugin/Block/FooterBlockBase.php#L80

tests/Kernel/CorporateBlocks/CorporateBlocksFooterTest.php
drishu: is the ticket mentioned EWPP-468 ?

drishu:
@ademarco , @sergepavle about https://github.com/openeuropa/oe_theme/pull/642#pullrequestreview-533350159, I investigated, the service is declared properly and there is a second post_update which flushed the caches for it to be discovered,
so the fact that its not avail is strange indeed, the solution I found was to invalidate the kernel, see:
https://github.com/openeuropa/oe_corporate_blocks/pull/63/commits/6e36c13874df4d51ba7c798db4fe4ab9c954a19a

tests/Kernel/CorporateBlocks/CorporateBlocksTestBase.php
drishu: I switched to assertSession()->elementExists in the functional test which already has these messages implemented.

templates/navigation/oe-corporate-blocks-eu-footer--standardised.html.twig
```diff
@@ -0,0 +1,87 @@
...
+    'logo': {
+      'path': 'https://europa.eu/',
+      'language': current_language_id,
+      'src_mobile': _logo_path,
+      'src_desktop': _logo_path,
+    },
+    'description': "Discover more on <a href=':path' class='ecl-link ecl-link--standalone'>@label</a>"|t({
+      ':path': 'https://europa.eu/',
+      '@label': 'europa.eu'
+    }),
```
ademarco: If possible, let's make sure we keep ECL classes outside the translatable string. Since that link will not change according to language (at least we assume so), we can pass the entire link to the t() function.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  protected $configFactory;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = [
+    'block',
+    'page_header_metadata_test',
```
ademarco: Do we need this one here?

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  /**
+   * Test European Commission footer core block rendering.
+   */
+  public function testEcFooterCoreBlockRendering(): void {
+    $data = $this->getFixtureContent('ec_footer.yml');
+    $this->renderCorporateBlocksFooter('ec', $data);
+
+    $this->getSession()->reload();
```
ademarco: Instead of reloading, let's visit the site homepage.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+    $items = $data['institution_links'];
+
+    foreach ($items as $key => $expected) {
+      $index = $key + 1;
+      $actual = $section->find('css', "ul li:nth-child({$index}) > a");
+      $this->assertListLink($actual, 'standardised', $expected);
+    }
+
+    // Update settings, assert footer changed.
+    $this->updateSiteSettings();
```
ademarco: This is similar to the issue we had outlined in a previous review: the caller needs to pass to the method the values to set, even if it's used once, or else the caller needs to know what that method does.

Change this so that we can set site owner and name explicitly and assert them here below.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+    $this->getSession()->reload();
+
+    $section = $assert->elementExists('css', 'footer.ecl-footer-standardised section.ecl-footer-standardised__section1');
+
+    $actual = $section->find('css', 'a.ecl-footer-standardised__title');
+    $this->assertEquals('OpenEuropa', $actual->getText());
```
ademarco: This is the default value, change it to another one by using the method above.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+   */
+  protected function assertEclLogoPresence(NodeElement $section, string $branding): void {
+    $this->assertSession()->elementsCount('css', "a img.ecl-footer-{$branding}__logo-image-mobile", 1, $section);
+    $this->assertSession()->elementsCount('css', "a img.ecl-footer-{$branding}__logo-image-desktop", 1, $section);
+  }
+
+  /**
+   * Update the config needed from the site settings form.
+   */
+  protected function updateSiteSettings(): void {
```
ademarco: As suggested above.
```suggestion
  protected function updateSiteSettings(string $site_owner, string $site_name): void {
```

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+
+  /**
+   * Override corporate block footer config with test data.
+   *
+   * @param string $type
+   *   The type of block, ec or eu.
+   * @param array $test_data
+   *   The test data for config and assertion.
+   */
+  protected function renderCorporateBlocksFooter(string $type, array $test_data): void {
```
ademarco:
```suggestion
  protected function overrideCorporateBlocksFooter(string $type, array $test_data): void {
```

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+   * Test European Union footer standardised block rendering.
+   */
+  public function testEuFooterStandardisedBlockRendering(): void {
+    $this->configFactory->getEditable('oe_theme.settings')->set('branding', 'standardised')->save();
+    $this->configFactory->getEditable('oe_theme.settings')->set('component_library', 'eu')->save();
+
+    $data = $this->getFixtureContent('eu_footer.yml');
+    $this->renderCorporateBlocksFooter('eu', $data);
+
+    $this->getSession()->reload();
```
ademarco: To start, let's visit the site homepage instead.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+    $items = $data['institution_links'];
+
+    foreach ($items as $key => $expected) {
+      $index = $key + 1;
+      $actual = $section->find('css', "ul li:nth-child({$index}) > a");
+      $this->assertListLink($actual, 'core', $expected);
+    }
+
+    // Update settings, assert footer changed.
+    $this->updateSiteSettings();
```
ademarco: This is similar to the issue we had outlined in a previous review: the caller needs to pass to the method the values to set, even if it's used once, or else the caller needs to know what that method does.

Change this so that we can set site owner and assert it here below.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+  /**
+   * Test European Union footer core block rendering.
+   */
+  public function testEuFooterCoreBlockRendering(): void {
+    $this->configFactory->getEditable('oe_theme.settings')->set('component_library', 'eu')->save();
+
+    $data = $this->getFixtureContent('eu_footer.yml');
+    $this->renderCorporateBlocksFooter('eu', $data);
+
+    $this->getSession()->reload();
```
ademarco: To start, let's visit the site homepage.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+    $items = $data['legal_navigation'];
+
+    foreach ($items as $key => $expected) {
+      $index = $key + 1;
+      $actual = $section->find('css', "ul li:nth-child({$index}) > a");
+      $this->assertListLink($actual, 'standardised', $expected);
+    }
+
+    // Update settings, assert footer changed.
+    $this->updateSiteSettings();
```
ademarco: This is similar to the issue we had outlined in a previous review: the caller needs to pass to the method the values to set, even if it's used once, or else the caller needs to know what that method does.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,435 @@
...
+    // Rebuild the ui_pattern definitions to collect the ones provided by
+    // oe_theme itself.
+    $this->container->get('plugin.manager.ui_patterns')->clearCachedDefinitions();
+    $this->configFactory = $this->container->get('config.factory');
+  }
+
+  /**
+   * Test European Commission footer core block rendering.
+   */
+  public function testEcFooterCoreBlockRendering(): void {
```
ademarco: If we start a new test every time then the cache will not be populated when visiting the page for the first time, preventing us from testing proper invalidation.

Let's have one, big test, instead of 4 separate once. This will allow us to test cache invalidation when we switch from the installation default (ec/core) to ec/standardised, eu/core and eu/standardised: we will just change configuration, reload the page, and the new footer should be there.

ademarco (changes requested):

ademarco (changes requested):
Also @drishu  please rebase this branch on `EPIC-footer` epic branch?

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -79,6 +79,7 @@ public function getFunctions(): array {
     return [
       new \Twig_SimpleFunction('to_ecl_icon', [$this, 'toEclIcon'], ['needs_context' => TRUE]),
       new \Twig_SimpleFunction('get_link_icon', [$this, 'getLinkIcon'], ['needs_context' => TRUE]),
+      new \Twig_SimpleFunction('ecl_footer_link', [$this, 'eclFooterLinks'], ['needs_context' => TRUE]),
```
22Alexandra:
```suggestion
      new \Twig_SimpleFunction('ecl_footer_links', [$this, 'eclFooterLinks'], ['needs_context' => TRUE]),
```

templates/navigation/oe-corporate-blocks-ec-footer--standardised.html.twig
```diff
@@ -0,0 +1,71 @@
+{#
+/**
+ * @file
+ * Theme override for the corporate European Commission footer block.
```
22Alexandra:
```suggestion
 * Theme override for the corporate European Commission Standardised footer block.
```

templates/navigation/oe-corporate-blocks-ec-footer--core.html.twig
```diff
@@ -0,0 +1,40 @@
+{#
+/**
+ * @file
+ * Theme override for the corporate European Commission footer block.
```
22Alexandra:
```suggestion
 * Theme override for the corporate European Commission Core footer block.
```

templates/navigation/oe-corporate-blocks-eu-footer--core.html.twig
```diff
@@ -0,0 +1,48 @@
+{#
+/**
+ * @file
+ * Theme override for the corporate European Union footer block.
```
22Alexandra:
```suggestion
 * Theme override for the corporate European Union Core footer block.
```

templates/navigation/oe-corporate-blocks-eu-footer--standardised.html.twig
```diff
@@ -0,0 +1,87 @@
+{#
+/**
+ * @file
+ * Theme override for the corporate European Union footer block.
```
22Alexandra:
```suggestion
 * Theme override for the corporate European Union Standardised footer block.
```

patches/@ecl-twig/ec-component-footer-standardised+2.33.2.patch
```diff
@@ -0,0 +1,11 @@
+patch-package
```
22Alexandra, line 1: @drishu I don't think this patch is needed. I checked on my local without applying the patch and ecl-icon--xs class is applied by ecl-twig. Can you please check again without the patch?

22Alexandra: Nvm. It's needed.

templates/navigation/oe-corporate-blocks-eu-footer--standardised.html.twig
drishu: done as we agreed

tests/Functional/CorporateFooterRenderTest.php
drishu: did before, dono why, but indeed works well without.

tests/Functional/CorporateFooterRenderTest.php
drishu: done, implied a bit of refactoring

modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
drishu: done

oe_theme.theme
```diff
@@ -1652,3 +1490,28 @@ function oe_theme_theme_suggestions_block_alter(array &$suggestions, array $vari
...
+ */
+function oe_theme_theme_suggestions_oe_corporate_blocks_eu_footer_alter(array &$suggestions, array $variables) {
+  $branding = theme_get_setting('branding') ?? 'core';
+  $suggestions[] = 'oe_corporate_blocks_eu_footer__' . $branding;
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_theme_preprocess_oe_corporate_blocks_eu_footer(array &$variables) {
```
ademarco: :copyright:  :spaghetti: ... also test didn't catch this, so I guess there should also be something to fix there.

tests/Functional/CorporateFooterRenderTest.php
```diff
@@ -0,0 +1,419 @@
...
+      $actual = $section->find('css', "ul li:nth-child({$index}) > a");
+      $this->assertListLink($actual, 'core', $expected);
+    }
+
+    // Update settings, assert footer changed.
+    $this->updateSiteSettings('http://publications.europa.eu/resource/authority/corporate-body/ACP-EU_JA', 'EC Site Name');
+    $this->drupalGet('<front>');
+
+    $actual = $assert->elementExists('css', 'div.ecl-footer-core__description');
+    $expected = new FormattableMarkup('This site is managed by the @name', ['@name' => 'ACP–EU Joint Assembly']);
```
ademarco: Change all these expected values into simple string, you don't need to pass it as a formattable markup class.

ademarco (changes requested):

ademarco (dismissed):

22Alexandra (dismissed):

22Alexandra (approved):

imanoleguskiza (approved):

PR #649, opened 2020-11-23
EWPP-483: Theme organisation type contacts on publications.

imanoleguskiza:
### Description

Theme organisation type contacts on publications.

### Change log

- Added: Theme organisation type contacts on publications.

- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


tests/Functional/ContentPublicationRenderTest.php
```diff
@@ -0,0 +1,162 @@
...
+   * {@inheritdoc}
+   */
+  public static $modules = [
+    'config',
+    'system',
+    'oe_theme_helper',
+    'path',
+    'oe_theme_content_organisation',
+    'oe_theme_content_publication',
+    'oe_theme_content_entity_contact',
```
ademarco: We shouldn't enable dependencies on functional tests, the modules should do it. In this case we only need to list here `oe_theme_content_publication` and the rest should be enabled correctly by that module, via its dependencies. Same goes for the rest, `path`. Not sure why we need `block`?

tests/Functional/ContentRenderTestBase.php
```diff
@@ -202,7 +202,7 @@ protected function createContactEntity(string $name, string $bundle, int $status
       ],
       'status' => $status,
     ]);
-
+    $contact->save();
```
ademarco, line 206: Doesn't this work also without saving the entity in here? I wouldn't maybe change it in here, without checking everywhere else if that's ok and if similar methods should be changed too along aside it. If we need to save the entity this method creates, we can do it in the caller method, as the entity is returned.

modules/oe_theme_content_publication/oe_theme_content_publication.module
```diff
@@ -7,9 +7,53 @@
...
+  foreach ($variables['items'] as &$item) {
+    /** @var \Drupal\Core\Entity\ContentEntityInterface $contact */
+    $contact = $item['content']['#oe_contact'];
+    $cacheable_metadata  = CacheableMetadata::createFromRenderArray($item['content']);
+    // If the contact is not an organisation we don't care.
+    if ($contact->bundle() !== 'oe_organisation_reference') {
+      continue;
+    }
+    $organisation = $contact->oe_node_reference->entity;
+    if (!$organisation) {
```
ademarco: Let's use `!$organisation instanceof ...` here as we use it as an object down below. This will also help with autosuggesting in phpstorm.

modules/oe_theme_content_publication/oe_theme_content_publication.module
```diff
@@ -7,9 +7,53 @@
...
+      // If by some chance an organisation is not linked we do not render
+      // the contact.
+      $item['content']['#access'] = FALSE;
+      continue;
+    }
+    $cacheable_metadata->addCacheableDependency($organisation);
+
+    $organisation_contact = $organisation->oe_organisation_contact->entity;
+    if (!$organisation_contact) {
```
ademarco: Same here, let's use `instanceof`.

modules/oe_theme_content_publication/oe_theme_content_publication.module
```diff
@@ -7,9 +7,53 @@
...
+      // Do not render the contact if the organisation does not have contact
+      // details of its own.
+      $item['content']['#access'] = FALSE;
+      $cacheable_metadata->applyTo($item);
+      continue;
+    }
+    $cacheable_metadata->addCacheableDependency($organisation_contact);
+    $cacheable_metadata->applyTo($item);
+
```
ademarco: Empty line.

ademarco (changes requested):

modules/oe_theme_content_entity_contact/oe_theme_content_entity_contact.module
```diff
@@ -23,7 +23,7 @@ function oe_theme_content_entity_contact_oe_contact_view_alter(array &$build, Co
     ];
 
     foreach ($label_overrides as $name => $label_override) {
-      if (!$entity->get($name)->isEmpty() && isset($build[$name]['#title'])) {
+      if ($entity->hasField($name) &&!$entity->get($name)->isEmpty() && isset($build[$name]['#title'])) {
```
yenyasinn:
```suggestion
      if ($entity->hasField($name) && !$entity->get($name)->isEmpty() && isset($build[$name]['#title'])) {
```

modules/oe_theme_content_organisation/oe_theme_content_organisation.post_update.php
```diff
@@ -26,11 +25,59 @@ function oe_theme_content_organisation_post_update_00001() {
...
+  }
+
+  $entity = $entity_storage->createFromStorageRecord($config);
+  $entity->save();
+}
+
+/**
+ * Create the oe_contact 'details' view mode for organisation contacts.
+ */
+function oe_theme_content_organisation_post_update_00003() {
```
yenyasinn: oe_theme_content_organisation_post_update_00002() and oe_theme_content_organisation_post_update_00003() are identical. Might be move creation of view display to separate function in oe_theme?

modules/oe_theme_content_publication/oe_theme_content_publication.module
```diff
@@ -54,3 +56,44 @@ function oe_theme_content_publication_node_view_alter(array &$build, EntityInter
...
+    if (!$organisation instanceof ContentEntityInterface) {
+      // If by some chance an organisation is not linked we do not render
+      // the contact.
+      $item['content']['#access'] = FALSE;
+      continue;
+    }
+    $cacheable_metadata->addCacheableDependency($organisation);
+
+    $organisation_contact = $organisation->oe_organisation_contact->entity;
+    if (!$organisation_contact instanceof ContentEntityInterface) {
```
yenyasinn:
```suggestion
    if (!$organisation_contact instanceof Drupal\oe_content_entity_contact\Entity\ContactInterface) {
```

tests/Functional/ContentPublicationRenderTest.php
```diff
@@ -276,6 +276,77 @@ public function testPublicationRendering(): void {
     $this->assertCount(4, $content_items);
     $this->assertContentHeader($content_items[3], 'Contact', 'contact');
     $this->assertContactEntityDefaultDisplay($content_items[3], 'publication_contact');
+
+    // Assert Organisation Contact.
+    $organisation_contact = $this->createContactEntity('organisation_contact', 'oe_general', 1);
```
yenyasinn:
```suggestion
    $organisation_contact = $this->createContactEntity('organisation_contact', 'oe_general', CorporateEntityInterface::PUBLISHED);
```

tests/Functional/ContentPublicationRenderTest.php
```diff
@@ -276,6 +276,77 @@ public function testPublicationRendering(): void {
...
+    $organisation_contact->save();
```
yenyasinn: Instead of this would be better to add `$contact->save()` in `ContentRenderTestBase::createContactEntity()`

tests/Functional/ContentPublicationRenderTest.php
```diff
@@ -276,6 +276,77 @@ public function testPublicationRendering(): void {
...
+    // Assert that the contact is not being rendered since no organisation is
+    // being referenced.
+    $content = $this->assertSession()->elementExists('css', '.ecl-row.ecl-u-mt-l .ecl-col-lg-9');
+    $content_items = $content->findAll('xpath', '/div');
+    $this->assertCount(3, $content_items);
+
+    // Add an organisation to the publication contact and assert it still
+    // nothing gets rendered since the organisation does not have a
+    // contact itself.
+    $publication_contact->set('oe_node_reference', [
```
yenyasinn:
```suggestion
    $publication_contact->set('oe_node_reference', $organisation);
```

tests/Functional/ContentPublicationRenderTest.php
```diff
@@ -276,6 +276,77 @@ public function testPublicationRendering(): void {
...
+    ]);
+    $publication_contact->save();
+    $this->drupalGet($node->toUrl());
+    $content = $this->assertSession()->elementExists('css', '.ecl-row.ecl-u-mt-l .ecl-col-lg-9');
+    $content_items = $content->findAll('xpath', '/div');
+    $this->assertCount(3, $content_items);
+
+    // Add a contact to the organisation and assert it gets rendered
+    // in the publication page.
+    $organisation->set('oe_organisation_contact', [
```
yenyasinn:
```suggestion
    $organisation->set('oe_organisation_contact', $organisation_contact);
```

tests/Functional/ContentRenderTestBase.php
```diff
@@ -203,7 +203,6 @@ protected function createContactEntity(string $name, string $bundle, int $status
       ],
       'status' => $status,
     ]);
-
```
yenyasinn, line 206: `$contact->save()` as in another create* methods

tests/Functional/ContentRenderTestBase.php
yenyasinn: Our create* methods work inconsistent - createMedia*() save media and return saved entity

tests/Functional/ContentPublicationRenderTest.php
yenyasinn: Our create* methods work inconsistent - createMedia*() save media and return saved entity

modules/oe_theme_content_publication/oe_theme_content_publication.module
imanoleguskiza: We never use the full namespace in this context though.

imanoleguskiza: Ah, nevermind, didn't realize its a completely different class :sob:

tests/Functional/ContentPublicationRenderTest.php
imanoleguskiza: I think you are right, I will add it to the createContactEntity to be consistent.

modules/oe_theme_content_organisation/oe_theme_content_organisation.post_update.php
imanoleguskiza: I think you have a point but we would need to do everywhere so I think I rather keep it as is for now, get the epic finished and find a better solution after.

yenyasinn (dismissed):

modules/oe_theme_content_entity_contact/oe_theme_content_entity_contact.module
```diff
@@ -23,7 +23,7 @@ function oe_theme_content_entity_contact_oe_contact_view_alter(array &$build, Co
     ];
 
     foreach ($label_overrides as $name => $label_override) {
-      if (!$entity->get($name)->isEmpty() && isset($build[$name]['#title'])) {
+      if ($entity->hasField($name) && !$entity->get($name)->isEmpty() && isset($build[$name]['#title'])) {
```
ademarco, line 26: We don't do this on the other implementation of `hook_ENTITY_TYPE_view_alter()`, should we? Why we needed this here in the first place?

tests/Functional/ContentRenderTestBase.php
imanoleguskiza: I decided to keep similar methods consistent and add the contact->save in the end. The only other place where this is used is on the news test and so it doesn't really make any difference in that regard

ademarco (changes requested):
It's two times that I'm trying and I'm getting always the following when testing an update scenario:

```
git checkout 2.11.0
rm -rf build/*/contrib/ build/core/ composer.lock vendor/
composer update
run drupal:site-install --skip-permissions-setup
```

Visit the site, warm the cache, etc.

```
git checkout EWPP-483
rm -rf build/*/contrib/ build/core/ composer.lock vendor/
composer update
drush updb -y
```

Update works well, no errors, but then when clearing the cache:

```
>  [notice] Update completed: oe_theme_content_publication_post_update_00006
>  [notice] Update started: oe_theme_content_publication_post_update_00007
>  [notice] Update completed: oe_theme_content_publication_post_update_00007
>  [notice] Update started: oe_theme_content_publication_post_update_00008
>  [notice] Update completed: oe_theme_content_publication_post_update_00008
 [success] Finished performing updates.
~/S/oe_theme (EWPP-483|✔) $ drush cr

In FieldStorageConfigStorage.php line 174:
                                                                                                                              
  Unable to determine class for field type 'file_link' found in the 'field.storage.media.oe_media_remote_file' configuration  
                                                                                                                              

In DiscoveryTrait.php line 53:
                                                                                                                                                                                                             
  The "file_link" plugin does not exist. Valid plugin IDs for Drupal\Core\Field\FieldTypePluginManager are: address, address_zone, address_country, contact_storage_options_email, datetime, daterange, ent  
  ity_reference_revisions, file_uri, file, image, link, oe_featured_media, timeline_field, list_integer, list_float, list_string, path, skos_concept_entity_reference, telephone, text_with_summary, text,   
  text_long, typed_link, boolean, uri, decimal, created, timestamp, email, string_long, language, string, float, map, password, uuid, integer, entity_reference, changed                                     
```                                                                                                                    

Clearing the cache again will make the error go away.

**EDIT:** This was fixed in https://github.com/openeuropa/oe_media/pull/136

modules/oe_theme_content_entity_contact/oe_theme_content_entity_contact.module
imanoleguskiza: Because here we are assuming the fields are there to change, but some contact bundles might not have those fields (like the organisation bundle we introduce are themign here).  For the other entity alters we check for the bundle, which is an alternative option if you rather do it that way.

ademarco: Right sorry, I didn't pay enough attention to it.

modules/oe_theme_content_organisation/oe_theme_content_organisation.post_update.php
```diff
@@ -26,11 +25,59 @@ function oe_theme_content_organisation_post_update_00001() {
...
+  if ($existing) {
+    return t('Contact entity view display already exists, skipping.');
+  }
+
+  $entity = $entity_storage->createFromStorageRecord($config);
+  $entity->save();
+}
+
+/**
+ * Create the oe_contact 'details' view mode for organisation contacts.
```
ademarco:
```suggestion
 * Create the oe_contact 'full' view mode for organisation contacts.
```

modules/oe_theme_content_publication/oe_theme_content_publication.info.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
   - oe_content:oe_content_publication
   - oe_theme:oe_theme_content_entity_contact
   - oe_theme:oe_theme_helper
+  - oe_theme:oe_theme_content_organisation
```
ademarco: We shouldn't do it, I believe that we can not have content type depending on each other, given how we designed the whole thing

composer.json
```diff
@@ -44,7 +44,7 @@
         "openeuropa/behat-transformation-context": "~0.1",
         "openeuropa/code-review": "~1.5",
         "openeuropa/drupal-core-require-dev": "^8.7",
-        "openeuropa/oe_content": "dev-EPIC-Publication",
+        "openeuropa/oe_content": "dev-EWPP-483-no-deps",
```
yenyasinn:
```suggestion
        "openeuropa/oe_content": "dev-EWPP-483",
```

modules/oe_theme_content_organisation/oe_theme_content_organisation.module
```diff
@@ -14,6 +14,8 @@ use Drupal\media\MediaInterface;
 use Drupal\media_avportal\Plugin\media\Source\MediaAvPortalPhotoSource;
 use Drupal\media\Plugin\media\Source\Image;
 use Drupal\oe_theme\ValueObject\ImageValueObject;
+use Drupal\Core\Entity\ContentEntityInterface;
```
yenyasinn: These classes aren't used anymore

tests/Functional/ContentPublicationRenderTest.php
```diff
@@ -26,6 +27,9 @@ class ContentPublicationRenderTest extends ContentRenderTestBase {
     'path',
     'oe_theme_helper',
     'oe_theme_content_publication',
+    'oe_theme_content_organisation',
+    'oe_content_organisation_reference',
```
yenyasinn: `oe_content_organisation_reference` can be removed because dependency exists in oe_theme_content_organisation_reference

ademarco (approved):

modules/oe_theme_content_publication/oe_theme_content_publication.info.yml
ademarco: Damn, this was stored in my browser local cache somehow, I'll resolve it.

modules/oe_theme_content_organisation/oe_theme_content_organisation.post_update.php
ademarco: Damn, this was stored in my browser local cache somehow, I'll resolve it.

yenyasinn (approved):

PR #650, opened 2020-11-23
EWPP-544: Render file patterns only when file is set.

22Alexandra:


src/DocumentMediaValueExtractor.php
```diff
@@ -77,6 +77,10 @@ protected static function getFromRemoteFile(MediaInterface $media): ?FileValueOb
     }
 
     $file_link = $media->get('oe_media_remote_file')->first();
+    if ($media->get('oe_media_remote_file')->first()->getSize() == 0) {
```
sergepavle: ===

sergepavle: We can reuse $file_link variable.

templates/content/media--document--default.html.twig
```diff
@@ -6,8 +6,10 @@
  * @see ./core/themes/stable/templates/content/media.html.twig
  */
 #}
-{{ pattern('file_translation', {
-  'button_label': 'Download'|t,
-  'file': file|default([]),
-  'translations': translations|default([]),
-}) }}
+{% if file is not empty %}
```
sergepavle: We can drop `is not empty`. Everything else will be handled by twig.

sergepavle (changes requested):
I've left a few comments.

sergepavle (dismissed):

sergepavle (dismissed):

upchuk (approved):

sergepavle (approved):

sergepavle (approved):

PR #653, opened 2020-11-24
EWPP-468: Add site-specific other links to corporate footers.

drishu:
### Description

Add site-specific other links to corporate footers.

### Change log

- Changed:
  - sections 2 and 3 to corporate footer templates
  - adapted the test with new footer link sections and other links

### DIFF
--- a/templates/navigation/oe-corporate-blocks-ec-footer--standardised.html.twig
+++ b/templates/navigation/oe-corporate-blocks-ec-footer--standardised.html.twig
@@ -76,42 +76,44 @@
   }]) %}
 {% endif %}
 
-{% set _sections = [
-  {
-    'title': {
+{% block content %}
+  {% set _sections = [
+    {
+      'title': {
       'link': {
         'label': site_specific_footer.site_identity,
         'path': url('<front>'),
       }
     },
-    'description': site_owner ? 'This site is managed by the @name'|t({'@name': site_owner}),
-    'section_id': 1,
-  },
-  _section_2,
-  _section_3,
-  {
-    'section_id': 6,
-    'list_class_name': 'ecl-footer-standardised__list--condensed',
-  },
-  {
-    'title': {
+      'description': site_owner ? 'This site is managed by the @name'|t({'@name': site_owner}),
+      'section_id': 1,
+    },
+    _section_2,
+    _section_3,
+    {
+      'section_id': 6,
+      'list_class_name': 'ecl-footer-standardised__list--condensed',
+    },
+    {
+      'title': {
       'link': {
         'label': corporate_footer.corporate_site_link.label,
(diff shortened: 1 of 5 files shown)


georgebaev (commented):
@drishu I found the following misalignment for a specific configuration of footer links:
image
image
Please see how the `Related sites` section is not top-aligned with the `Contact us` section. 

I am not sure if this misalignment is relevant to this PR's functionality

ademarco (changes requested):
So when there are 3 sections and the only filled in one is not the first, and we have no social links:

image

Then this happens: 

image

I think we need to remove empty sections, so initially related links will be on top of social links.

Or, we keep the way it is now, so related links will always be on the right, unless there are no social links, in which case we don't bother.

I think we need to involve @kugta

ademarco:
@drishu actually checking the empty behaviour here: https://ec.europa.eu/component-library/playground/ec/?path=/story/components-footers-standardised--default 

Let's make sure that when we don't have the left column, the right column is displayed on the left:

Peek 2020-11-25 14-09

Also mind the top margin, we get this now because we print empty sections.

image

templates/navigation/oe-corporate-blocks-ec-footer--standardised.html.twig
```diff
@@ -11,53 +11,70 @@
...
+    {% set _section_3 = _section_3|merge([_last_in_section_2]) %}
 
-{% for section in site_specific_footer.other_links|slice(1) %}
+  {% endif %}
 
-  {% set _section_3 = _section_3|merge([{
-    'title': section.links ? section.label,
-    'links': section.links ? ecl_footer_links(section.links),
-    'section_id': 3,
+  {# Follow section must always be the last on on the left. #}
```
georgebaev: There is a double "on" in the comment

templates/navigation/oe-corporate-blocks-eu-footer--standardised.html.twig
```diff
@@ -11,52 +11,70 @@
...
+      {% set _section_2 = [] %}
+    {% endif %}
+
+    {% set _section_2 = _section_2[:_section_2|length - 1] %}
+    {% set _last_in_section_2 = _last_in_section_2|merge({'section_id': 3}) %}
+    {% set _section_3 = _section_3|merge([_last_in_section_2]) %}
+
+  {% endif %}
+
+  {# Follow section must always be the last on on the left. #}
```
georgebaev: There is a double "on" in the comment.

georgebaev (dismissed):

22Alexandra (dismissed):

ademarco (dismissed):

ademarco (approved):

georgebaev (approved):

PR #657, opened 2020-11-26
EWPP-485: Adjust selectors for behat tests.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/tests/Behat/OeThemeTestContext.php
+++ b/tests/Behat/OeThemeTestContext.php
@@ -352,10 +352,10 @@ class OeThemeTestContext extends RawDrupalContext {
     ];
 
     // Make sure a corporate footer is present on the page.
-    $this->assertSession()->elementExists('css', 'footer.ecl-footer-core');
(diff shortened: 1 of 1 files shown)


drishu (dismissed):

drishu (changes requested):
I think this branch, for now, should use 
`"openeuropa/oe_corporate_blocks": "dev-EWPP-485 as 3.x-dev"`
so we can see it together.
without this, the upgrade test fails with whitescreen.

sergepavle:
> I think this branch, for now, should use
> `"openeuropa/oe_corporate_blocks": "dev-EWPP-485 as 3.x-dev"`
> so we can see it together.
> without this, the upgrade test fails with whitescreen.

I wouldn't overcomplicate chains of PR. Main PR already exists for ewcms. So any changes in oe_corporate_blocks will be visible in PR for ewcms. This PR just for adjusting the behat step definition.

drishu (approved):

ademarco (approved):

PR #676, opened 2020-12-08
EWPP-583: Render iframe media type inside media container

yenyasinn:

### DIFF
--- a/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
+++ b/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
@@ -5,6 +5,7 @@ declare(strict_types = 1);
 namespace Drupal\oe_theme_helper\Plugin\Field\FieldFormatter;
 
 use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Cache\CacheableDependencyInterface;
 use Drupal\Core\Entity\Entity\EntityViewDisplay;
 use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\Core\Field\FieldItemInterface;
--- a/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
+++ b/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
@@ -175,41 +175,7 @@ class FeaturedMediaFormatter extends EntityReferenceFormatterBase {
     $source = $media->getSource();
 
     if ($source instanceof MediaAvPortalVideoSource || $source instanceof OEmbed || $source instanceof Iframe) {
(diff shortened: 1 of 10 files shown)


modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
```diff
@@ -175,41 +176,7 @@ protected function viewElement(FieldItemInterface $item, string $langcode): arra
...
-      if ($media->bundle() === 'video_iframe') {
-        $ratio = $media->get('oe_media_iframe_ratio')->value;
-        $params['ratio'] = str_replace('_', '-', $ratio);
-      }
-
-      $build['#params'] = $params;
-      $cacheability->applyTo($build);
-
-      return $build;
+      return $this->viewEmbededMedia($media, $build, $params, $cacheability);
```
yenyasinn, line 179: It was rewritten due to cyclomatic complexity error in PhpMd

georgebaev (approved):

tests/Kernel/MediaRenderTest.php
```diff
@@ -250,4 +254,31 @@ public function testDocumentMedia(): void {
...
+      'bundle' => 'iframe',
+      'name' => 'test iframe',
+      'oe_media_iframe' => '<iframe src="http://example.com/iframe_media"></iframe>',
+    ]);
+    $media->save();
+
+    $build = $this->mediaViewBuilder->view($media, 'oe_theme_main_content');
+    $html = $this->renderRoot($build);
+    $crawler = new Crawler($html);
+    $iframe = $crawler->filter('.ecl-media-container .ecl-media-container__media--ratio-16-9 iframe');
```
brummbar, line 272: As per specifications, if no ratio is specified in the iframe media, no ratio should be used in the media container.

brummbar (changes requested):
Early quick review, I didn't go through all the code.

tests/Kernel/MediaRenderTest.php
yenyasinn: ECL component components/ec-component-media-container/ecl-media-container.html.twig sets default ratio 16:9 if ratio isn't provided.
There isn't class that defines media container without ratio.
If we pass in ratio variable value `no-ratio` then class is `ecl-media-container__media--ratio-no-ratio` and media container looks broken:
Screenshot 2020-12-10 at 11 12 43

Therefore if we want not to set any ratio ECL's media container should be adjusted.

brummbar: Well if the media container is meant only to handle responsiveness, then it's as easy as not using it when a custom iframe size is needed. Let's change the code this way.

yenyasinn: @brummbar sorry, but what does "Let's change the code this way." mean?

yenyasinn:
Closed in favour https://github.com/openeuropa/oe_theme/pull/690

PR #688, opened 2020-12-15
EWPP-635: Port security fixes to 2.11.x

22Alexandra:


composer.json
```diff
@@ -43,10 +43,10 @@
         "openeuropa/behat-transformation-context": "~0.1",
         "openeuropa/code-review": "~1.5",
         "openeuropa/drupal-core-require-dev": "^8.7",
-        "openeuropa/oe_content": "dev-master",
+        "openeuropa/oe_content": "1.11.1",
```
sergepavle, line 46: I don't sure, but perhaps it is not necessary to lock version but just declare maximum version like `<=1.11.1`

sergepavle (dismissed):

composer.json
22Alexandra: It would use 1.3.0 for oe_content and the module oe_content_entity_contact is not available in this version. Same for the other PR.

sergepavle: Ok. In this case simpler to lock...

sergepavle (approved):

ademarco (approved):

PR #689, opened 2020-12-16
EWPP-614: Make sure that multiple values of the same field are always separated by a comma when displayed as meta.

22Alexandra:


modules/oe_theme_content_publication/src/Plugin/PageHeaderMetadata/PublicationContentType.php
```diff
@@ -79,27 +79,27 @@ public function getMetadata(): array {
...
    * @param \Drupal\rdf_skos\Plugin\Field\SkosConceptReferenceFieldItemList $items
    *   Field item list object.
    *
-   * @return array
-   *   List of items.
+   * @return string
+   *   Comma separated string.
    */
-  protected function getSkosMeta(SkosConceptReferenceFieldItemList $items): array {
+  protected function getCommaSeparatedSkosMeta(SkosConceptReferenceFieldItemList $items): string {
```
sergepavle: Seems like we have duplication of code https://github.com/openeuropa/oe_theme/blob/EWPP-614/modules/oe_theme_content_news/src/Plugin/PageHeaderMetadata/NewsContentType.php#L139. This method can be relocated to the parent class.

modules/oe_theme_content_news/oe_theme_content_news.module
```diff
@@ -33,6 +34,33 @@ function oe_theme_content_news_preprocess_node__oe_news(&$variables) {
...
+
+  // Add news locations to page metadata.
+  if (!$node->get('oe_news_location')->isEmpty()) {
+    $variables['metas']['location'] = _get_comma_separated_skos_meta($node->get('oe_news_location'));
+  }
+
+  // Add news authors to page metadata.
+  if (!$node->get('oe_author')->isEmpty()) {
+    $variables['metas']['authors'] = _get_comma_separated_skos_meta($node->get('oe_author'));
+  }
```
sergepavle: As I understand it would be good to reorganize code for reducing code duplications everywhere https://github.com/openeuropa/oe_theme/blob/EWPP-614/modules/oe_theme_content_news/src/Plugin/PageHeaderMetadata/NewsContentType.php#L108.
Probably static method for building meta for new content type can be a good solution but it's only one of the solutions.
Anyway, it would be nice to have a single method/function for the formation of metas elements for rendering.

sergepavle (dismissed):

modules/oe_theme_content_news/oe_theme_content_news.module
```diff
@@ -33,6 +34,33 @@ function oe_theme_content_news_preprocess_node__oe_news(&$variables) {
...
 }
 
+/**
+ * Implements template_preprocess_node() for the News node type.
+ */
+function oe_theme_content_news_preprocess_node__oe_news__teaser(&$variables) {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $variables['node'];
+
+  // Add news types to page metadata.
```
drishu: to teaser metadata. same below

modules/oe_theme_content_news/oe_theme_content_news.module
```diff
@@ -33,6 +34,33 @@ function oe_theme_content_news_preprocess_node__oe_news(&$variables) {
   }
 }
 
+/**
+ * Implements template_preprocess_node() for the News node type.
+ */
+function oe_theme_content_news_preprocess_node__oe_news__teaser(&$variables) {
```
drishu: if we do it like this for news teaser, should we not do it also for publication teaser ?

modules/oe_theme_content_news/oe_theme_content_news.module
```diff
@@ -33,6 +34,33 @@ function oe_theme_content_news_preprocess_node__oe_news(&$variables) {
...
+    $variables['metas']['types'] = ContentUtility::getCommaSeparatedReferencedEntityLabels($node->get('oe_news_types'));
+  }
+
+  // Add publication date to page metadata.
+  $timestamp = $node->get('oe_publication_date')->date->getTimestamp();
+  $variables['metas']['date'] = \Drupal::service('date.formatter')->format($timestamp, 'oe_theme_news_date');
+
+  // Add news locations to page metadata.
+  if (!$node->get('oe_news_location')->isEmpty()) {
+    $variables['metas']['location'] = ContentUtility::getCommaSeparatedReferencedEntityLabels($node->get('oe_news_location'));
```
drishu: location can have only one value https://github.com/openeuropa/oe_content/blob/master/modules/oe_content_news/config/install/field.storage.node.oe_news_location.yml#L15, dont think getCommaSeparatedReferencedEntityLabels is needed

templates/content/node--oe-news--teaser.html.twig
```diff
@@ -13,8 +13,10 @@
     'url': url,
     'title': content.extra_field_oe_theme_helper_short_title_with_fallback|field_value,
     'meta': [
-      content.oe_news_types|field_value|safe_join(', '),
-      content.oe_publication_date|field_value,
+      metas.types,
+      metas.date,
+      metas.location,
```
drishu: News teaser should have only News type + Publication date as meta:
https://webgate.ec.europa.eu/fpfis/wikis/display/EWPP/News+Content+Type+Specification+-+Version+3.0.0
&
https://webgate.ec.europa.eu/fpfis/wikis/download/attachments/376047978/News%20CT%20-%20Desktop%20Pool%20page%20.png?version=2&modificationDate=1601992582732&api=v2

modules/oe_theme_helper/src/ContentUtility.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_helper;
+
+use Drupal\Core\Field\EntityReferenceFieldItemListInterface;
+
+/**
+ * Utility class for content data.
```
drishu:
```suggestion
 * Provides field value-related conversions and operations.
```
I'm shying away from "content data" because below we are clearly handling fields and their values.

modules/oe_theme_helper/src/ContentUtility.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+class ContentUtility {
```
drishu: class name could be more descriptive as to what it should be used for, but not a blocker for me..

Anyway, below code belongs in a formatter which then can be called where needed.. but that's a big change for the architecture so lets discuss with @ademarco 

This way the preprocessing disappears, the formatter is set in the view display config, and we can still have this class for shorthand of rendering the field with the formatter where needed programatically.

drishu (changes requested):
Unrelated I think, but found during testing:
`web_1       | 172.19.0.1 - - [12/Jan/2021:21:08:40 +0000] "GET /build/en/node/add/oe_project HTTP/1.1" 200 16482
web_1       | [12-Jan-2021 21:10:32] ERROR: failed to ptrace(ATTACH) child 1254: Operation not permitted (1)
web_1       | [12-Jan-2021 21:10:32] WARNING: [pool www] child 1254, script '/var/www/html/build/index.php' (request: "GET /build/index.php") executing too slow (73.602350 sec), logging
`

modules/oe_theme_content_news/oe_theme_content_news.module
nagyad: Ultimately it's the same code.

drishu (dismissed):

modules/oe_theme_helper/src/ContentUtility.php
ademarco: Yeah I don't think this is the way to go, to be honest. Basically we are capturing this into a static class so we can reuse it in the page metadata and in the pattern enhancement in EWCMS. But we do that at the cost of:

- Effectively breaking dependency injection in page metadata plugins, since then we use this class' static method in them, which is like calling `\Drupal::service('entity.repository')->getTranslationFromContext($entity)->label();` in a plugin
- Loosing the convenience of doing everything in twig templates: now we need to resort to preprocesses, while before we have a simple `content.oe_news_types|field_value|safe_join(', ')` taking care of things.

I think an improvement would have been having a formatter and then using that formatter everywhere, in the teaser twig templates here, and in the pattern enhancer in EWCMS. Or even a simple field template could have worked.

This would still leave us with the page metadata method, which is a bit of a duplicate but it's not that problematic, as all page metadata plugin extend the same base class.

ademarco: I had a look trying to simply print the field as is in the template, just to see the type of suggestions we would get. So by simply doing this:

image

We would get this, which is quite easy to show it as comma separated, by using only Twig:

image

So basically we wouldn't need a formatter either, just a twig template. Like that we don't really have to bother about not reusing logic in page metadata plugins.

ademarco (changes requested):

drishu (dismissed):

drishu (dismissed):

ademarco (approved):

drishu (approved):

PR #695, opened 2021-01-08
EWPP-607: Render languages categorised on language selector block and page.

nagyad:


22Alexandra (dismissed):

tests/Behat/DrupalContext.php
```diff
@@ -126,4 +127,30 @@ public static function disableDatetimeTesting(AfterFeatureScope $scope): void {
     \Drupal::service('module_installer')->uninstall(['datetime_testing']);
   }
 
+  /**
+   * Enables a non-EU language for the given scenario.
```
upchuk: Creates a,,

tests/Behat/DrupalContext.php
```diff
@@ -126,4 +127,30 @@ public static function disableDatetimeTesting(AfterFeatureScope $scope): void {
...
+   * @BeforeScenario @enable-non-eu-language
+   */
+  public static function enableNonEuLanguage(BeforeScenarioScope $scope): void {
+    $language = ConfigurableLanguage::createFromLangcode('is');
+    $language->setThirdPartySetting('oe_multilingual', 'category', 'non_eu');
+    $language->save();
+  }
+
+  /**
+   * Disables a non-EU language for the given scenario.
```
upchuk: Deletes a..

upchuk (changes requested):

oe_theme.theme
```diff
@@ -598,15 +596,48 @@ function oe_theme_preprocess_pager(array &$variables): void {
...
+ *   The link items passed from the preprocess function.
+ * @param string $current_language_id
+ *   The current language id.
+ *
+ * @return array
+ *   Array of variables to be used for rendering.
+ */
+function _oe_theme_preprocess_languages(array $link_items, string $current_language_id = ''): array {
+  $variables = [];
+  $oe_multilingual = \Drupal::moduleHandler()->moduleExists('oe_multilingual');
```
upchuk: Let's not care about oe multilingual here. We just need to know: are there "categories" on the language enttities, if so, we split, otherwise we don't change anything. This is also means that if we ONLY have EU languages, we don't want any change from the current situation because it adds a title at the top as well `EU Official Languages`.

Basically we need to be 100% backwards compatible if there are no non-EU languages configured as well.

Ensure also that the tests account for both cases

templates/compositions/ec-component-language-list/language-list-selector.html.twig
```diff
@@ -39,6 +46,12 @@
 
 {# Internal logic - Process properties #}
 
+{% if eu_languages is not empty and eu_languages is iterable %}
+  {% set items = eu_languages %}
+  {% set eu_category = 'EU official languages'|t %}
```
upchuk: Do we have translations for these strings we can ship? There was that huge excel file with translations somewhere.

tests/Kernel/LanguageSwitcherTest.php
```diff
@@ -42,6 +49,41 @@ public function testLanguageSwitcherLinkListRendering(): void {
         preg_replace('/[\x00-\x1F\x7F\xA0]/u', '', $actual)
       );
     }
+
+    // Set Hungarian as non-EU language to test the non-EU category.
```
upchuk: Don't use EU languages as non-EU language in test please. Install a new one that is not EU.

oe_theme.theme
```diff
@@ -1102,6 +1152,13 @@ function oe_theme_preprocess_block__system_branding_block(&$variables) {
 
   $route_options = ['absolute' => TRUE];
   $variables['site_logo_href'] = Url::fromRoute('<front>', [], $route_options)->toString();
+
+  // Ensure the site logo falls back to english version if the language is
+  // not EU language.
+  $language_code = \Drupal::languageManager()->getCurrentLanguage()->getId();
```
upchuk, line 1154: We also have the logo in the footer when the European Union theme flavour is used:

Screen Shot 2021-01-12 at 09 19 51

upchuk (changes requested):

templates/compositions/ec-component-language-list/language-list-selector.html.twig
nagyad: No translation for this combination of strings in that xls.

oe_theme.theme
```diff
@@ -632,8 +662,24 @@ function oe_theme_preprocess_links__language_block(array &$variables): void {
...
+    switch ($category) {
+      case 'eu':
+        $variables['eu_languages'][] = $language;
+        break;
+
+      case 'non_eu':
+        $variables['non_eu_languages'][] = $language;
+        break;
+
+      case 'no_category':
```
upchuk: Mmm, let's not use our own thing here and use a `default` case. Meaning: if it's not eu or non_eu, it goes in this bucket. So that if the category is set as banana, it's not a valid category. No?

templates/compositions/ec-component-language-list/language-list-selector.html.twig
```diff
@@ -71,7 +87,10 @@
   {{- _label -}}
  </a>
 {% include '@ecl-twig/language-list/language-list' with {
+  eu_category: eu_category,
```
upchuk, line 90: If the category vars don't exist yet, what happens here? Shouldn;t we initialize the variable?

tests/Kernel/LanguageSwitcherTest.php
```diff
@@ -15,6 +16,65 @@ class LanguageSwitcherTest extends MultilingualAbstractKernelTestBase {
...
+
+    // Assert there is only one link in the non-EU category.
+    $actual = $crawler->filter('.ecl-language-list--overlay .ecl-language-list__non-eu a');
+    $this->assertCount(1, $actual);
+
+    // Assert the Icelandic language link.
+    $actual = $crawler->filter(".ecl-language-list--overlay .ecl-language-list__non-eu a.ecl-language-list__link[lang={$lang_prefix}]")->text();
+    $this->assertEquals(
+      $icelandic->getName(),
+      preg_replace('/[\x00-\x1F\x7F\xA0]/u', '', $actual)
```
upchuk, line 80: Why is this necessary?

tests/Kernel/LanguageSwitcherTest.php
```diff
@@ -15,6 +16,65 @@ class LanguageSwitcherTest extends MultilingualAbstractKernelTestBase {
...
+    );
+
+    // Test backwards-compatibility by removing categories from the existing
+    // languages.
+    foreach ($this->container->get('language_manager')->getLanguages() as $language) {
```
upchuk: Or, ConfigurableLanguage::loadMultiple()

tests/Kernel/LanguageSwitcherTest.php
```diff
@@ -15,6 +16,65 @@ class LanguageSwitcherTest extends MultilingualAbstractKernelTestBase {
    * Test language switcher link list rendering.
    */
   public function testLanguageSwitcherLinkListRendering(): void {
+    $this->assertDefaultLanguageBlock();
+
+    $lang_config = $this->container->get('config.factory')->get('language.negotiation');
+
+    // Create the Icelandic language.
+    $language = ConfigurableLanguage::createFromLangcode('is');
+    $language->setThirdPartySetting('oe_multilingual', 'category', 'non_eu');
```
upchuk, line 49: Before you set the category, run an assertion that our defaults are still in place. We can have Icelanding but if not categorised, the defaults apply: only EU langs show, no category title, etc.

upchuk (changes requested):

oe_theme.theme
nagyad: Yes, I was hesitating with this and default, because if it is anything else it won't go anywhere not even this bucket as you say, for banana. For me it's okay like that as well.

templates/compositions/ec-component-language-list/language-list-selector.html.twig
nagyad: Twig takes care of that, no need initialising vars.

tests/Kernel/LanguageSwitcherTest.php
nagyad: I did not create this, this code is from before. But it was done by Antonio to remove any non-printable characters from $actual.

tests/Kernel/LanguageSwitcherTest.php
```diff
@@ -15,6 +16,91 @@ class LanguageSwitcherTest extends MultilingualAbstractKernelTestBase {
...
   public function testLanguageSwitcherLinkListRendering(): void {
+    $this->assertDefaultLanguageBlock();
+
+    $lang_config = $this->container->get('config.factory')->get('language.negotiation');
+
+    // Create the Icelandic language.
+    $language = ConfigurableLanguage::createFromLangcode('is');
+    $language->save();
+    // Load the language object.
+    $icelandic = $this->container->get('language_manager')->getLanguage('is');
```
upchuk: You know the lang ID. It's `is`, you load it based on that. Also, why you still loading it since you have it from when you created it above at line 24.

tests/Behat/DrupalContext.php
```diff
@@ -126,4 +127,30 @@ public static function disableDatetimeTesting(AfterFeatureScope $scope): void {
     \Drupal::service('module_installer')->uninstall(['datetime_testing']);
   }
 
+  /**
+   * Creates a non-EU language for the given scenario.
+   *
+   * @param \Behat\Behat\Hook\Scope\BeforeScenarioScope $scope
+   *   The scope.
+   *
+   * @BeforeScenario @enable-non-eu-language
```
upchuk, line 136: Why we don't have a parameter to pass which language to create here? We might use this later. I just noticed. Perhaps a step like this already exists?

upchuk (changes requested):
Let's make an EPIC branch for the multilingual (same name as on EWCMS) to ensure we don't merge the epic into 2.x before the entire functionality is finished.

tests/Behat/DrupalContext.php
upchuk: Ah nevermind, i thought it was a step.

upchuk (approved):

hernani (approved):

PR #698, opened 2021-01-14
EWPP-675: Remove media container from op publication list rendering.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ docker-compose.*.yml
 
 /images/*
 !/images/required.svg
+/.gitattributes
--- a/templates/field/field--media--webtools-op-publication-list.html.twig
+++ b/templates/field/field--media--webtools-op-publication-list.html.twig
@@ -4,8 +4,4 @@
  * Theme override for the webtools op publication list fields on media items.
  */
 #}
-{% for item in items %}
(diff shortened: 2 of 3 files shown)


22Alexandra (approved):

.gitignore
```diff
@@ -24,3 +24,4 @@ docker-compose.*.yml
 
 /images/*
 !/images/required.svg
+/.gitattributes
```
drishu, line 27: this was added automatically

templates/field/field--media--webtools-op-publication-list.html.twig
```diff
@@ -4,8 +4,4 @@
  * Theme override for the webtools op publication list fields on media items.
  */
 #}
-{% for item in items %}
-  {% include '@ecl-twig/media-container' with {
-    'embedded_media': item.content,
-  } %}
-{% endfor %}
+{%- include 'field--bare.html.twig' -%}
```
drishu, line 7: if we delete the template altogether then we do get the field title and some attribute-less divs so I guess this is better.

drishu (approved):

PR #705, opened 2021-01-22
EWPP-640: Update to last version of ECL (2.35.2 version) with security fixes.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra (approved):

22Alexandra (changes requested):
Still needs tests fixes.

22Alexandra (dismissed):

oe_theme.theme
```diff
@@ -1289,6 +1290,17 @@ function oe_theme_preprocess_paragraph__oe_social_media_follow(array &$variables
   }
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_pattern_text_featured_media(array &$variables): void {
```
drishu: As far as I can see this is here just to make the tests work, if this is the case I suggest rather using one of below workarounds:
1) change the render.yml so that the form api does this work for us, ex:
```
video:
  '#type': 'inline_template'
  '#template': "<iframe id=\"videoplayerI-181645\" src=\"//ec.europa.eu/avservices/play.cfm?ref=I-181645\"></iframe>"
```
or 
```
video:
  '#type': 'html_tag'
  '#tag': 'iframe'
  '#attributes':
    id: 'videoplayerI-181645'
    src: '//ec.europa.eu/avservices/play.cfm?ref=I-181645'
```
2) or, move this code to parse $structure in RenderingTest 

in case I'm wrong and this code is not just for making the test work, please do leave me steps to reproduce.

drishu (changes requested):

oe_theme.theme
sergepavle: Unfortunately, it is not just for the testing problem. In ecl twig's template, they removed '|raw' filter from the video variable. So HTML for video was with autoescaping. unfortunately just passing render array didn't work.
On 'components' page it works because the ui_pattern module automatically doing marking input data as 'safe' by applying markup objects.
For reproducing the issue as I understand you can use code in any template like that:
```
{{ pattern('text_featured_media', {
  'video': '<iframe id=\"videoplayerI-181645\" src=\"//ec.europa.eu/avservices/play.cfm?ref=I-181645\"></iframe>',
  'video_ratio': 'invalid',
}) }}
```

drishu: I see but there is no real case right now where these would be just sent as string or '#markup' instead of a field's renderable array, but we can leave it in case it is sent as '#markup' for any future case

drishu (dismissed):
Approved with one reserve https://github.com/openeuropa/oe_theme/pull/705#discussion_r578402465 but not a blocker for me

oe_theme.theme
imanoleguskiza: I agree with Andras here, we are not covering any specific case and we are opening the door for more injection security issues. I would say lets only add this when and if we need it and use one of Andras' suggestions on the tests.

imanoleguskiza (changes requested):

oe_theme.theme
sergepavle: I've checked and indeed Andras's solution work. it is not very clear why simple #markup doesn't work but html_tag element type works.

oe_theme.theme
```diff
@@ -13,6 +13,7 @@ use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Entity\Entity\EntityViewDisplay;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Element;
+use Drupal\Core\Render\Markup;
```
imanoleguskiza: Not needed.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -2331,7 +2331,12 @@
     '#id': text_featured_media
     '#fields':
       title: "Heading"
-      video: "<iframe id=\"videoplayerI-181645\" src=\"//ec.europa.eu/avservices/play.cfm?ref=I-181645\"></iframe>"
+      video:
```
imanoleguskiza, line 2334: Since we can't provide the raw video anymore, we should update the documentation in text_featured_media.ui_patterns.yml to clarify we expect a render array now.

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

drishu (dismissed):

PR #715, opened 2021-01-26
EWPP-614: Set labels in teaser view mode to be hidden.

nagyad:


modules/oe_theme_content_publication/oe_theme_content_publication.post_update.php
```diff
@@ -153,3 +153,23 @@ function oe_theme_content_publication_post_update_00008(): void {
...
+    'oe_author',
+    'oe_publication_date',
+    'oe_publication_thumbnail',
+    'oe_publication_type',
+    'oe_teaser',
+  ];
+  foreach ($fields as $field) {
+    $component = $display->getComponent($field);
+    $component['label'] = 'hidden';
+    $display->setComponent($field, $component);
```
ademarco, line 182: On a site we cannot assume these fields are on the teaser, we need to check that they are before changing the label.

modules/oe_theme_content_publication/oe_theme_content_publication.post_update.php
```diff
@@ -153,3 +153,23 @@ function oe_theme_content_publication_post_update_00008(): void {
   $display = $storage->createFromStorageRecord($display_values);
   $display->save();
 }
+
+/**
+ * Set labels in teaser view mode to be hidden.
+ */
+function oe_theme_content_publication_post_update_00009() {
+  $display = EntityViewDisplay::load('node.oe_publication.teaser');
```
ademarco, line 162: We cannot assume that the teaser is still there on a site using this module, let's first check that it exists.

ademarco (changes requested):

ademarco (dismissed):

drishu (dismissed):

ademarco (approved):

drishu (approved):

PR #721, opened 2021-02-03
EWPP-722: Add theme suggestions for text fields to replace multiple field templates.

22Alexandra:


oe_theme.theme
```diff
@@ -174,6 +174,21 @@ function oe_theme_theme_suggestions_field_alter(array &$suggestions, array $vari
...
+    $suggestions[] = 'field__text_long__' . $element['#entity_type'] . '__' . $element['#bundle'] . '__' . $field_definition->getName();
+  }
+  if ($element['#field_type'] === 'text_with_summary') {
+    /** @var \Drupal\field\Entity\FieldConfig $field_definition */
+    $field_definition = $element['#items']->getFieldDefinition();
+    // Add suggestions based on the parent entity and bundle types.
+    $suggestions[] = 'field__text_with_summary__' . $element['#entity_type'];
+    $suggestions[] = 'field__text_with_summary__' . $element['#entity_type'] . '__' . $element['#bundle'] . '__' . $field_definition->getName();
+  }
+
```
sergepavle: Those blocks of code look pretty similar, So looks like can be merged.
```suggestion
  if (in_array($element['#field_type'], ['entity_reference_revisions', 'text_long', 'text_with_summary'])) {
    /** @var \Drupal\field\Entity\FieldConfig $field_definition */
    $field_definition = $element['#items']->getFieldDefinition();
    // Add suggestions based on the parent entity and bundle types.
    $suggestions[] = 'field__' . $element['#field_type'] . '__' . $element['#entity_type'];
    $suggestions[] = 'field__' . $element['#field_type'] . '__' . $element['#entity_type'] . '__' . $element['#bundle'] . '__' . $field_definition->getName();
  }

```

templates/field/field--text-long--node.html.twig
```diff
@@ -0,0 +1,11 @@
+{#
```
sergepavle: Seems like a template for paragraph 'field--text-long--paragraph.html.twig' is missed (in this case we can remove more specific field--paragraph--field-oe-text-long.html.twig.

sergepavle: as I can see same we can do for oe-contact--full.html.twig. where we can also add config for oe_contact entity type.

sergepavle (changes requested):

sergepavle (changes requested):
we can remove previous
https://github.com/openeuropa/oe_theme/pull/721/commits/ecb2c4f7fb2b732607235538b9035483305dc533#diff-14faf6dd33cb73959ac854a38b66b254211a2a19578daa194dba5385cd13d95dR173-R174

sergepavle (dismissed):

drishu (approved):

oe_theme.theme
```diff
@@ -174,6 +174,14 @@ function oe_theme_theme_suggestions_field_alter(array &$suggestions, array $vari
     $suggestions[] = 'field__entity_reference_revisions__' . $element['#entity_type'] . '__' . $element['#bundle'] . '__' . $field_definition->getName();
   }
 
+  if (in_array($element['#field_type'], ['text_long', 'text_with_summary'])) {
```
ademarco: Why only for those two field types? The issue is about providing theme suggestion for field types in general, then to be used for those two specific types.

templates/field/field--text-with-summary--node.html.twig
```diff
@@ -0,0 +1,11 @@
...
+/**
+ * @file
+ * Theme override for the text with summary fields used in nodes.
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+<div class="ecl-editor">
+  {%- include 'field--bare.html.twig' -%}
+</div>
```
ademarco: If we do this, then every field of type text with summary will be printed without label. This is true for all the rest too.

ademarco (changes requested):

drishu (changes requested):
For testing the changes we already check field of these types for ec-editor class in functional rendering tests, but to also test for labels I suggest we create a new GenericFieldsTest where we test fields of different types for ecl formatting, what do you think @ademarco ?

oe_theme.theme
```diff
@@ -174,6 +174,14 @@ function oe_theme_theme_suggestions_field_alter(array &$suggestions, array $vari
...
+    /** @var \Drupal\field\Entity\FieldConfig $field_definition */
+    $field_definition = $element['#items']->getFieldDefinition();
+    // Add suggestions based on the parent entity and bundle types.
+    $suggestions[] = 'field__' . $element['#field_type'] . '__' . $element['#entity_type'];
```
ademarco: Can we also add this below, so we don't have to specify the entity type?
```php
    $suggestions[] = 'field__' . $element['#field_type'];
```

sergepavle (dismissed):

sergepavle (approved):

upchuk (approved):

ademarco (approved):

PR #726, opened 2021-02-11
EWPP-747: Theme vocabularies field on CFP, Events, News content types.

sergepavle:
…types for further overriding.

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/templates/content/field-group-html-element--node--oe-event--group-top.html.twig
+++ b/templates/content/field-group-html-element--node--oe-event--group-top.html.twig
@@ -21,7 +21,8 @@
     {{ element.extra_field_oe_theme_content_event_registration_button }}
   </div>
   <div class="ecl-col-12 ecl-col-md-6 ecl-u-mt-l ecl-u-mt-md-none">
-    {{ element.extra_field_oe_theme_content_event_details }}
-    {% block oe_event_extra_details %}{% endblock %}
+    {% block secondary_details %}
+      {{ element.extra_field_oe_theme_content_event_details }}
+    {% endblock %}
   </div>
 </div>
--- a/templates/content/field-group-html-element--node--oe-event--group-top.html.twig
+++ b/templates/content/field-group-html-element--node--oe-event--group-top.html.twig
@@ -22,5 +22,6 @@
   </div>
   <div class="ecl-col-12 ecl-col-md-6 ecl-u-mt-l ecl-u-mt-md-none">
     {{ element.extra_field_oe_theme_content_event_details }}
+    {% block oe_event_extra_details %}{% endblock %}
   </div>
 </div>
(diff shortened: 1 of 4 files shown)


nagyad (approved):

templates/content/field-group-html-element--node--oe-event--group-top.html.twig
```diff
@@ -22,5 +22,6 @@
   </div>
   <div class="ecl-col-12 ecl-col-md-6 ecl-u-mt-l ecl-u-mt-md-none">
     {{ element.extra_field_oe_theme_content_event_details }}
+    {% block oe_event_extra_details %}{% endblock %}
   </div>
 </div>
```
ademarco, line 30: We don't prefix block names in templates, as their scope is limited to the template itself.

Let's make this extendible like this: 

```twig
<div id="event-details" class="ecl-row">
  <div class="ecl-col-12 ecl-col-md-6">
    {% block primary_details %}
      {{ element.extra_field_oe_theme_content_event_summary }}
      {{ element.extra_field_oe_theme_content_event_registration_button }}
    {% endblock %}
  </div>
  <div class="ecl-col-12 ecl-col-md-6 ecl-u-mt-l ecl-u-mt-md-none">
    {% block secondary_details %}
      {{ element.extra_field_oe_theme_content_event_details }}
    {% endblock %}
  </div>
</div>
```
Then when extending it we use Twig's [parent function](https://twig.symfony.com/doc/3.x/functions/parent.html):

```twig
{% extends "field-group-html-element--node--oe-event--group-top.html.twig" %}
{% block secondary_details %}
  {{ parent() }}
  ...add here the extra part...
{% endblock %}
```

ademarco (changes requested):

drishu (approved):

nagyad (approved):

templates/content/node--oe-news--teaser.html.twig
```diff
@@ -6,17 +6,24 @@
...
+    {{ pattern('list_item', {
+      'variant': 'thumbnail_primary',
+      'url': url,
+      'title': label,
+      'meta': meta,
+      'detail': detail,
+      'image': image_url ? { 'src': image_url },
+      'additional_information': additional_information,
+    }) }}
+  {% endspaceless %}
```
ademarco: Move the spaceless outside the block, to wrap the whole template, as it was done before.

ademarco (changes requested):

ademarco (approved):

templates/content/node--oe-news--teaser.html.twig
sergepavle: Moved out spaceless.

PR #728, opened 2021-02-12
EWPP-421: Do not print additional information item wrapper if the item itself is empty.

brummbar:


ademarco (approved):

nagyad (approved):

templates/patterns/list_item/pattern-list-item--variant-date.html.twig
```diff
@@ -46,7 +46,9 @@
     {% endif %}
     {% if additional_information is not empty %}
       {% for item in additional_information %}
-        <div class="ecl-content-item__additional_information ecl-u-mb-s">{{ item }}</div>
+        {% if item|render|striptags|trim is not empty %}
+          <div class="ecl-content-item__additional_information ecl-u-mb-s">{{ item }}</div>
```
sergepavle, line 58: This looks a little bit tricky. Here are changes in a pattern that can be used by any custom site.
As I understand after applying these changes usage of this pattern with HTML inside additional_information can't be possible.
For instance, if somebody will put #markup with HTML like ``.

ademarco: You are right @sergepavle, we keep stumbling on this somehow.

ademarco (dismissed):
Blocking after @sergepavle 's remarks.

sergepavle (approved):

nagyad (approved):

PR #733, opened 2021-02-23
EWPP-752: Theme Description List paragraph.

22Alexandra:


tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -635,4 +636,85 @@ public function testFactsFigures(): void {
...
+    ]);
+
+    $paragraph->save();
+    $html = $this->renderParagraph($paragraph);
+    $crawler = new Crawler($html);
+
+    $this->assertCount(1, ($crawler->filter('h2.ecl-u-type-heading-2')));
+    $this->assertEquals('Overview', trim($crawler->filter('h2.ecl-u-type-heading-2')->text()));
+    $this->assertCount(1, $crawler->filter('dl.ecl-description-list.ecl-description-list--horizontal.ecl-description-list--featured'));
+    $this->assertCount(5, $crawler->filter('dt.ecl-description-list__term'));
```
drishu, line 698: can't we assert here using the FieldListAssert class? if no, please detail.

templates/paragraphs/paragraph--oe-description-list.html.twig
```diff
@@ -0,0 +1,19 @@
+{#
+/**
+ * @file
+ * Theme override to display the 'Description list' type paragraph.
+ *
+ * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
```
drishu: shouldn't this extend templates/paragraphs/paragraph.html.twig? thinking out loud, I see the other templates for paragraphs also don't do this, but why?

templates/paragraphs/paragraph--oe-description-list.html.twig
```diff
@@ -0,0 +1,19 @@
...
+ */
+#}
```
drishu: we could wrap this in {% block paragraph %} and the content part in {% block content %} for anyone wanting to extend this further (as i's done also in the contrib modules template)

drishu (changes requested):

templates/paragraphs/paragraph--oe-description-list.html.twig
22Alexandra: I'm not sure I understand what you mean. The template we extend in paragraphs templates comes from the Paragraphs module which is a contrib module, is not ours. Our module is oe_paragraphs.
The same applies for the other comment related to the template.

tests/Kernel/Paragraphs/ParagraphsTest.php
22Alexandra: We cannot use the FieldListAssert because we need to get the formatted description for each item.
This assertion for example https://github.com/openeuropa/oe_theme/pull/733/files#diff-20f9f6ae60aacc67f4bc0dbe305153caa31575ccc3256873e83597b1c69b3778R711 would be `'<p>Description 2</p>'` in the pattern assert class.

templates/paragraphs/paragraph--oe-description-list.html.twig
drishu: we extend the paragraphs base template in @oe_theme/paragraphs/paragraph.html.twig, was saying since we do that we should extend that one for variants.
But since we dont for the others neither I guess we can consider our paragraph.html.twig as a fallback

tests/Kernel/Paragraphs/ParagraphsTest.php
drishu: we discussed we will improve in a followup for checking markup instead of plain text.
The main idea is to improve the tests so that instead of checking values like
```
'body' => 'Description 5', 
```
we check for 
```
'body' => '<p>Description <strong>5</strong></p>',
```
or even sent other patter asserts as expected value

drishu (dismissed):
drone will likely fail till EWPP-768 gets resolved, so I approve as this code is ok for me
Once description_list_field is released composer will need to be amended here

drishu (dismissed):

imanoleguskiza (changes requested):
The PR is good but we need to add the ecl-editor class to the term definitions, otherwise we will have theming issues.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -635,4 +636,91 @@ public function testFactsFigures(): void {
     $this->assertCount(1, $crawler->filter('div.ecl-fact-figures.ecl-fact-figures--col-3 div.ecl-fact-figures__items'));
   }
 
+  /**
+   * Test 'Description List' paragraph rendering.
+   */
+  public function testDescriptionList(): void {
+    // Remove the auto-paragraph and url-to-link filters from the plain text.
+    /** @var \Drupal\filter\Entity\FilterFormat $format */
+    $format = FilterFormat::load('plain_text');
```
upchuk: No need to remove the filters..serves no purpose.

tests/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -635,4 +636,91 @@ public function testFactsFigures(): void {
...
+    $this->assertEquals('Term 4', trim($crawler->filter('dt.ecl-description-list__term:nth-child(7)')->html()));
+    // @todo Remove when php 7.2 is not supported anymore.
+    if (version_compare(PHP_VERSION, '7.3') < 0) {
+      $this->assertEquals('<div class="ecl-editor">Description <strong>4</strong>' . "\n" . '</div>', trim($crawler->filter('dd.ecl-description-list__definition:nth-child(8)')->html()));
+    }
+    else {
+      $this->assertEquals('<div class="ecl-editor">Description <strong>4</strong></div>', trim($crawler->filter('dd.ecl-description-list__definition:nth-child(8)')->html()));
+    }
+    $this->assertEquals('Term 5', trim($crawler->filter('dt.ecl-description-list__term:nth-child(9)')->html()));
+    $this->assertEquals('<div class="ecl-editor"><p>Description <strong>5</strong></p></div>', trim($crawler->filter('dd.ecl-description-list__definition:nth-child(10)')->html()));
```
upchuk, line 715: Let's do a reorder and assert it's correct.

upchuk (changes requested):

templates/field/description-list.html.twig
```diff
@@ -0,0 +1,30 @@
+{#
+/**
+ * @file
+ * Theme override for the description list field template of the description list paragraph type.
```
imanoleguskiza: This is no longer just for the paragraph type, update the comment.

templates/field/description-list.html.twig
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @see ./modules/contrib/description_list_field/templates/description-list.html.twig
+ *
+ * @todo Assess if the field_list pattern should be used as a follow up.
+ */
+#}
+{% set _items = [] %}
+{% if items %}
```
imanoleguskiza, line 12: if items is not empty

templates/field/description-list.html.twig
```diff
@@ -0,0 +1,30 @@
...
+    {% set _description %}
+      <div class="ecl-editor">{{ _item.description }}</div>
+    {% endset %}
+    {% set _items = _items|merge([{
+      term: _item.term,
+      definition: _description,
+    }]) %}
+  {% endfor %}
+{% endif %}
+{% if _items %}
```
imanoleguskiza, line 23: if _items is not empty

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

upchuk (approved):

imanoleguskiza (approved):

PR #738, opened 2021-02-26
EWPP-834: Drop support of dropdown legacy.

sergepavle:


ecl-builder.config.js
```diff
@@ -36,6 +36,8 @@ module.exports = {
     { from: path.resolve(nodeModules, '@ecl/ec-preset-editor/dist'), to: path.resolve(outputFolder, 'dist') },
     { from: path.resolve(nodeModules, '@ecl/ec-preset-legacy-website/dist'), to: path.resolve(outputFolder, 'dist/ec') },
     { from: path.resolve(nodeModules, '@ecl/eu-preset-legacy-website/dist'), to: path.resolve(outputFolder, 'dist/eu') },
```
ademarco: We can remove these two right?

package.json
```diff
@@ -20,11 +20,13 @@
     "@ecl/ec-component-media-container": "2.35.0",
     "@ecl/ec-preset-editor": "2.35.0",
     "@ecl/ec-preset-legacy-website": "2.35.0",
+    "@ecl/ec-preset-website": "2.35.0",
     "@ecl/ec-theme-default": "2.35.0",
     "@ecl/ec-utility-background": "2.35.0",
     "@ecl/ec-utility-border": "2.35.0",
     "@ecl/eu-component-form-help-block": "2.35.0",
     "@ecl/eu-preset-legacy-website": "2.35.0",
```
ademarco: Also from here, we can drop these two if we don't need them

templates/patterns/dropdown/pattern-dropdown.html.twig
```diff
@@ -12,8 +12,20 @@
   {% set _links = _links|merge([_link]) %}
 {% endfor %}
 
-{% include '@ecl-twig/dropdown-legacy' with {
+{% set content %}
+  <ul class="ecl-unordered-list ecl-unordered-list--no-bullet">
+    {% for item in _links %}
+      <li class="ecl-unordered-list__item">
+        {{- item.label|raw -}}
```
ademarco: Why do we use `|raw` here?

js/ecl_expandable.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @file
+ * ECL Expandable behavior for dropdown pattern.
+ */
+(function (ECL, Drupal) {
+  Drupal.behaviors.eclExpandableDropdown = {
+    attach: function attach() {
+      var dropdowns = document.querySelectorAll('[data-ecl-expandable-dropdown]');
```
ademarco: Not sure why we need this?

ademarco (changes requested):

js/ecl_expandable.js
sergepavle: 'Expandable' ECL component is doing collapsing only by click on the button. But previously used 'dropdown-legacy' ECL component had collapsing also by the loss of focus. By this js code, I applied this behavior to the 'Dropdown' pattern which now based on 'expandable' ECL component.

templates/patterns/dropdown/pattern-dropdown.html.twig
sergepavle: fixed

ecl-builder.config.js
sergepavle: fixed

package.json
sergepavle: fixed

oe_theme.theme
```diff
@@ -757,6 +757,9 @@ function oe_theme_preprocess_pattern_accordion(array &$variables): void {
  */
 function oe_theme_preprocess_pattern_dropdown(array &$variables): void {
   $variables['id'] = Html::getUniqueId('ecl-button-dropdown');
+  // Add additional javascript for reproducing the behavior of
+  // dropdown legacy ECL component.
+  $variables['#attached']['library'][] = 'oe_theme/ecl_expandable';
```
ademarco: We can add this to the pattern definition, check UI Patterns documentation on how to do it. So we don't need to add this code and on the oe_theme.libraries.yml

ademarco (changes requested):
Reopening.

22Alexandra (approved):

README.md
```diff
@@ -128,6 +128,16 @@ In 2.9.0 we dropped support for the legacy ECL site header. To do so we had to m
...
 
+### Upgrade to 2.17.0
+
+#### Dropdown UI pattern
+
+In 2.17.0 we dropped support for the `Dropdown` ui pattern which will be removed in the next major version. Few options
+which can help to mitigate the impact:
+
+- Move introduced in 2.17.0 rework of `Dropdown` implementation UI pattern in your custom component or subtheme.
+- Use ECL Expandable (as an alternative to previously used `legacy-dropdown` ECL component) component directly.
```
ademarco: Let's use this one:

```markdown
In 2.17.0 we dropped support for the `Dropdown` pattern, which will be removed in the next major version. The pattern is still available, albeit hidden on the pattern overview page. The new `Dropdown` pattern is based on the ECL Expandable component and should work as a drop-in replacement of the old one.
````

ademarco (approved):

PR #742, opened 2021-03-01
EWPP-696: Person CT theming

yenyasinn:


sass/components/_description_list.scss
```diff
@@ -7,3 +7,9 @@
   @extend .ecl-u-bg-blue-5;
   padding: $ecl-spacing-s;
 }
+
```
yenyasinn: Screenshot 2021-03-09 at 12 23 51

templates/field/field--node--oe-person-transparency-links--oe-person--full.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
```
yenyasinn, line 1: Screenshot 2021-03-09 at 12 32 40

Transparency links field is a link field that can not be built as on mockup.

templates/field/field--oe-document-reference--oe-publication--full.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
```
yenyasinn, line 1: Screenshot 2021-03-09 at 14 06 24

yenyasinn (commented):
Screenshot 2021-03-09 at 14 13 15
Full view page.

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -210,3 +217,54 @@ function template_preprocess_oe_theme_helper_call_status(array &$variables) {
...
+    $cacheability->applyTo($variables);
+    return;
+  }
+
+  $image = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_medium_no_crop');
+  $variables['image'] = $image;
+
+  $cacheability->addCacheableDependency($image);
+  $cacheability->addCacheableDependency($thumbnail->entity);
+  $cacheability->addCacheableDependency(\Drupal::entityTypeManager()->getStorage('media_type')->load($media->bundle()));
```
imanoleguskiza: Do we need to add cache tags for the media type? What will cause this tags to be invalidated that is relevant to the image?

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -210,3 +217,54 @@ function template_preprocess_oe_theme_helper_call_status(array &$variables) {
...
+ *
+ * @param array $variables
+ *   An array of variables.
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   Entity to be used to get media.
+ * @param string $field_name
+ *   Field name.
+ *
+ */
+function _oe_theme_helper_prepare_image_from_media_field(array &$variables, EntityInterface $entity, string $field_name): void {
```
imanoleguskiza: This method is very specific to the inpage navigation setting. It uses a particular variable name ("image") and style ("oe_theme_medium_no_crop") that might be different in other contexts which would make it unusable. I think we wither need to extend the method to allow for more customization or rename/refactor it in the context of inpage navigation.

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -210,3 +217,54 @@ function template_preprocess_oe_theme_helper_call_status(array &$variables) {
 function oe_theme_helper_locale_translation_projects_alter(&$projects) {
   $projects['oe_theme_helper']['info']['interface translation server pattern'] = drupal_get_path('module', 'oe_theme_helper') . '/translations/%project-%language.po';
 }
+
+/**
+ * Prepares variables for in-page navigation.
```
imanoleguskiza: This is not what this method does.

templates/content/oe-theme-helper-in-page-navigation--with-image.html.twig
```diff
@@ -0,0 +1,31 @@
+{#
```
imanoleguskiza, line 1: Why is this file inside templates/content and not modules/oe_theme_helper/templates? The related helper method is in oe_theme_helper.module

modules/oe_theme_content_person/oe_theme_content_person.module
```diff
@@ -0,0 +1,101 @@
...
+ * @param array $variables
+ *   An array of variables.
+ */
+function oe_theme_content_person_preprocess_oe_theme_helper_in_page_navigation__node__oe_person(array &$variables): void {
+  _oe_theme_helper_prepare_image_from_media_field($variables, $variables['entity'], 'oe_person_photo');
+
+  if (empty($variables['image'])) {
+    // Set default image.
+    $image_path = base_path() . drupal_get_path('theme', 'oe_theme') . '/images/user_icon.svg';
+    $variables['image'] = ImageValueObject::fromArray(['src' => $image_path]);
```
imanoleguskiza, line 30: Should we not use the oe_theme_medium_no_crop style here?

modules/oe_theme_content_person/oe_theme_content_person.module
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+declare(strict_types = 1);
+
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\oe_theme\ValueObject\ImageValueObject;
+use Drupal\oe_content_person\PersonNodeWrapper;
+
+/**
```
imanoleguskiza, line 14: This whole comment section is weird to me. We don't specify this is a hook_preprocess_HOOK, we point to the base template but we add variables that are specific to the specific template and we define the parameters which we never do in hooks.

modules/oe_theme_content_person/oe_theme_content_person.module
```diff
@@ -0,0 +1,101 @@
...
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_content_person_preprocess_field__node__oe_person_organisation__oe_person(&$variables) {
+  foreach ($variables['items'] as $key => $item) {
+    $node = $variables['items'][$key]['content']['#node'];
+    // Since Entity reference revision module doesn't provide formatter to show
+    // label of the referenced entity, we will prepare it here.
+    $variables['items'][$key]['content'] = [
+      '#markup' => $node->label(),
```
imanoleguskiza: This is a quick and dirty solution, next time we have the same situation we will need to duplicate the code. I think its fine for now but lets open a follow up to contribute a label formatter to entity reference revisions (or create our own in oe_theme)

modules/oe_theme_content_person/oe_theme_content_person.module
```diff
@@ -0,0 +1,101 @@
...
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_content_person_preprocess_field__node__oe_person_jobs__oe_person__teaser(&$variables) {
+  // Prepare person jobs to be shown in the field list pattern.
+  $view_builder = \Drupal::entityTypeManager()->getViewBuilder('oe_person_job');
+  foreach ($variables['items'] as $key => $item) {
+    $person_job = $variables['items'][$key]['content']['#oe_person_job'];
+    // Body has to be filled with at least empty space. Otherwise whole line
+    // will be hidden.
+    $body = ' ';
```
imanoleguskiza: This seems wrong to me, looking at the ecl description-list component there is no indication that we can't have a term without a description so the issue must lie somewhere else. Checking the field_list pattern we are forcing each term/description pair to have a description but afaik there is no real reason for that. @ademarco I think we should remove that arbitrary requirement from field_list

modules/oe_theme_content_person/oe_theme_content_person.module
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Implements hook_ENTITY_TYPE_view_alter().
+ */
+function oe_theme_content_person_node_view_alter(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display) {
+  if ($entity->bundle() !== 'oe_person') {
+    return;
+  }
+
+  if (in_array($build['#view_mode'], ['full', 'teaser'])) {
```
imanoleguskiza, line 42: Usually we do early returns instead of encasing logic in an if.

modules/oe_theme_content_person/oe_theme_content_person.module
```diff
@@ -0,0 +1,101 @@
...
+    // Override field labels.
+    if ($entity->get('oe_departments')->count() === 1) {
+      $build['oe_departments']['#title'] = t('Department');
+    }
+  }
+
+  if ($build['#view_mode'] === 'teaser') {
```
imanoleguskiza, line 49: Same as above.

modules/oe_theme_helper/src/Plugin/field_group/FieldGroupFormatter/PatternFormatterBase.php
```diff
@@ -131,8 +132,11 @@ public function preRender(&$element, $rendering_object) {
       ],
     ];
 
-    // Pass along the pattern render array.
-    $element = [
+    // Pass along the pattern render array but keep info about group.
```
imanoleguskiza: Lets add a comment on why we want to keep certain information about the group.

sass/components/_description_list.scss
```diff
@@ -7,3 +7,9 @@
...
+.ecl-description-list--horizontal {
```
imanoleguskiza: Not sure if this is the final solution here. We had the same issue on the horizontal description list paragraph and what we did there is we added the ecl-editor tag to the contents of the description. Also, what happens if the contents of the description are not inside a p? Would we have the issue then?

templates/content/pattern-field-list--field-group--group-organisation--node--oe-person--teaser.html.twig
```diff
@@ -0,0 +1,10 @@
+{#
+/**
+ * @file
+ * Extends field list pattern.
+ */
+#}
+{% extends '@oe_theme/patterns/field_list/pattern-field-list.html.twig' %}
+{% block content %}
+  <div class="ecl-u-mb-m">{{ parent() }}</div>
```
imanoleguskiza, line 9: Why do we need to add this? The list item where this is rendered already has a ecl-u-mb-s margin.

templates/field/field--node--oe-consultation-documents--full.html.twig
```diff
@@ -7,7 +7,7 @@
  */
 #}
 {% for item in items %}
-  <div class="ecl-u-mt-m">
+  <div class="ecl-u-mb-l">
```
imanoleguskiza, line 10: Why the increase in the margin?

templates/field/field--node--oe-person-interests-intro--oe-person--full.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Declaration of interests introduction field template .
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+{% for item in items %}
+  <div class="ecl-u-mb-l ecl-editor">
```
imanoleguskiza, line 10: Why is this large but the introduction to the biography its medium?

templates/field/field--oe-document-reference--oe-publication--full.html.twig
imanoleguskiza: Although I have nothing agains this, the current implementation seems a bit weird to me: only publications have the border at the bottom so if you have linked some document medias and THEN you add the publication you have a border at the very end of the page. It seems to me that maybe the border should be between all the items of this section, but it is not clear on the reqs so I we could clarify it at a later stage,

tests/Kernel/ContentRenderTestBase.php
```diff
@@ -180,4 +184,59 @@ protected function createMediaImage(string $name): MediaInterface {
...
+   *   Entity name. Is used as a parameter for test data.
+   * @param string $bundle
+   *   Entity bundle.
+   * @param int $status
+   *   Entity status.
+   *
+   * @return \Drupal\oe_content_entity_contact\Entity\ContactInterface
+   *   Contact entity.
+   */
+  protected function createContactEntity(string $name, string $bundle, int $status = CorporateEntityInterface::PUBLISHED): ContactInterface {
```
imanoleguskiza, line 198: Rather than duplicating this code should we not add this in a trait?

imanoleguskiza (changes requested):

imanoleguskiza (changes requested):
We need to prevent the organization contacts from being rendered if they themselves do not have a contact. SImilar to what we do here: https://github.com/openeuropa/oe_theme/pull/649/files#diff-7b4a15be3af43d244b7cd08933663ffdc848eddd70c22fa4bc85578b6c92f5a0R17

modules/oe_theme_content_person/oe_theme_content_person.module
yenyasinn: This whole comment section has been copied from https://github.com/openeuropa/oe_theme/pull/742/files/d1c1a98b608ef027839f46fc507c64b9e6d95782#diff-a58b5be76163f987bef86e1896c507d0b52f1feb602d18788444353d19e9618eL18.

modules/oe_theme_content_person/oe_theme_content_person.module
yenyasinn: We shouldn't. Image already has size that oe_theme_medium_no_crop prepares.

modules/oe_theme_content_person/oe_theme_content_person.module
yenyasinn: From conversation with @ademarco:

> Currently I'm thinking what to do with Organisation. It isn't mentioned in spec but looks like we need to show Organisation label. Issue is that entity reference revision module doesn't have formatter to show label of referenced entity.
> As a fast solution is - extra field just to show Organisation label.
but https://www.drupal.org/project/entity_reference_revisions/issues/2937835#comment-12442457 looks as more consistent approach that can be reused.

After discussion @ademarco decided to use this approach.

modules/oe_theme_content_person/oe_theme_content_person.module
yenyasinn: Up to you, guys. Reference to the issue https://github.com/openeuropa/oe_theme/blob/2.x/templates/patterns/field_list/pattern-field-list.html.twig#L10

modules/oe_theme_content_person/oe_theme_content_person.module
yenyasinn: Based on the implementations of hook_ENTITY_TYPE_view_alter() in the project it isn't.
What do you suggest?

modules/oe_theme_content_person/oe_theme_content_person.module
yenyasinn: Based on the implementations of hook_ENTITY_TYPE_view_alter() in the project it isn't.
What do you suggest?

modules/oe_theme_helper/oe_theme_helper.module
yenyasinn: Renamed to _oe_theme_helper_inpage_nav_prepare_image()

modules/oe_theme_helper/oe_theme_helper.module
yenyasinn: This is existing code from https://github.com/openeuropa/oe_theme/pull/742/files/d1c1a98b608ef027839f46fc507c64b9e6d95782#diff-a58b5be76163f987bef86e1896c507d0b52f1feb602d18788444353d19e9618eL63.

@drishu can you answer?

templates/content/oe-theme-helper-in-page-navigation--with-image.html.twig
yenyasinn: It is helper template similar field--bare.html.twig to avoid code duplication.

templates/content/pattern-field-list--field-group--group-organisation--node--oe-person--teaser.html.twig
yenyasinn: Screenshot 2021-03-10 at 20 31 52

We need it to keep height between lines the same. Structure of elements is here /admin/structure/types/manage/oe_person/display/teaser

templates/field/field--node--oe-consultation-documents--full.html.twig
yenyasinn: To make it consistent with Person. See image on the https://github.com/openeuropa/oe_theme/pull/742/files/d1c1a98b608ef027839f46fc507c64b9e6d95782#diff-c3ac261fd2860e4533d87febebc2ba8e7bcf9e0935ad90d7fb2e30ff0df3cda0

templates/field/field--node--oe-person-interests-intro--oe-person--full.html.twig
yenyasinn: Because of design
Screenshot 2021-03-10 at 20 51 11

tests/Kernel/ContentRenderTestBase.php
yenyasinn: createMediaImage() and creation of another entities can be also placed to traits. We have started to refactor tests but then it was stopped.

Please discuss with @ademarco.

modules/oe_theme_helper/oe_theme_helper.module
yenyasinn: Updated

yenyasinn:
> We need to prevent the organization contacts from being rendered if they themselves do not have a contact. SImilar to what we do here: https://github.com/openeuropa/oe_theme/pull/649/files#diff-7b4a15be3af43d244b7cd08933663ffdc848eddd70c22fa4bc85578b6c92f5a0R17

Fixed

templates/content/node--oe-person--teaser.html.twig
```diff
@@ -0,0 +1,22 @@
+{#
+/**
+ * @file
+ * Theme to display a Person node in the teaser view mode.
+ */
+#}
+{#
+{% set image_url = content.oe_organisation_logo|field_value|render %}
```
nagyad: Why do we need this here?

templates/content/oe-theme-helper-in-page-navigation--node--oe-organisation.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
+/**
+ * @file
+ * Theme override to display in-page navigation for Organisation nodes.
+ *
+ * @see ./modules/oe_theme_helper/templates/oe-theme-helper-in-page-navigation.html.twig
+ */
+#}
+{% include "oe-theme-helper-in-page-navigation--with-image.html.twig" %}
```
nagyad: Use extends instead. Otherwise other templates can't pass variables and blocks further than this.

templates/content/oe-theme-helper-in-page-navigation--node--oe-person.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
+/**
+ * @file
+ * Theme override to display in-page navigation for Person nodes.
+ *
+ * @see ./modules/oe_theme_helper/templates/oe-theme-helper-in-page-navigation.html.twig
+ */
+#}
+{% include "oe-theme-helper-in-page-navigation--with-image.html.twig" %}
```
nagyad: Use extends instead. Otherwise other templates can't pass variables and blocks further than this.

nagyad (changes requested):

modules/oe_theme_helper/oe_theme_helper.module
yenyasinn: From my point of view this cache tag is excessive. In the same time it can be here.

templates/content/oe-theme-helper-in-page-navigation--node--oe-organisation.html.twig
```diff
@@ -6,4 +6,4 @@
  * @see ./modules/oe_theme_helper/templates/oe-theme-helper-in-page-navigation.html.twig
  */
 #}
-{% include "oe-theme-helper-in-page-navigation--with-image.html.twig" %}
+{% extends "oe-theme-helper-in-page-navigation--with-image.html.twig" %}
```
nagyad: Please use proper referral, with "@oe_theme/content/oe-theme-helper-in-page-navigation--with-image.html.twig"

templates/content/oe-theme-helper-in-page-navigation--node--oe-person.html.twig
```diff
@@ -6,4 +6,4 @@
  * @see ./modules/oe_theme_helper/templates/oe-theme-helper-in-page-navigation.html.twig
  */
 #}
-{% include "oe-theme-helper-in-page-navigation--with-image.html.twig" %}
+{% extends "oe-theme-helper-in-page-navigation--with-image.html.twig" %}
```
nagyad: Please use proper referral, with "@oe_theme/content/oe-theme-helper-in-page-navigation--with-image.html.twig"

nagyad (changes requested):

modules/oe_theme_content_person/oe_theme_content_person.module
ademarco: To be clear: I decided not to use the patch, given the time constraints that we used to have when this was being worked on, I haven't actually suggested the approach used here.

modules/oe_theme_content_person/oe_theme_content_person.module
ademarco: @imanoleguskiza I agree we should remove that, as in: we should accept having a description term without a definition, and vice-versa.

modules/oe_theme_helper/oe_theme_helper.module
ademarco: We can actually use here the "oe_theme_helper_media_thumbnail_url" formatter, and we will get all the tags we need from it.

modules/oe_theme_content_person/oe_theme_content_person.module
ademarco: After discussing it with @yenyasinn we concluded that it's worth trying out the patch at comment **[#31](https://www.drupal.org/project/entity_reference_revisions/issues/2937835#comment-12442457)** (not the last one), as this will allow to remove this function completely and, possibly, also the extra field on the person job.

modules/oe_theme_content_person/src/Plugin/ExtraField/Display/PersonJobLabelExtraField.php
```diff
@@ -0,0 +1,41 @@
...
+ * @ExtraFieldDisplay(
+ *   id = "oe_theme_content_person_job_label",
+ *   label = @Translation("Person job label"),
+ *   bundles = {
+ *     "oe_person_job.oe_default",
+ *   },
+ *   visible = true
+ * )
+ */
+class PersonJobLabelExtraField extends ExtraFieldDisplayFormattedBase {
```
ademarco, line 23: As discussed with @yenyasinn we will try to remove this in favor of the entity revision reference label formatter.

ademarco (changes requested):

modules/oe_theme_content_person/src/Plugin/ExtraField/Display/PersonJobLabelExtraField.php
yenyasinn: Label is shown in the Person Job full view mode itself. So we show it as an extra field.

Screenshot 2021-03-26 at 13 00 36

sass/components/_description_list.scss
ademarco: Yep, the right way of doing this is via the ecl-editor class, as the screenshot below shows. I'm going to take care of it.

image

modules/oe_theme_content_person/src/Plugin/ExtraField/Display/PersonJobLabelExtraField.php
ademarco: Got it, thanks!

modules/oe_theme_content_person/oe_theme_content_person.module
ademarco: @imanoleguskiza 's comment is correct, the fact that was taken from somewhere else doesn't make it necessarily correct.

modules/oe_theme_content_person/oe_theme_content_person.module
ademarco: Resolving after talking with @imanoleguskiza

modules/oe_theme_content_person/oe_theme_content_person.module
ademarco: Resolving after talking with @imanoleguskiza

ademarco (approved):

imanoleguskiza (approved):

nagyad (approved):

PR #747, opened 2021-03-04
EWPP-632: Link list card container breaks if content is too long.

georgebaev:
### Description

Fixes the overflow text in the link list cards

### Change log

- Added: CSS statement


upchuk:
Not needed, fixed in ECL.

PR #755, opened 2021-03-15
EWPP-812: Add heading to timeline pattern.

22Alexandra:


oe_theme.theme
```diff
@@ -1361,6 +1361,9 @@ function oe_theme_preprocess_paragraph__oe_facts_figures(array &$variables): voi
  */
 function oe_theme_preprocess_paragraph__oe_timeline(array &$variables): void {
   $paragraph = $variables['paragraph'];
+  if (!$paragraph->get('field_oe_title')->isEmpty()) {
```
upchuk: Let's put a check here if the field exists in case oe theme is updated while oe_paragraphs doesn't yet get the field on a site.

upchuk (changes requested):

upchuk (changes requested):
Approving but blocking the merge for the full epic.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -2215,6 +2216,7 @@
           body: "Lorem Ipsum 6"
   assertions:
     count:
+      'h2.ecl-u-type-heading-2': 1
```
sergepavle, line 2222: I think it would be nice in another test to check that we don't have this h2 in case the title is empty.

sergepavle (changes requested):

sergepavle (dismissed):

sergepavle (approved):

upchuk (approved):

PR #761, opened 2021-03-18
EWPP-832: Print description in highlight variant of list item pattern.

22Alexandra:


templates/patterns/list_item/pattern-list-item--variant-highlight.html.twig
```diff
@@ -11,6 +11,7 @@
       path: url,
       type: "standalone"
     },
-    image: image
+    image: image,
+    description: detail,
```
yenyasinn:
```suggestion
    description: detail|smart_trim(length),
```
There is `length` parameter in the interface of List item pattern - "Used to trim the length of the 'Detail' field (optional).". I think we should support it here also. If length is not provided full description is shown. 
See another variants of List item patterns.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -1083,6 +1083,27 @@
       article.ecl-card div.ecl-card__body div.ecl-card__description: 0
     equals:
       article.ecl-card header.ecl-card__header h1.ecl-card__title .ecl-link: "Item title"
+- array:
```
yenyasinn, line 1077: If we support length variable I would suggest to add one more test to check trimming of description field.

22Alexandra: I updated the existing one. Doesn't make any difference if there are two tests checking for description.

yenyasinn (dismissed):

nagyad (dismissed):

nagyad (approved):

yenyasinn (approved):

PR #766, opened 2021-03-26
EWPP 855: Create in-page visibility plugin for the block.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
+++ b/modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
@@ -12,13 +12,13 @@ use Drupal\oe_theme_inpage_navigation\InPageNavigationHelper;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
- * Provides a configurable 'Inpage Navigation State' condition.
+ * Provides a configurable 'In-page Navigation State' condition.
  *
- * This condition checks if the inpage navigation for node is enabled.
+ * This condition checks if the in-page navigation is enabled for the node.
  *
  * @Condition(
  *   id = "oe_theme_inpage_navigation_state",
- *   label = @Translation("Inpage navigation"),
+ *   label = @Translation("In-page navigation"),
  *   context_definitions = {
  *     "node" = @ContextDefinition("entity:node", required = FALSE)
  *   }
@@ -42,7 +42,7 @@ class InPageNavigationState extends ConditionPluginBase implements ContainerFact
    */
   public function defaultConfiguration() {
     return [
-      'inpage_navigation_state' => NULL,
+      'inpage_navigation_state' => FALSE,
     ] + parent::defaultConfiguration();
   }
 
@@ -80,10 +80,6 @@ class InPageNavigationState extends ConditionPluginBase implements ContainerFact
       return $this->isNegated();
     }
 
-    if ($this->configuration['inpage_navigation_state'] === NULL) {
-      return TRUE;
-    }
-
     return InPageNavigationHelper::isInPageNavigationEnabled($node) === $this->configuration['inpage_navigation_state'];
   }
 
@@ -91,16 +87,12 @@ class InPageNavigationState extends ConditionPluginBase implements ContainerFact
    * {@inheritdoc}
    */
   public function summary() {
-    if (is_null($this->configuration['inpage_navigation_state'])) {
-      return $this->t('Any inpage navigation state');
-    }
-
     $state = $this->configuration['inpage_navigation_state'] ? $this->t('enabled') : $this->t('disabled');
     if ($this->isNegated()) {
-      return $this->t('The inpage navigation should not be @state', ['@state' => $state]);
+      return $this->t('The in-page navigation should not be @state', ['@state' => $state]);
     }
 
-    return $this->t('The inpage navigation should be @state', ['@state' => $state]);
+    return $this->t('The in-page navigation should be @state', ['@state' => $state]);
   }
 
 }
--- a/modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
+++ b/modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
@@ -84,7 +84,7 @@ class InPageNavigationState extends ConditionPluginBase implements ContainerFact
       return TRUE;
     }
 
-    return InPageNavigationHelper::isInPageNavigation($node) === $this->configuration['inpage_navigation_state'];
+    return InPageNavigationHelper::isInPageNavigationEnabled($node) === $this->configuration['inpage_navigation_state'];
   }
 
   /**
--- a/modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
+++ b/modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
@@ -1,0 +1,101 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_theme_inpage_navigation\Plugin\Condition;
+
+use Drupal\Core\Condition\ConditionPluginBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\node\NodeInterface;
+use Drupal\oe_theme_inpage_navigation\InPageNavigationHelper;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a configurable 'Inpage Navigation State' condition.
+ *
+ * This condition checks if the inpage navigation for node is enabled.
+ *
+ * @Condition(
+ *   id = "oe_theme_inpage_navigation_state",
+ *   label = @Translation("Inpage navigation"),
+ *   context_definitions = {
+ *     "node" = @ContextDefinition("entity:node", required = FALSE)
+ *   }
+ * )
+ */
+class InPageNavigationState extends ConditionPluginBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return [
+      'inpage_navigation_state' => NULL,
+    ] + parent::defaultConfiguration();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form['inpage_navigation_state'] = [
+      '#type' => 'radios',
+      '#title' => $this->t('Inpage navigation'),
+      '#default_value' => $this->configuration['inpage_navigation_state'],
+      '#options' => [$this->t('Disabled'), $this->t('Enabled')],
+      '#description' => $this->t('Choose with which inpage navigation state this condition should be met.'),
+    ];
+    return parent::buildConfigurationForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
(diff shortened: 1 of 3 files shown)


modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,283 @@
...
+   * @param array|null $inpage_navigation_node_state
+   *   Inpage navigation state in node.
+   * @param array $expected
+   *   Array with expected values of 'summary' and 'result'.
+   *
+   * @dataProvider providerTestCondition
+   */
+  public function testCondition(?bool $inpage_navigation_condition, ?bool $inpage_navigation_condition_negate, ?array $inpage_navigation_node_state, array $expected): void {
+    $manager = \Drupal::service('plugin.manager.condition');
+    $condition = $manager->createInstance('oe_theme_inpage_navigation_state');
```
yenyasinn:
```suggestion
    $condition = $this->container->get('plugin.manager.condition')->createInstance('oe_theme_inpage_navigation_state');
```

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,283 @@
...
+    $condition->setConfiguration(['inpage_navigation_state' => $inpage_navigation_condition, 'negate' => $inpage_navigation_condition_negate]);
+
+    if ($inpage_navigation_node_state) {
+      $condition->setContextMapping([
+        'node' => 'node',
+      ]);
+      NodeType::create(['type' => 'example', 'name' => 'Example'])->save();
+      \Drupal::service('emr.installer')->installEntityMetaTypeOnContentEntityType('oe_theme_inpage_navigation', 'node', ['example']);
```
yenyasinn:
```suggestion
     $this->container->get('emr.installer')->installEntityMetaTypeOnContentEntityType('oe_theme_inpage_navigation', 'node', ['example']);
```

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,283 @@
...
+  public function testCondition(?bool $inpage_navigation_condition, ?bool $inpage_navigation_condition_negate, ?array $inpage_navigation_node_state, array $expected): void {
+    $manager = \Drupal::service('plugin.manager.condition');
+    $condition = $manager->createInstance('oe_theme_inpage_navigation_state');
+    $condition->setConfiguration(['inpage_navigation_state' => $inpage_navigation_condition, 'negate' => $inpage_navigation_condition_negate]);
+
+    if ($inpage_navigation_node_state) {
+      $condition->setContextMapping([
+        'node' => 'node',
+      ]);
+      NodeType::create(['type' => 'example', 'name' => 'Example'])->save();
```
yenyasinn: Provided code is correct but please consider using build/core/modules/node/tests/src/Traits/ContentTypeCreationTrait.php and build/core/modules/node/tests/src/Traits/NodeCreationTrait.php:
```
    $this->drupalCreateContentType([
      'type' => 'page',
      'name' => 'Page',
    ]);
$node = $this->drupalCreateNode();
```
because it is commonly used way to create test nodes in contrib modules.

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,283 @@
...
+      \Drupal::service('emr.installer')->installEntityMetaTypeOnContentEntityType('oe_theme_inpage_navigation', 'node', ['example']);
+      $node = Node::create(['type' => 'example', 'title' => 'some title']);
+      $node->save();
+      if ($inpage_navigation_node_state['inpage_navigation']) {
+        InPageNavigationHelper::setInPageNavigation($node);
+      }
+      $contexts['node'] = EntityContext::fromEntity($node);
+      \Drupal::service('context.handler')->applyContextMapping($condition, $contexts);
```
yenyasinn:
```suggestion
      $this->container->get('context.handler')->applyContextMapping($condition, $contexts);
```

modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form['inpage_navigation_state'] = [
+      '#type' => 'radios',
+      '#title' => $this->t('Inpage navigation'),
+      '#default_value' => $this->configuration['inpage_navigation_state'],
+      '#options' => [$this->t('Disabled'), $this->t('Enabled')],
```
upchuk: Let's put these on two lines like everywhere else in Drupal

modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
+    $node = $this->getContextValue('node');
+
+    if (!$node instanceof NodeInterface) {
+      return $this->isNegated();
```
upchuk, line 85: Let's put a comment here:
```
If we don't have a node, we return FALSE, unless the plugin is configured to be negated, in which case, it's the opposite```

It makes it a bit easier to follow.

File: modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php [hunk 7]
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    return InPageNavigationHelper::isInPageNavigationEnabled($node) === $this->configuration['inpage_navigation_state'];
```
upchuk: Here you do === but from what I see, you are storing 1 or 0 for the state configuration. Also, don''t we care about `isNegated()` here?

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,227 @@
...
+ */
+class InPageNavigationStateTest extends AbstractKernelTestBase {
+
+  use ContentTypeCreationTrait;
+  use NodeCreationTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = [
```
upchuk:
```
protected static $modules = [
```

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,227 @@
...
+   * @param bool|null $inpage_navigation_condition
+   *   Inpage navigation state condition.
+   * @param bool|null $inpage_navigation_condition_negate
+   *   Is negate.
+   * @param array|null $inpage_navigation_node_state
+   *   Inpage navigation state in node.
+   * @param array $expected
+   *   Array with expected values of 'summary' and 'result'.
+   *
+   * @dataProvider providerTestCondition
```
upchuk: This means we need to run instead of 1 kernel test, 12 kernel tests (if i counted correctly). The theme build takes 556 hours aleady to complete so we need to optimize here.

All I see the test is doing is instantiating the plugin, creating aa node and asserting the evaluation. Let's instead loop through all the cases and run them all in the same test by reinstantiating the plugin and deleting the node after the assertions.

upchuk (changes requested):

modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
sergepavle: It is handled in this method: \Drupal\Core\Condition\ConditionManager::execute

upchuk: OK, so if `ConditionManager::execute` handles that, why if there is no Node available, we return `isNegated()`? Shouldn't we return FALSE and if it is negated, the condition manager would return true?

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,220 @@
...
+    }
+  }
+
+  /**
+   * Data provider for testCondition().
+   *
+   * @return array[]
+   *   The test data.
+   */
+  public function providerTestCondition(): array {
```
upchuk: This is no longer a data provider so it should not be called as such, in the method nor in the comment above. Instead, a better comment can be written:

```
Provides a series of test scenarios for the InPageNavigationState condition plugin.
@see self::testCondition()
```

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Tests the inpage navigation state condition.
+   */
+  public function testCondition(): void {
+    /** @var $condition \Drupal\Core\Condition\ConditionPluginBase */
+    $condition = $this->container->get('plugin.manager.condition')->createInstance('oe_theme_inpage_navigation_state');
+    $node = $this->createNode(['type' => 'example', 'title' => 'some title']);
+    foreach ($this->providerTestCondition() as $test_case_name => $test_case) {
+      $condition_instance = clone $condition;
```
upchuk: Why are you cloning and not just instantiating here?

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
```diff
@@ -0,0 +1,220 @@
...
+      $node_instance = clone $node;
```
upchuk: You can just create the node here and on line 81 delete the node instead of the meta. the meta gets deleted two with it.

upchuk (changes requested):

modules/oe_theme_inpage_navigation/src/Plugin/Condition/InPageNavigationState.php
sergepavle: I expected that if we don't have an available node in context, the conditional plugin will return always FALSE.
If we return just FALSE, evaluate() method return TRUE in this test case: `enabled in-page navigation, negative, no node`.  
I used isNegated() method to enforce `ConditionManager::execute` method always returns FALSE if we don't have node object in context.

upchuk: ok

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
sergepavle: Updated

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
sergepavle: Updated

modules/oe_theme_inpage_navigation/tests/src/Kernel/Plugin/Condition/InPageNavigationStateTest.php
sergepavle: Updated

upchuk (approved):

hernani (approved):

PR #772, opened 2021-04-07
EWPP-856: Create inpage navigation block.

sergepavle:


modules/oe_theme_inpage_navigation/js/inpage_navigation.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * @file
+ * ECL inpage navigation initializer.
+ */
+(function (ECL, Drupal) {
```
hernani: We need to do two extra things in this function:
- In case the h2.ecl-u-type-heading-2 has no id, it should be generated and added so that it can be used by the inpage nav.
- We need to loop first in regions marked with a certain data attribute (e.g. data-inpage-navigation-provider) and mark all found h2 with the ecl-u-type-heading-2 class. Afterwards they should be found and taken in consideration by the current code and used in the in-page navigation.

templates/layout/page.html.twig
```diff
@@ -38,7 +38,18 @@
   <div class="ecl-container">
     <div class="ecl-row">
       <div class="ecl-col-sm-12">
-        {{ page.content }}
+        {% if page.left_sidebar %}
```
hernani: Let's not handle the logic regarding columns and new region directly in this PR. In terms of current regions, just guarantee the data attribute is added to the content region, so that h2 elements are taken in consideration.

templates/layout/region--content.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
```
hernani: Not needed with new approach ?

oe_theme.info.yml
```diff
@@ -17,6 +17,7 @@ regions:
   site_header_secondary: "Site header secondary"
   page_header: "Page header"
   navigation: "Navigation"
+  left_sidebar: "Left sidebar"
```
hernani: Not needed with new approach.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
```diff
@@ -0,0 +1,175 @@
...
+
+  /**
+   * Test in-page entity meta relation plugin.
+   */
+  public function testInPageNavigationForm(): void {
+
+    // Create a document for Publication.
+    $media_document = $this->createMediaDocument('publication_document');
+
+    // Create a Publication node with required fields only.
```
hernani: Let's test also a content type without inpage navigation elements like page and test that h2 added in content regions are correctly handled.

modules/oe_theme_inpage_navigation/tests/modules/oe_theme_inpage_navigation_test/oe_theme_inpage_navigation_test.module
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types = 1);
+
+/**
+ * Implements hook_theme_suggestions_HOOK().
+ */
+function oe_theme_inpage_navigation_test_theme_suggestions_oe_theme_helper_in_page_navigation_alter(&$suggestions, $variables) {
+  $suggestions[] = $variables['theme_hook_original'] . '__test_overridden';
```
hernani: Not required with new approach?

modules/oe_theme_inpage_navigation/js/inpage_navigation.js
sergepavle: I've added autogeneration of ids for headers without ID attribute and `data-inpage-navigation-source-area` attribute for content block.

templates/layout/page.html.twig
sergepavle: Reverted. Changes in page layout will go to ewcms.

templates/layout/region--content.html.twig
sergepavle: Removed

oe_theme.info.yml
sergepavle: Removed

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
sergepavle: Added additional test.

modules/oe_theme_inpage_navigation/tests/modules/oe_theme_inpage_navigation_test/oe_theme_inpage_navigation_test.module
sergepavle: Removed.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
```diff
@@ -0,0 +1,302 @@
...
+    $ct_with = NodeType::load('oe_publication');
+    $ct_with->setThirdPartySetting('oe_theme_inpage_navigation', 'enabled', TRUE);
+    $ct_with->save();
+  }
+
+  /**
+   * Test content with enabled legacy in-page navigation by default.
+   */
+  public function testContentWithInPageNav(): void {
+
```
brummbar: No empty lines at the beginning of functions.

modules/oe_theme_inpage_navigation/templates/oe-theme-inpage-navigation-block.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Theme override to display in-page navigation for nodes.
```
brummbar: #doubt

modules/oe_theme_inpage_navigation/src/Plugin/Block/InpageNavigationBlock.php
```diff
@@ -0,0 +1,31 @@
...
+namespace Drupal\oe_theme_inpage_navigation\Plugin\Block;
+
+use Drupal\Core\Block\BlockBase;
+
+/**
+ * Provides an In-page navigation block.
+ *
+ * @Block(
+ *   id = "oe_theme_inpage_navigation_menu",
+ *   admin_label = @Translation("Inpage Navigation block"),
```
brummbar:
```suggestion
 *   admin_label = @Translation("Inpage navigation block"),
```

modules/oe_theme_inpage_navigation/js/inpage_navigation.js
```diff
@@ -0,0 +1,43 @@
...
+      const headers = document.querySelectorAll(headers_selector);
+      const li_html = [];
+      for (let h = 0; h < headers.length; h++) {
+        let title = headers[h].textContent.trim();
+        if (!headers[h].hasAttribute('id')) {
+          // Set id if not available yet in h2 tag.
+          headers[h].setAttribute('id', title.replace(/\W/g,'-').toLowerCase())
+        }
+        let id = $(headers[h]).uniqueId().attr('id');
+        li_html.push('<li class="ecl-inpage-navigation__item"><a href="#' + id + '" class="ecl-link ecl-link--standalone ecl-inpage-navigation__link" data-ecl-inpage-navigation-link="">' + title + '</a></li>');
```
brummbar: Drupal has the Drupal.theme function which allows to "theme" something. So we can use
```
Drupal.theme.oe_theme_inpage_navigation_item = function (...params) {
  return '<div>my html</div>';
}
```

which you can call after with `Drupal.theme('oe_theme_inpage_navigation_item')`.

modules/oe_theme_inpage_navigation/js/inpage_navigation.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * @file
+ * ECL inpage navigation initializer.
+ */
+(function (ECL, Drupal, $) {
+  Drupal.behaviors.eclInPageNavigation = {
+    attach: function attach(context, settings) {
+      // List of heading selectors.
+      const selectors = [];
```
brummbar: Drupal uses ES6 for JS development but then transpiles everything. Files without `.es6.js` should be plain JS.

modules/oe_theme_inpage_navigation/js/inpage_navigation.js
```diff
@@ -0,0 +1,43 @@
...
+      })
+      let headers_selector = selectors.join(', ');
+      // Run query on all required selectors.
+      const headers = document.querySelectorAll(headers_selector);
+      const li_html = [];
+      for (let h = 0; h < headers.length; h++) {
+        let title = headers[h].textContent.trim();
+        if (!headers[h].hasAttribute('id')) {
+          // Set id if not available yet in h2 tag.
+          headers[h].setAttribute('id', title.replace(/\W/g,'-').toLowerCase())
```
brummbar: The creation of the ID is not unique unfortunately. We should make sure that we don't have twice the same id. 
Also I've checked what Github does and found some implementations for the slug generation:

- https://github.com/markedjs/marked/blob/master/src/Slugger.js : very similar to what Github does, minus emoji stripping, which we could ignore for now. This would have to be copied over in our code;
- https://github.com/simov/slugify : comes as JS standalone, but also encodes a lot of characters, which others like Github don't do. I don't like much the transliteration;
- https://github.com/Flet/github-slugger : similar to the code in Slugger.js above but it also does emoji removal, similar to Github. We would need to run browserify on it, to have it as standalone JS.

I think we should go for the first option for now, having that library locally.

modules/oe_theme_inpage_navigation/js/inpage_navigation.js
```diff
@@ -0,0 +1,43 @@
...
+(function (ECL, Drupal, $) {
+  Drupal.behaviors.eclInPageNavigation = {
+    attach: function attach(context, settings) {
+      // List of heading selectors.
+      const selectors = [];
+      // Collect heading selectors.
+      document.querySelectorAll('[data-inpage-navigation-source-area]').forEach(function (element) {
+        let header_selector = element.getAttribute('data-inpage-navigation-source-area');
+        selectors.push('[data-inpage-navigation-source-area="' + header_selector + '"] ' + header_selector);
+      })
```
brummbar: At first I didn't understand why build a single selector instead of running multiple queries, but I believe it was done to benefit of `document.querySelectorAll()` that returns non-duplicate elements in the correct (page) order. The problem is that a header selector could contain an escaped character, like `"`, and there is no easy way to re-escape them to generate the selector. Additionally, the above code doesn't take into account multiple selectors passed to the attribute, like `h2, h3`.
Due to this, we can resort to something like this to get the job done:

```
// Loop through all the elements marked as source areas.
Array.prototype.forEach.call(document.querySelectorAll('[data-inpage-navigation-source-area]'), function (area) {
  var selectors = area.getAttribute('data-inpage-navigation-source-area');

  // Loop through all the elements that are referenced by the specified selector(s), and mark them as source
  // elements. We cannot collect the elements at this stage, as multiple nested areas can be present in the page.
  // This could lead to scenarios where elements are collected multiple times, or not collected following the
  // order of appearance in the page.
  // The :scope pseudo-class is needed to make sure that the selectors are applied inside the parent.
  // @see https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll#user_notes
  Array.prototype.forEach.call(area.querySelectorAll(':scope ' + selectors), function (element) {
    element.setAttribute('data-inpage-navigation-source-element', '');
  });
});

// Collect all the elements marked as source. Now the elements will be unique and ordered correctly.
Array.prototype.forEach.call(document.querySelectorAll('[data-inpage-navigation-source-element]'), function (element) {
  var title = element.textContent.trim();

  // Generate an unique ID if not present.
  if (!element.hasAttribute('id')) {
    element.setAttribute('id', slug(title));
  }

  // Cleanup the markup from the helper attribute added above.
  element.removeAttribute('data-inpage-navigation-source-element');

  items_markup += Drupal.theme('oe_theme_inpage_navigation', element.getAttribute('id'), title);
});
```

modules/oe_theme_inpage_navigation/js/inpage_navigation.js
```diff
@@ -0,0 +1,43 @@
...
+        let id = $(headers[h]).uniqueId().attr('id');
+        li_html.push('<li class="ecl-inpage-navigation__item"><a href="#' + id + '" class="ecl-link ecl-link--standalone ecl-inpage-navigation__link" data-ecl-inpage-navigation-link="">' + title + '</a></li>');
+      }
+
+      // In-page navigation blocks.
+      const elements = document.querySelectorAll('[data-ecl-inpage-navigation]');
+      for (let i = 0; i < elements.length; i++) {
+        if (li_html.length === 0) {
+          // Adjust layout if we going to remove in-page element.
+          $(elements[i]).closest('.ecl-col-lg-3').next('.ecl-col-lg-9').removeClass('ecl-col-lg-9').addClass('ecl-col-lg-12')
```
brummbar: We are hardcoding here the classes and expected behaviour. We could leave it like this for now, but we need to remove these assumptions from here.

modules/oe_theme_inpage_navigation/tests/modules/oe_theme_inpage_navigation_test/config/install/filter.format.full_html.yml
```diff
@@ -0,0 +1,27 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - media
```
brummbar: This misses the dependency on the test module. Minor, but when I tested and did a uninstall-reinstall, it failed.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
```diff
@@ -0,0 +1,302 @@
...
+        'oe_subject' => 'http://data.europa.eu/uxp/1000',
+        'oe_author' => 'http://publications.europa.eu/resource/authority/corporate-body/AASM',
+        'oe_content_content_owner' => 'http://publications.europa.eu/resource/authority/corporate-body/COMMU',
+        'uid' => 0,
+        'status' => 1,
+      ]);
+    $node->save();
+    $this->drupalGet($node->toUrl());
+
+    // Assert content part.
```
brummbar: Let's improve these comments: Assert that the main page content block is marked as source area.
But I don't understand what we are testing here, the article? Or the data attribute.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
```diff
@@ -0,0 +1,302 @@
...
+    $inpage_nav_expected_values = [
+      'title' => 'Page contents',
+      'list' => [
+        ['label' => 'Details', 'href' => '#details'],
+        ['label' => 'Files', 'href' => '#files'],
+      ],
+    ];
+    $inpage_nav_assert->assertPattern($inpage_nav_expected_values, $navigation->getOuterHtml());
+
+    // Add body field to node and check in-page navigation links.
```
brummbar: To improve comment.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
```diff
@@ -0,0 +1,302 @@
...
+    $node->save();
+    $this->drupalGet($node->toUrl());
+    $inpage_nav_expected_values['list'] = [
+      ['label' => 'Details', 'href' => '#details'],
+      ['label' => 'Description', 'href' => '#description'],
+      ['label' => 'Files', 'href' => '#files'],
+    ];
+    $inpage_nav_assert->assertPattern($inpage_nav_expected_values, $navigation->getOuterHtml());
+
+    // Remove body field value of node and check in-page navigation links.
```
brummbar: Is this needed? I don't think it has value.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
```diff
@@ -0,0 +1,302 @@
...
+    $node->set('body', NULL);
+    $node->save();
+    $this->drupalGet($node->toUrl());
+    $inpage_nav_expected_values['list'] = [
+      ['label' => 'Details', 'href' => '#details'],
+      ['label' => 'Files', 'href' => '#files'],
+    ];
+    $inpage_nav_assert->assertPattern($inpage_nav_expected_values, $navigation->getOuterHtml());
+
+    // Remove body field value of node and check in-page navigation links.
```
brummbar: Comment is copied from previous one.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
```diff
@@ -0,0 +1,302 @@
...
+      'title' => 'Page contents',
+      'list' => [
+        ['label' => 'Heading from body field', 'href' => '#heading-from-body-field'],
+        ['label' => 'Related links', 'href' => '#related-links'],
+      ],
+    ];
+    $inpage_nav_assert->assertPattern($inpage_nav_expected_values, $navigation->getOuterHtml());
+
+    // Check that in-page navigation block count presence of
+    // h2.ecl-u-type-heading-2 elements inside content region.
```
brummbar: I don't think this should be tested like this. I'd add some custom markup in the test page template, and verify that everything is working. The fact that oe_theme uses this class for blocks should not matter in the test.

brummbar (changes requested):

modules/oe_theme_inpage_navigation/js/inpage_navigation.js
sergepavle: incorporating the first solution was possible by a copy of the code. It wasn't possible to use inside this js import of external module.
Looks like the incorporation of js possible by custom including of js as a module.

modules/oe_theme_inpage_navigation/js/inpage_navigation.js
sergepavle: It will be implemented by follow up ticket.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
sergepavle: I've removed this as it doesn't bring big value. Anyway attribute already in place and done within templates.

modules/oe_theme_inpage_navigation/tests/src/FunctionalJavascript/InPageNavigationBlockTest.php
sergepavle: I used block with heading not only for checking that it has an influence on in-page navigation but also to ensure that we have disappeared in-page navigation block if we don't have any more headings.
Do you think this test case can be also?

hernani (dismissed):

hernani (approved):

brummbar (approved):

PR #776, opened 2021-04-12
EWPP-969: Move Social media links outside Media fieldgroup on Person full view display.

22Alexandra:


tests/Functional/ContentPersonRenderTest.php
```diff
@@ -253,16 +253,6 @@ public function testPersonRendering(): void {
     ])->save();
     $this->drupalGet($node->toUrl());
 
-    $inpage_nav_expected_values['list'][] = [
```
imanoleguskiza, line 256: We removed this but we need to add it whenever we test the media items.

imanoleguskiza (changes requested):

imanoleguskiza (approved):

imanoleguskiza (dismissed):

ademarco (dismissed):
We need to provide an upgrade path to update the view mode, as the theme is already released.

brummbar (approved):
Blind

imanoleguskiza (approved):

PR #779, opened 2021-04-16
EWPP-987: Adds overview field group to full view display.

22Alexandra:


modules/oe_theme_content_organisation/config/install/core.entity_view_display.node.oe_organisation.full.yml
```diff
@@ -48,13 +52,24 @@ third_party_settings:
...
       parent_name: group_in_page_navigation
-      weight: 2
+      weight: 3
       format_type: oe_theme_helper_in_page_navigation_item
       region: content
       format_settings:
         id: ''
         classes: ''
       label: Contact
+    group_overview:
```
upchuk, line 62: Don't we need an update path here as well? Since we did one in oe_content?

upchuk (changes requested):

modules/oe_theme_content_organisation/config/install/core.entity_view_display.node.oe_organisation.full.yml
22Alexandra: I'd say we update in ewcms the display only since we would override for everybody and there might be already changes done by custom sites, no?

upchuk: Discussed with @ademarco and yeah, while this is true, people run this locally and if they dont want the changes, they can easily revert. So let's do it.

upchuk (approved):

sergepavle (approved):

sergepavle (approved):

PR #787, opened 2021-04-22
EWPP-989: Theme "Leadership and organisation" section.

22Alexandra:

PR #789, opened 2021-04-22
EWPP-1044: Include the Link field in the Contact rendering

yenyasinn:


tests/Kernel/ContactRenderTest.php
```diff
@@ -250,4 +256,26 @@ protected function assertFeaturedMediaField(string $html, string $name): void {
...
+   * Asserts links in the bottom of full view rendering.
+   *
+   * @param \Symfony\Component\DomCrawler\Crawler $element
+   *   Link element.
+   * @param string $link
+   *   Expected href attribute.
+   * @param string $title
+   *   Expected link title.
+   */
+  protected function assertBottomLink(Crawler $element, string $link, string $title) {
```
upchuk: :void

modules/oe_theme_content_entity_contact/config/install/core.entity_view_display.oe_contact.oe_press.full.yml
```diff
@@ -87,12 +89,24 @@ content:
...
     region: content
     label: hidden
     settings:
       image_style: ''
     third_party_settings: {  }
+  oe_link:
+    type: link
+    weight: 4
+    region: content
+    label: above
```
upchuk: label: hidden if you wanna keep things consistent with oe_press_contact_url

upchuk (changes requested):

upchuk (dismissed):

templates/oe_contact/oe-contact--full.html.twig
```diff
@@ -21,8 +21,13 @@
       </div>
     {% endif %}
     {{ content.group_contact_information }}
-    {{ content.oe_press_contact_url }}
-  </div>
+    {% if entity.oe_press_contact_url is not empty or entity.oe_link is not empty %}
```
ademarco: Since we are using the entity object here, we can use `.isEmpty()` method, unless `is not empty` somehow forwards the check to that method, but I doubt it.

```suggestion
    {% if not entity.oe_press_contact_url.isEmpty() or not entity.oe_link.isEmpty() %}
```

templates/field/field--oe-contact--oe-link--full.html.twig
```diff
@@ -0,0 +1,21 @@
...
+{% for item in items %}
+  <div class="ecl-u-border-bottom ecl-u-border-color-grey-15 ecl-u-pt-l ecl-u-pb-l">
+    {% include '@ecl-twig/link' with {
+      link: {
+        type: 'standalone',
+        label: item.content['#title'],
+        path: item.content['#url'],
+        icon_position: 'after'
+      },
+      icon: get_link_icon(item.content['#url'].toString())
```
ademarco, line 19: By accessing link render array portions we don't allow cache to bubble up. The easiest way to fix it is to render the link and assign the output to a catch all variable, that will never be displayed, like that:

```twig
   {% set bubble_cache = item.content|render %}
```

Check how it was done in [this PR](https://github.com/openeuropa/oe_theme/pull/787/files#diff-e8ef914d08dc95ae7d28b49062142ef7e0181ea1dff787b83d2a5cbd5cf7a6f1R10).

templates/field/field--oe-contact--oe-press-contact-url--full.html.twig
```diff
@@ -7,15 +7,15 @@
...
-</div>
+  <div class="ecl-u-border-bottom ecl-u-border-color-grey-15 ecl-u-pt-l ecl-u-pb-l">
+    {% include '@ecl-twig/link' with {
+      link: {
+        type: 'standalone',
+        label: 'Press contacts'|t,
+        path: item.content['#url'],
+        icon_position: 'after'
+      },
+      icon: get_link_icon(item.content['#url'].toString())
```
ademarco, line 19: By accessing link render array portions we don't allow cache to bubble up. The easiest way to fix it is to render the link and assign the output to a catch all variable, that will never be displayed, like that:

```twig
   {% set bubble_cache = item.content|render %}
```

Check how it was done in [this PR](https://github.com/openeuropa/oe_theme/pull/787/files#diff-e8ef914d08dc95ae7d28b49062142ef7e0181ea1dff787b83d2a5cbd5cf7a6f1R10).

ademarco (changes requested):

templates/field/field--oe-contact--oe-press-contact-url--full.html.twig
yenyasinn: Added

templates/field/field--oe-contact--oe-link--full.html.twig
yenyasinn: Added

ademarco (approved):

upchuk (approved):

PR #804, opened 2021-05-12
EWPP-1087: Make tables created via WYSIWYG behave correctly when disp…

yenyasinn:


modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -0,0 +1,188 @@
...
+
+    // Ensure that we have tables in the code.
+    if (stristr($text, '<table') !== FALSE) {
+      $dom = Html::load($text);
+      $body = $dom->getElementsByTagName('body')->item(0);
+      foreach ($dom->getElementsByTagName('table') as $table) {
+        $new_table = $this->getEclTable($table);
+        if ($new_table) {
+          $updated_node = $dom->importNode($new_table, TRUE);
+          $body->replaceChild($updated_node, $table);
```
brummbar: I wrapped a table in a div and this line fails, as the child it's not attached to the dom body.

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -0,0 +1,188 @@
...
+class FilterEclTable extends FilterBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function process($text, $langcode) {
+    $result = new FilterProcessResult($text);
+
+    // Ensure that we have tables in the code.
+    if (stristr($text, '<table') !== FALSE) {
```
brummbar: Instead of having a long if(), we can have an early return when the condition is FALSE.

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -0,0 +1,188 @@
...
+use Drupal\filter\Plugin\FilterBase;
+
+/**
+ * Provides a filter to support ECL tables.
+ *
+ * @Filter(
+ *   id = "filter_ecl_table",
+ *   title = @Translation("ECL table support"),
+ *   description = @Translation("Add classes and attributes to the table to align it on mobile devices."),
+ *   type = Drupal\filter\Plugin\FilterInterface::TYPE_TRANSFORM_REVERSIBLE,
```
brummbar: It's not reversable, because if the markup would contain already some classes instances, it wouldn't be possible to know the original status and to revert to the original markup.

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @return \DOMElement|null
+   *   Table with ECL support or NULL.
+   */
+  protected function getEclTable(\DOMElement $table): ?\DOMElement {
+    // Create a clone to manipulate with.
+    $table_node = $table->cloneNode(TRUE);
+
+    // Skip if "tbody" tag doesn't exist.
+    $tbody_node = $table_node->getElementsByTagName('tbody')->item(0);
```
brummbar: Tables can be imported during a migration, and might not have the tbody tag.

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @param \DOMElement $table
+   *   Table as a DOMElement object.
+   *
+   * @return array
+   *   List of columns titles.
+   */
+  protected function getTableHeader(\DOMElement $table): array {
+    $header = [];
+    $thead = $table->getElementsByTagName('thead')->item(0);
```
brummbar: Same as above, thead is optional and might not be present.

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * Checks whether cell is merged or not.
+   *
+   * @param \DOMElement $node
+   *   Table cell.
+   *
+   * @return bool
+   *   TRUE if cell is merged, FALSE otherwise.
+   */
+  protected function isCellMerged(\DOMElement $node): bool {
```
brummbar: We can check easily for row/colspans before doing any processing of a table by using xpath.
```
$xpath = new \DOMXPath($dom);
$span_cells = $xpath->query('.//*[self::th or self::td][(@colspan and @colspan > 1) or (@rowspan and @rowspan > 1)]', $table);
if ($span_cells->count() !== 0) {
  continue;
}
```

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -0,0 +1,188 @@
...
+    $merged_cells = FALSE;
+    $this->processTableCells($tbody_node, $header, $merged_cells, $th_exist);
+
+    $tfoot_node = $table_node->getElementsByTagName('tfoot')->item(0);
+    if (!empty($tfoot_node)) {
+      // Process tfoot.
+      $this->processTableCells($tfoot_node, $header, $merged_cells, $th_exist);
+    }
+
+    if (!$merged_cells && $th_exist) {
```
brummbar: As written below , we could use Xpath once again to check that tables have a th. If none is found, we can skip the table early.

modules/oe_theme_helper/tests/src/Kernel/Plugin/Filter/FilterEclTableTest.php
```diff
@@ -0,0 +1,101 @@
...
+  protected $filter;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+    $this->installConfig(['system']);
+
+    // Get FilterEclTable object.
```
brummbar:
```suggestion
    // Create an instance of the filter.
```

modules/oe_theme_helper/tests/src/Kernel/Plugin/Filter/FilterEclTableTest.php
```diff
@@ -0,0 +1,101 @@
...
+   * Tests the "ECL table support" filter.
+   *
+   * @covers ::process
+   */
+  public function testEclTableFilter(): void {
+    $tests = $this->getTestData();
+
+    foreach ($tests as $source => $value) {
+      $result = $this->filter->process($source, $this->filter)->getProcessedText();
+      $this->assertStringContainsString($value, $result);
```
brummbar: Why are we using this method instead of a simple `assertContains()`?

modules/oe_theme_helper/tests/src/Kernel/Plugin/Filter/FilterEclTableTest.php
```diff
@@ -0,0 +1,101 @@
...
+use Drupal\filter\FilterPluginCollection;
+
+/**
+ * Test FilterEclTable plugin.
+ *
+ * @coversDefaultClass \Drupal\oe_theme_helper\Plugin\Filter\FilterEclTable
+ *
+ * @group batch2
+ */
+class FilterEclTableTest extends KernelTestBase {
```
brummbar: We don't need to use a kernel test. We are testing only the filter, so we can use a unit test like it was done for `\Drupal\Tests\oe_media_iframe\Unit\FilterIframeTagTest` (in oe_media).

modules/oe_theme_helper/tests/src/Kernel/Plugin/Filter/FilterEclTableTest.php
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Gets raw data to filter and expected results.
+   *
+   * @return array
+   *   An associative array, whereas each key is an arbitrary input string and
+   *   each value is an expected result.
+   */
+  protected function getTestData(): array {
```
brummbar: Since we will convert this test to unit, we can use a real data provider. We will have the comments as keys to each test case.

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -0,0 +1,188 @@
...
+      foreach ($tr_node->childNodes as $node) {
+        if (!$node instanceof \DOMElement) {
+          continue;
+        }
+        // Set class to "th" and "td" tags in to support vertical headers.
+        $existing_class = $node->getAttribute('class');
+        $new_class = $existing_class ? "$existing_class ecl-table__cell" : 'ecl-table__cell';
+        $node->setAttribute('class', $new_class);
+
+        if ($node->tagName === 'td') {
```
brummbar: Why we don't add the header name for th elements?

brummbar (changes requested):
Pushed a commit with a working example on how to use xpath to do all the processing.

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
brummbar: Actually thead is required in order to make the responsive styling work correctly.

modules/oe_theme_helper/tests/src/Unit/FilterEclTableTest.php
```diff
@@ -0,0 +1,89 @@
...
+      ],
+      'Table with multi header - leave as-is' => [
+        '<table><thead><tr><th>Column 1</th><th colspan="2">Columns 2 - 3</th><th>Column 4</th></tr><tr><th></th><th>Column 2</th><th>Column 3</th><th></th></tr></thead><tbody><tr><td>1-1</td><td>1-2</td><td>1-3</td><td>1-4</td></tr><tr><td>2-1</td><td>2-2</td><td>2-3</td><td>2-4</td></tr></tbody></table>',
+        '<table><thead><tr><th>Column 1</th><th colspan="2">Columns 2 - 3</th><th>Column 4</th></tr><tr><th></th><th>Column 2</th><th>Column 3</th><th></th></tr></thead><tbody><tr><td>1-1</td><td>1-2</td><td>1-3</td><td>1-4</td></tr><tr><td>2-1</td><td>2-2</td><td>2-3</td><td>2-4</td></tr></tbody></table>',
+      ],
+      'Broken table - leave as-is' => [
+        '<table<thead><tr><th>Column 1</th><th>Column 2</th></tr></thead><tbody><tr><td>1-1</td><td>1-2</td></tr><tr><td>2-1</td><td>2-2</td></tr></tbody></table>',
+        '<table><tr><th>Column 1</th><th>Column 2</th></tr><tbody><tr><td>1-1</td><td>1-2</td></tr><tr><td>2-1</td><td>2-2</td></tr></tbody></table>',
+      ],
```
brummbar: Unfortunately, when treating HTML, DOMDocument doesn't care about [malformed HTML at all](https://www.php.net/manual/en/domdocument.loadhtml.php).
So this test scenario is misleading, and it can be seen as the actual html changes from the original.
We could either track parsing errors and do not process, or serialise the HTML only if changes happened. But we will do like most of the core filters do, and we re-serialise back even if no processing has occurred. Yolo.

modules/oe_theme_helper/tests/src/Unit/FilterEclTableTest.php
```diff
@@ -0,0 +1,89 @@
...
+   *
+   * @dataProvider processDataProvider
+   * @covers ::process
+   */
+  public function testProcess(string $html, string $expected): void {
+    $filter = new FilterEclTable([], 'filter_ecl_table', ['provider' => 'test']);
+    $filter->setStringTranslation($this->getStringTranslationStub());
+
+    $processed_text = $filter->process($html, NULL)->getProcessedText();
+    $this->assertContains($expected, $processed_text);
```
brummbar: We should actually test for exact text, so we can verify fully the work done on the HTML.

brummbar (approved):
I pushed some more tests.

nagyad (approved):

PR #809, opened 2021-05-18
EWPP-1112: Style the iframe paragraph.

yenyasinn:


templates/paragraphs/paragraph--oe-iframe-media.html.twig
```diff
@@ -0,0 +1,32 @@
+{#
+/**
+ * @file
+ * Theme override to display the 'Iframe' paragraph.
+ *
+ * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
+ */
+#}
+
+{# Allow cache from fields to bubble up since we use only values from them. #}
```
yenyasinn: It is approach from https://github.com/openeuropa/oe_theme/pull/789/files#diff-96c12e6bb038adadfcef1cc7c2c635be8d18e956e391efe3797981ce3952f32cR10 and https://github.com/openeuropa/oe_theme/pull/787/files#diff-e8ef914d08dc95ae7d28b49062142ef7e0181ea1dff787b83d2a5cbd5cf7a6f1R10

templates/paragraphs/paragraph--oe-iframe-media.html.twig
```diff
@@ -0,0 +1,32 @@
...
+{% set bubble_cache = content.field_oe_iframe_media_full_width|render %}
+{% set bubble_cache = content.field_oe_media|render %}
```
sergepavle, line 12: Seems like we don't need so much this variable.

sergepavle (dismissed):

oe_theme.theme
```diff
@@ -977,6 +977,38 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
...
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_paragraph__oe_iframe_media(&$variables) {
+  /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
+  $paragraph = $variables['paragraph'];
+
+  // Get the full width value.
+  if (!$paragraph->get('field_oe_iframe_media_full_width')->isEmpty()) {
+    $variables['full_width'] = (bool) $paragraph->get('field_oe_iframe_media_full_width')->value;
```
upchuk: Mmm, if this is a boolean field, shouldn't this just be `$variables['full_width'] = (bool) $paragraph->get('field_oe_iframe_media_full_width')->value;`? If not filled in, it means it's false.

templates/paragraphs/paragraph--oe-iframe-media.html.twig
upchuk: Neither of these two renders are needed:

* `content.field_oe_iframe_media_full_width|render` -> this is a bool field on the paragraph that you are templating here, so if its value changes, it gets invalidated based on the paragraph. You are also not using anything from its render array but just use the value in the preprocessor.
* `field_oe_media` -> is indeed not on the paragraph, so since you use a value from the media, you just need to mark that media a cache dependency on the preprocessor. So when it changes, it invalidates this as well. Moreover, you need to check access to the media (which you didnt) and ensure also the access caches are added.

For neither of these cases you are not extracting something from a render array to justify extra renders here.

upchuk (dismissed):

oe_theme.theme
```diff
@@ -977,6 +977,52 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
...
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_paragraph__oe_iframe_media(&$variables) {
+  /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
+  $paragraph = $variables['paragraph'];
+
+  // Get the full width value.
+  $variables['full_width'] = (bool) $paragraph->get('field_oe_iframe_media_full_width')->value;
```
22Alexandra: We should get the value only after we know we have a media entity.

oe_theme.theme
```diff
@@ -977,6 +977,52 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
...
+    return;
+  }
+
+  // Caches are handled by the formatter usually. Since we are not rendering
+  // the original render arrays, we need to propagate our caches to the
+  // paragraph template.
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
+  $cacheability->addCacheableDependency($media);
+  $media_type = \Drupal::entityTypeManager()->getStorage('media_type')->load($media->bundle());
+  $cacheability->addCacheableDependency($media_type);
```
22Alexandra, line 1009: Get the media type after ensuring the user has access to view the media entity.

tests/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -688,4 +697,97 @@ public function testBanner(): void {
...
+    // Assert ratio.
+    $media->set('oe_media_iframe_ratio', '1_1')->save();
+    $html = $this->renderParagraph($paragraph);
+    $crawler = new Crawler($html);
+    $iframe = $crawler->filter('figure.ecl-media-container.ecl-media-container--fullwidth div.ecl-media-container__media.ecl-media-container__media--ratio-1-1 iframe');
+    $this->assertContains('http://example.com/iframe', $iframe->attr('src'));
+
+    // Assert title and full width.
+    $paragraph->set('field_oe_title', 'Iframe paragraph title');
+    $paragraph->set('field_oe_iframe_media_full_width', FALSE)->save();
```
22Alexandra, line 808: This should be the default behaviour. Not sure why testing it after testing the rendering for full width being enabled.

oe_theme.theme
yenyasinn: I don't think so. I add cache metadata of the media image even if user doesn't have access to that image because we don't know exactly why user doesn't have access. Might be he doesn't have access to the all media images.

tests/Kernel/Paragraphs/MediaParagraphsTest.php
yenyasinn: By default it is FALSE and here I just return paragraph to the default state.

oe_theme.theme
yenyasinn: Yes. Fixed

22Alexandra (dismissed):

sergepavle (dismissed):

nagyad (dismissed):

sergepavle (dismissed):

ademarco (dismissed):

imanoleguskiza (approved):

upchuk (approved):

PR #820, opened 2021-05-28
EWPP-1143: Apply access checks for media render.

nagyad:


modules/oe_theme_content_news/oe_theme_content_news.module
```diff
@@ -24,12 +25,24 @@ function oe_theme_content_news_preprocess_node__oe_news(&$variables) {
 
   $media = $node->get('oe_news_featured_media')->entity;
   if ($media instanceof MediaInterface) {
+    $cacheability = CacheableMetadata::createFromRenderArray($variables);
+    $cacheability->addCacheableDependency($media);
+
+    // Run access checks on the media entity.
+    $access = $media->access('view', $variables['user'], TRUE);
+    $cacheability->addCacheableDependency($access);
```
yenyasinn, line 33: I would propose:
```suggestion
    $cacheability->addCacheableDependency($access);
    $cacheability->applyTo($variables);
```

so won't need to use `$cacheability->applyTo($variables);` twice below

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
```diff
@@ -152,6 +152,8 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
    *
    * @return array
    *   The ECL media-container parameters.
+   *
+   * @SuppressWarnings(PHPMD.CyclomaticComplexity)
```
yenyasinn, line 156: Why do we need to test our code on the cyclomatic complexity if when it happens we use such workaround? Might be code should be refactored?

tests/Functional/ContentEventRenderTest.php
```diff
@@ -125,15 +125,22 @@ public function testEventFeaturedMediaTranslation(): void {
...
     ];
 
     foreach ($node->getTranslationLanguages() as $node_langcode => $node_language) {
       $node = \Drupal::service('entity.repository')->getTranslationFromContext($node, $node_langcode);
       $this->drupalGet($node->toUrl());
       $this->assertSession()->elementExists('css', 'figure[class="ecl-media-container"] img[src*="' . $file_urls[$node_langcode] . '"][alt="default ' . $node_langcode . ' alt"]');
     }
+
+    // Unpublish the media and assert it is not rendered anymore.
+    $media->set('status', 0);
```
yenyasinn, line 139: It is correct, but also core/lib/Drupal/Core/Entity/EntityPublishedTrait.php provides setPublished() method. What should we prefer?

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
```diff
@@ -152,6 +152,8 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
    *
```
yenyasinn, line 152: We had a ticket "Contact body should take full width when no image is present" https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EWPP-716. 

Currently wrapper is shown around absent image in the contact entity when media image is unpublished. Please check condition in templates/oe_contact/oe-contact--full.html.twig.

Screenshot 2021-05-31 at 12 11 33

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationDescription.php
```diff
@@ -92,6 +92,15 @@ public function viewElements(ContentEntityInterface $entity) {
 
     $cacheability = CacheableMetadata::createFromRenderArray($build);
     $cacheability->addCacheableDependency($media);
+
+    // Run access checks on the media entity.
+    $access = $media->access('view', NULL, TRUE);
+    $cacheability->addCacheableDependency($access);
+    if (!$access->isAllowed()) {
+      $cacheability->applyTo($build);
+      return [];
```
yenyasinn:
```suggestion
     return $build;
```
Otherwise body field value is absent when media is unpublished.

tests/Functional/ContentEventRenderTest.php
nagyad: In tests, I prefer to be explicit and use the status field instead of published flag.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
```diff
@@ -165,12 +167,16 @@ protected function viewElement(FieldItemInterface $item, string $langcode): arra
...
-    // the original render arrays, we need to propagate our caches to the
-    // oe_theme_helper_featured_media template.
     $cacheability->addCacheableDependency($media);
 
+    // Run access checks on the media entity.
+    $access = $media->access('view', NULL, TRUE);
+    $cacheability->addCacheableDependency($access);
+    if (!$access->isAllowed()) {
+      $cacheability->applyTo($build);
+      return [];
```
yenyasinn:
```suggestion
      return $build;
```
Otherwise cache metadata isn't added.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/FeaturedMediaFormatter.php
nagyad: Not in this case.

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/MediaThumbnailUrlFormatter.php
```diff
@@ -76,10 +76,14 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
       $cache->addCacheableDependency($media);
       $cache->addCacheableDependency($items->get($delta)->_accessCacheability);
 
+      // Run access checks on the media entity.
+      $access = $media->access('view', NULL, TRUE);
+      $cache->addCacheableDependency($access);
```
upchuk, line 80: Doens't this cacheability get added at line 77? I'm not 100%, but check.
```
$cache->addCacheableDependency($items->get($delta)->_accessCacheability);
```

upchuk (changes requested):

upchuk (changes requested):
What about:

* oe_theme_content_entity_organisation_preprocess_oe_organisation

tests/Functional/ContentNewsRenderTest.php
```diff
@@ -140,6 +140,15 @@ public function testNewsRendering(): void {
     $this->assertContentHeader($contacts_content, 'Contacts');
     $this->assertContactDefaultRender($contacts_content, 'news_contact');
 
+    // Add a different and unpublished media and assert it is not rendered
+    // in the contact.
+    $contact_media = $this->createMediaImage('news_contact_media');
```
upchuk: Well, this would not assert the caching. You should retrieve the media that was created for that default contact, unpublish it, and then assert it is no longer rendered.

tests/Functional/ContentProjectRenderTest.php
```diff
@@ -263,6 +263,15 @@ public function testProjectRendering(): void {
     $this->assertEquals($stakeholder_sub_headers[0]->getText(), 'Coordinators');
     $this->assertStakeholderOrganisationRendering($project_stakeholders, 'coordinator');
 
+    // Create image for a new logo that is unpublished and assert that is not
+    // rendered.
+    $media = $this->createMediaImage('coordinator_logo');
```
upchuk: Here as well, you are creating a new one. An existing one should be asserted that it shows, then unpublish, then assert it no longer shows.

upchuk (changes requested):

upchuk (dismissed):

upchuk (approved):

yenyasinn (approved):

PR #829, opened 2021-06-08
EWPP-1149: Port facts and figures.

22Alexandra:


templates/patterns/facts_figures/facts_figures.ui_patterns.yml
```diff
@@ -7,19 +7,19 @@ facts_figures:
       label: "Facts"
       description: "List of facts"
       preview:
-        - icon: "instagram"
+        - icon: "infographic"
```
imanoleguskiza: This is an issue, because currently we do support this type of icons on the paragraph itself. We need to either drop support for them in the paragraph or somehow find a way to support them.

imanoleguskiza (changes requested):

sergepavle (approved):

imanoleguskiza (approved):

PR #831, opened 2021-06-08
EWPP-1153: Fix breadcrumb page title cache.

nagyad:


oe_theme.theme
```diff
@@ -81,10 +81,17 @@ function oe_theme_preprocess(&$variables) {
...
-  $request     = \Drupal::request();
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
+  $cacheability->addCacheContexts(['route']);
+
+  $request = \Drupal::request();
   $route_match = \Drupal::routeMatch();
-  $title       = \Drupal::service('title_resolver')
-    ->getTitle($request, $route_match->getRouteObject());
+
+  $node = $route_match->getRouteObject();
```
brummbar: The route object is not a node, it's an instance of `\Symfony\Component\Routing\Route`.

oe_theme.theme
```diff
@@ -81,10 +81,17 @@ function oe_theme_preprocess(&$variables) {
...
+  $title = \Drupal::service('title_resolver')->getTitle($request, $node);
+
+  $cacheability->addCacheableDependency($node);
```
brummbar: The Route class doesn't implement the CacheableDependencyInterface, so this line sets the whole block uncacheable (see `\Drupal\Core\Cache\RefinableCacheableDependencyTrait::addCacheableDependency()`).
The only change to do in this function was to add 
```
CacheableMetadata::createFromRenderArray($variables)
    ->addCacheContexts(['route'])
    ->applyTo($variables);
```

tests/Functional/BreadcrumbRenderTest.php
```diff
@@ -0,0 +1,113 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_theme\Functional;
+
+/**
+ * Tests that breadcrumbs are rendered correctly.
+ *
+ * @group batch1
+ */
+class BreadcrumbRenderTest extends ContentRenderTestBase {
```
brummbar: We are not testing the rendering here, but the breadcrumb cache, so its behaviour. Let's name this simply BreadcrumbTest.

tests/Functional/BreadcrumbRenderTest.php
```diff
@@ -0,0 +1,113 @@
...
+    'oe_theme_helper',
+    'path',
+    'oe_theme_content_news',
+    'block',
+  ];
+
+  /**
+   * Tests that the breadcrumbs are rendered correctly for news content type.
+   */
+  public function testNewsBreadcrumbRendering(): void {
```
brummbar: The scope of this test is to test the caching, not the rendering, and caching in general, not of the news content type. If I see this method, I will never imagine that we are trying to cover the caching of the last breadcrumb item.
Let's name it `testBreadcrumbCaching`.

tests/Functional/BreadcrumbRenderTest.php
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param string $last_segment_title
+   *   The expected last segment.
+   */
+  protected function assertBreadcrumbs(string $last_segment_title): void {
+    $breadcrumbs = [
+      // Test the page header breadcrumb.
+      '.ecl-page-header-core .ecl-breadcrumb-core',
+      // Test the system breadcrumb.
+      '[class="ecl-breadcrumb-core"]',
```
brummbar: Why this and not a simple `.ecl-breadcrumb-core`?

tests/Functional/BreadcrumbRenderTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+    $this->drupalGet($node_1->toUrl());
+    $this->assertBreadcrumbs('Test news node');
+
+    $this->drupalGet($node_2->toUrl());
+    $this->assertBreadcrumbs('Test news article breadcrumb');
+  }
+
+  /**
+   * Helper to assert both system and page header breadcrumbs on the page.
```
brummbar: The thing is that the cache we are adding in this ticket only impacts the system breadcrumb rendering. The page header breadcrumb is tested in tests/features/page-header.feature:25.

tests/Functional/BreadcrumbRenderTest.php
```diff
@@ -0,0 +1,113 @@
...
+      // Check for the number of links.
+      $links_count = $page_breadrumb->findAll('css', 'ol.ecl-breadcrumb-core__container li.ecl-breadcrumb-core__segment a.ecl-breadcrumb-core__link');
+      $this->assertCount(2, $links_count);
+
+      // Check the last segment.
+      $current_page = $page_breadrumb->findAll('css', 'ol.ecl-breadcrumb-core__container li.ecl-breadcrumb-core__current-page');
+      $this->assertCount(1, $current_page);
+      $current_page = reset($current_page);
+      $this->assertEquals($last_segment_title, trim($current_page->getText()));
+    }
```
brummbar: The rendering of the breadcrumb is tested in `\Drupal\Tests\oe_theme\Kernel\BreadcrumbTest`. Here we just need to assert the titles of the elements, so we can verify the caching.

brummbar (changes requested):

tests/Functional/BreadcrumbRenderTest.php
nagyad: This is an exclusive selector for the element that has only ecl-breadcrumb-core class, like the system breadcrumb block. Otherwise it will select always the page header block one that has .ecl-page-header-core too.

tests/Functional/BreadcrumbTest.php
```diff
@@ -0,0 +1,116 @@
...
+ * Tests that breadcrumbs are cached correctly.
+ *
+ * @group batch1
+ */
+class BreadcrumbTest extends ContentRenderTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = [
```
brummbar: I missed this, it should be protected.

tests/Functional/BreadcrumbTest.php
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+   * Helper to assert system breadcrumbs on the page.
+   *
+   * @param array $expected
+   *   The expected breadcrumb titles in the expected order.
+   */
+  protected function assertSystemBreadcrumbs(array $expected): void {
+    // Get the last segment title nad the link titles.
```
brummbar:
```suggestion
    // Get the last segment title and the link titles.
```

tests/Functional/BreadcrumbTest.php
```diff
@@ -0,0 +1,116 @@
...
+    $last_segment_title = array_pop($expected);
+    $page_breadcrumb = $this->assertSession()->elementExists('css', '[class="ecl-breadcrumb-core"]');
+
+    // Assert the link titles.
+    $links = $page_breadcrumb->findAll('css', 'a');
+    $this->assertCount(count($expected), $links);
```
brummbar: Nit: use assertSameSize().

brummbar (changes requested):
Some tiny nits, rest is good.

brummbar (approved):

upchuk (approved):

PR #839, opened 2021-06-10
[EWPP-873] EWPP-876: Theme the link and new variants on text with featued media paragraphs.

yenyasinn:


oe_theme.theme
```diff
@@ -915,6 +915,21 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
     return;
   }
 
+  // Set pattern variant.
+  $variables['variant'] = 'left_simple';
```
nagyad: I would rather move this default setting to the twig and there put in the header that it defaults to this because of backwards compatibility.

templates/patterns/text_featured_media/pattern-text-featured-media.html.twig
```diff
@@ -12,13 +14,39 @@
 {% endif %}
 {% set text = text|render %}
 
-{% if image or video or text %}
+{% set media_position_class = ' ecl-u-order-md-last' %}
+{% if 'right' in variant %}
+  {% set media_position_class = '' %}
+{% endif %}
+
+{% set cta = '' %}
```
nagyad: What is cta? Can we have more self-explanatory variable names?

templates/patterns/text_featured_media/pattern-text-featured-media.html.twig
```diff
@@ -12,13 +14,39 @@
...
+{% if link %}
+  {% set link_type = 'standalone' %}
+  {% if 'featured' in variant %}
+    {% set link_type = 'cta' %}
```
nagyad: Let's make this type call_to_action.

tests/PatternAssertions/TextFeaturedMediaAssert.php
```diff
@@ -74,4 +77,57 @@ protected function assertVideoRatio(string $expected_ratio, string $selector, Cr
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getPatternVariant(string $html): string {
+    // Variant can be clearly recognised if image or video and link exist only.
+    // In case if the link isn't provided then "left_simple", "right_simple"
+    // will be returned based on the position of media.
+    // In case if the image or videa arn't provided then
```
nagyad: Fix typos.

templates/patterns/text_featured_media/pattern-text-featured-media.html.twig
```diff
@@ -12,13 +14,39 @@
...
+  {% endif %}
+
+  {% set cta %}
```
nagyad: This should be call_to_action_link.

nagyad (changes requested):

templates/patterns/text_featured_media/pattern-text-featured-media.html.twig
yenyasinn: `cta` is a type of link that is defined in components/ec-component-link/ecl-link.html.twig. We can't change it.

oe_theme.theme
```diff
@@ -901,6 +901,18 @@ function oe_theme_preprocess_paragraph__oe_text_feature_media(array &$variables)
...
+
+  // Add link field.
+  if (!$paragraph->get('field_oe_link')->isEmpty()) {
+    $link_field = $paragraph->get('field_oe_link')->first();
+    $variables['link'] = [
+      'path' => $link_field->getUrl()->toString(),
+      'label' => $link_field->get('title')->getValue(),
+    ];
+  }
+
```
22Alexandra: Are we supposed to display any paragraph content if there is no media available? 
According to the requirements [page](https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?pageId=553354339), the media should be always present in the frontend.
no media

I think this code should be moved after ensuring there is a media entity and the user has access to it.

tests/Kernel/fixtures/rendering.yml
```diff
@@ -2342,14 +2365,17 @@
     count:
       'div.ecl-row': 0
       'figure.ecl-media-container': 0
+      'div.ecl-col-md-6.ecl-u-mb-m.ecl-u-mb-md-none': 0
```
22Alexandra, line 2372: When no variant is passed, we should make sure the left_simple is used so we could just add the link.

22Alexandra (changes requested):
The tests need to be fixed.

oe_theme.theme
yenyasinn: I have not found mention that `the media should be always present in the frontend` in spec. Also "Featured Media" is optional field. There is a special logic that shows text in full width when Media is absent https://github.com/openeuropa/oe_theme/pull/839/files#diff-1fb33e85cf11b49744aeae0c0e8b9daaa3ad245ec4ca2fb9bbd735a02b1d14f7R68.

So, from my point of view rendering of Link field isn't dependent on the Featured media.

tests/Kernel/fixtures/rendering.yml
yenyasinn: This particular test ensures that content is empty. 
I've added case with empty variant below.

22Alexandra (dismissed):

templates/patterns/text_featured_media/pattern-text-featured-media.html.twig
```diff
@@ -2,23 +2,53 @@
...
+
 {% set video_ratio = video_ratio|replace({':': '-'}) %}
 {% set allowed_values = ['16-9','4-3','3-2','1-1'] %}
 {% if video_ratio not in allowed_values %}
   {% set video_ratio = '16-9' %}
 {% endif %}
 {% set text = text|render %}
 
-{% if image or video or text %}
+{% set media_position_class = ' ecl-u-order-md-last' %}
```
imanoleguskiza: We could use terniary operators here:
{% set media_position_class = 'right' in variant ? '' : ' ecl-u-order-md-last' %}

templates/patterns/text_featured_media/pattern-text-featured-media.html.twig
```diff
@@ -2,23 +2,53 @@
...
+{% if 'right' in variant %}
+  {% set media_position_class = '' %}
+{% endif %}
+
+{% set call_to_action_link = '' %}
+{% if link %}
+  {% set link_type = 'standalone' %}
```
imanoleguskiza: Same here, we could do {% set link_type = 'featured' in variant ? 'cta' : 'standalone' %}

tests/PatternAssertions/TextFeaturedMediaAssert.php
```diff
@@ -74,4 +77,57 @@ protected function assertVideoRatio(string $expected_ratio, string $selector, Cr
...
+    // In case if the link isn't provided then "left_simple", "right_simple"
+    // will be returned based on the position of media.
+    // In case if the image or video aren't provided then
+    // "left_simple", "left_featured" will be returned based on the link style.
+    // If image or video and link aren't provided then "left_simple" will be
+    // returned as default value.
+    $crawler = new Crawler($html);
+
+    $position_variant = 'left';
+    $media_position = $crawler->filter('.ecl-u-order-md-last');
```
imanoleguskiza, line 116: If there is no media this class is not present and the method will return a "right" type variant, but if there is no media we should return a "left" type variant right?

tests/PatternAssertions/TextFeaturedMediaAssert.php
```diff
@@ -74,4 +77,57 @@ protected function assertVideoRatio(string $expected_ratio, string $selector, Cr
...
+
+    $svg = $link_element->filter('svg.ecl-icon.ecl-icon--s.ecl-icon--primary.ecl-link__icon use');
+    self::assertContains('icons.svg#ui--' . $expected_link['icon'], $svg->attr('xlink:href'));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getPatternVariant(string $html): string {
+    // Variant can be clearly recognised if image or video and link exist only.
```
imanoleguskiza: The variant can be easily extracted by checking the presence and properties of the media and link.
If the link is not present then we return left_simple or right_simple based on the media position.
If the media is not present then left simple or left featured will be returned based on the link style.
If neither media nor image are present  then left_simple will be returned.

templates/patterns/text_featured_media/text_featured_media.ui_patterns.yml
```diff
@@ -1,6 +1,19 @@
 text_featured_media:
   label: "Text with Featured media"
   description: "Shows a rich text with a title and a featured media on the right, if any."
+  variants:
+    left_simple:
+      label: 'Left simple'
+      description: 'Text on the left, simple call to action.'
```
imanoleguskiza: Let's mention this is the default behavior.

imanoleguskiza (changes requested):

tests/PatternAssertions/TextFeaturedMediaAssert.php
yenyasinn: You are right. I've added check that media wrapper exists to ensure that there is media.

imanoleguskiza (dismissed):

imanoleguskiza (dismissed):

nagyad (dismissed):

nagyad (approved):

imanoleguskiza (approved):

PR #847, opened 2021-06-14
EWPP-1148: Adapt site header, page header and breacrumbs to ecl V3

imanoleguskiza:
### Description

Adapt site header, page header and breacrumbs to ecl V3
### Change log

- Added:
- Changed: Adapt site header, page header and breacrumbs to ecl V3
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


templates/patterns/page_header/pattern-page-header.html.twig
```diff
@@ -52,12 +46,12 @@
 
 {% include '@ecl-twig/page-header-' ~ ecl_branding ~ '/page-header-' ~ ecl_branding with {
   'title': title,
+  'thumbnail': thumbnail,
   'description': introduction,
   'breadcrumb': _breadcrumb,
-  'meta': _metas,
-  'infos': _infos,
-  'background_image': ecl_branding == 'core' and background_image_url is not empty,
+  'meta': metas,
```
sergepavle: I think we have to keep using 'filter_empty' filter.
```suggestion
  'meta': metas|filter_empty,
```

templates/patterns/page_header/pattern-page-header.html.twig
```diff
@@ -52,12 +46,12 @@
...
   'background_image_url': background_image_url,
   'composition': _language_switcher,
   'extra_classes': _extra_classes|join(' '),
+  'overlay': overlay,
```
sergepavle: We have to provide this parameter only for EC ecl branding.

templates/compositions/ec-component-language-list/language-list-selector.html.twig
```diff
@@ -66,7 +66,7 @@
 {% endif %}
 
 <a
-  class="ecl-link ecl-link--standalone ecl-site-header-{{ _ecl_branding }}__language-selector"
+  class="ecl-button ecl-button--ghost  ecl-site-header-{{ _ecl_branding }}__language-selector"
```
sergepavle, line 69: in templates/blocks/block--oe-search.html.twig classes seems like also should be updated.

sergepavle (changes requested):

templates/compositions/ec-component-language-list/language-list-selector.html.twig
imanoleguskiza: I took a look and I didn't find anything to be changed. Could you be more precise?

sergepavle (dismissed):

ademarco (changes requested):
Page header on EU / Core looks like below: we need to make sure both text and icon are visible and that the dropdown works correctly:

image

On EC / Core we can have trouble seeing the text, if a picture is below it:

image

imanoleguskiza:
I pushed a fix for the colors in ec/eu flavors and to make sure nothing is invisible, but I am not sure how you propose we fix the issue with some of the text not being visibile when there is an image in the background, I feel like this is an issue editors need to take into account when uploading images.

ademarco:
I've just got new reqs that we will now have a dedicated composition for the content language switcher, so we can discard fixing it in this PR. I'll open a follow up for that only.

ademarco (approved):

22Alexandra (approved):

PR #851, opened 2021-06-16
[EWPP-1194] EWPP-1196: Theme Contact paragraph.

22Alexandra:


modules/oe_theme_paragraphs_contact/oe_theme_paragraphs_contact.info.yml
```diff
@@ -0,0 +1,7 @@
+name: OpenEuropa Theme Paragraphs Contact
+type: module
+description: Companion module for the OE Paragraphs Contact module
+package: OpenEuropa
+core: 8.x
+dependencies:
+  - oe_paragraphs:oe_paragraphs_contact
```
yenyasinn, line 7:
```suggestion
  - oe_paragraphs:oe_paragraphs_contact
  - oe_theme:oe_theme_content_entity_contact
```
If we want to use Full view mode of Contact entity.

modules/oe_theme_paragraphs_contact/oe_theme_paragraphs_contact.install
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types = 1);
+
+use Drupal\Core\Entity\Entity\EntityViewDisplay;
+
+/**
+ * Implements hook_install().
+ *
+ * Sets contacts rendering to use the full view mode.
+ */
+function oe_theme_paragraphs_contact_install() {
```
yenyasinn: It is correct but usually we provide `config/install/core.entity_view_display.*.*.full.yml` file in oe_theme. Can we use hook_install or consistent way with config should be used? (I am for consistency)

oe_theme.theme
```diff
@@ -1614,3 +1614,12 @@ function oe_theme_preprocess_patterns_overview_page(&$variables) {
     }
   }
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_paragraph__oe_contact(&$variables) {
+  if (!$variables['paragraph']->get('field_oe_title')->isEmpty()) {
```
yenyasinn: We can use `$variables['title'] = $variables['paragraph']->get('field_oe_title')->value;`without condition - if it is empty it returns empty.

templates/paragraphs/paragraph--oe-contact.html.twig
```diff
@@ -0,0 +1,14 @@
+{#
+/**
+ * @file
+ * Theme override to display the Contact paragraph type.
+ *
+ * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
+ */
+#}
+{% if title is not empty %}
```
yenyasinn: Please consider to use
```
{% if not paragraph.field_oe_title.isEmpty() %}
  <h2 class="ecl-u-type-heading-2">
    {{ content.field_oe_title }}
  </h2>
{% endif %}
```
to avoid using oe_theme_preprocess_paragraph__oe_contact

modules/oe_theme_paragraphs_contact/oe_theme_paragraphs_contact.install
22Alexandra: We don't create a new view display for the paragraph, just changing the view mode of the contact entity to be rendered in full view mode instead of the default one that comes from the paragraphs module.

yenyasinn (dismissed):

modules/oe_theme_paragraphs_contact/oe_theme_paragraphs_contact.install
```diff
@@ -0,0 +1,23 @@
...
+  $entity_view_display = EntityViewDisplay::load('paragraph.oe_contact.default');
```
upchuk, line 22: This only needs to run if we are not in the middle of a config sync. See other hook_installs for example.

tests/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -737,4 +765,92 @@ public function testBanner(): void {
     );
   }
 
+  /**
+   * Test contact paragraph rendering.
+   *
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   */
+  public function testContact(): void {
```
upchuk: I would make a new test class for this. We are doubling the requirements due this one test which will run for all the other tests in this class.

tests/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -737,4 +765,92 @@ public function testBanner(): void {
...
+    $paragraph = Paragraph::create([
+      'type' => 'oe_contact',
+      'field_oe_contacts' => [$general_contact, $press_contact],
+    ]);
+    $paragraph->save();
+    $html = $this->renderParagraph($paragraph);
+    $crawler = new Crawler($html);
+
+    // Assert paragraph heading markup is not rendered if no title is set.
+    $this->assertCount(0, $crawler->filter('h2.ecl-u-type-heading-2'));
```
upchuk: Isnt there something we can borrow from ContactRenderTest to reuse some of the assertions? This seems quite duplicate since the full view mode is tested elsewhere

upchuk (changes requested):

tests/Kernel/Paragraphs/MediaParagraphsTest.php
22Alexandra: ContactRenderTest tests in detail each part of the contact being rendered in full view mode and would help only for asserting the featured media. In the paragraph test we only test that contact is being rendered using the full view mode.

tests/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -744,15 +744,19 @@ public function testBanner(): void {
    */
   public function testIframe(): void {
     // Set Iframe media translatable.
-    $this->container->get('content_translation.manager')->setEnabled('media', 'iframe', TRUE);
```
upchuk: Why are there changes in this file?

upchuk (changes requested):

yenyasinn (approved):

upchuk (approved):

PR #856, opened 2021-06-23
[EWPP-1142] EWPP-1238: Fix general purpose rendering tests

22Alexandra:


22Alexandra:
Pattern templates updates done based on the migrating information from ecl: https://github.com/ec-europa/europa-component-library/blob/v3-dev/docs/Migrating-v3.md

yenyasinn (approved):

templates/patterns/file/pattern-file.html.twig
```diff
@@ -10,7 +10,6 @@
   'meta': '(' ~ file.size|format_size ~ ' - ' ~ file.extension|upper ~ ')',
   'icon': {
     'path': ecl_icon_path,
-    'type': 'general',
     'name': file.extension|to_file_icon,
```
imanoleguskiza, line 13: There is a problem here, the "presentation" icon used in the "to_file_icon" filter is not longer available.

imanoleguskiza (dismissed):

imanoleguskiza (approved):

PR #864, opened 2021-06-25
EWPP-941: Datepicker styling.

yenyasinn:


sass/components/_list_pages.scss
```diff
@@ -0,0 +1,14 @@
+/**
+ * Expand datepicker on the list page form filter to have the same width as select boxes.
+ */
+.oe-list-pages-facets-form .ecl-datepicker {
```
upchuk, line 4: Why is this our responsibility to fix this? It's an ECL component.

upchuk (changes requested):

yenyasinn:
was fixed in ECL v3

PR #865, opened 2021-06-25
EWPP-943: Allow using different variants of the button component for input--submit template.

22Alexandra:


templates/form/input--submit.html.twig
```diff
@@ -13,6 +13,6 @@
 
 {% include '@ecl-twig/button' with {
   label: attributes.value,
-  variant: 'primary',
+  variant: attributes.variant|default('primary'),
```
upchuk: Mmmm, not really good to use the attributes array for this. That will also get printed onto the element in html.

Just make a separate variable just for that.

upchuk (changes requested):

upchuk (changes requested):
Also, some small test would be needed.

oe_theme.theme
```diff
@@ -321,6 +321,15 @@ function oe_theme_preprocess_input__date(array &$variables): void {
   $variables['placeholder'] = $ecl_format;
 }
 
+/**
+ * Implements template_preprocess_input__submit().
+ */
+function oe_theme_preprocess_input__submit(array &$variables): void {
+  if (isset($variables['element']['#variant'])) {
```
upchuk: Looking at the `Drupal\Core\Render\Element\Button` and it seems the element already supports a #button_type property. So we can use that instead and map it to the ECL "variant"

upchuk (changes requested):

upchuk (dismissed):

upchuk (approved):

imanoleguskiza (approved):

PR #868, opened 2021-07-02
[EWPP-1142] EWPP-1250: Remove visibility condition for the main navigation block.

22Alexandra:


modules/oe_theme_helper/templates/oe-theme-helper-site-navigation.html.twig
```diff
@@ -10,8 +10,10 @@
  * @ingroup themeable
  */
 #}
-{{ pattern('navigation_menu', {
-  label: 'Menu'|t,
-  items: menu_items,
-  site_name: site_name,
-}, 'vertical') }}
+{% if menu_items %}
```
imanoleguskiza: We don't need to check the menu_items twice, the pattern already checks if there is content to be rendered.

templates/patterns/navigation_menu/pattern-navigation-menu.html.twig
```diff
@@ -46,5 +46,11 @@
     items: _items,
     icon_path: ecl_icon_path
   } only %}
+{% elseif site_name and ecl_branding == 'standardised' %}
```
imanoleguskiza: No need for this, its part of the page layout so it should not be part of the navigation menu.

imanoleguskiza (dismissed):

nagyad (approved):

yenyasinn (approved):

PR #878, opened 2021-07-12
[EWPP-873] EWPP-1327: Theme text with featured media title.

yenyasinn:


templates/paragraphs/paragraph--oe-text-feature-media.html.twig
```diff
@@ -15,6 +15,7 @@
 {{ pattern('text_featured_media', {
   'variant': variant,
   'title': not paragraph.field_oe_title.isEmpty ? content.field_oe_title,
+  'text_title': not paragraph.field_oe_feature_media_title.isEmpty ? content.field_oe_feature_media_title,
```
22Alexandra, line 18: This won't work on update path because `core.entity_view_display.paragraph.oe_text_feature_media.default.yml` only has the `field_oe_feature_media_title` enabled on the install path.

22Alexandra (changes requested):
We need to update the view display to include the new title field.

22Alexandra (approved):

nagyad (approved):

PR #903, opened 2021-08-02
[EWPP-1360] EWPP-1363: Style Sources field in full view mode.

22Alexandra:
### Description

Style Sources field in full view mode.

### Change log

- Changed: Sources field in full view mode.

### DIFF
--- a/templates/field/field--node--oe-news-sources--oe-news--full.html.twig
+++ b/templates/field/field--node--oe-news-sources--oe-news--full.html.twig
@@ -1,0 +1,36 @@
+{#
+/**
+ * @file
+ * News sources field template.
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+<h2 class="ecl-u-type-heading-2">{{ 'Sources'|t }}</h2>
+<div class="ecl-list">
+  {% for item in items %}
+    <article class="ecl-content-item ecl-u-d-s-flex ecl-u-pb-m">
+      <div class="ecl-u-flex-grow-1">
+        <div class="ecl-content-item__title ecl-u-type-heading-5 ecl-u-mb-xs ecl-u-mt-none">
+          {%- if item.content['#url'] %}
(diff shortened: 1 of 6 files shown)


templates/field/field--node--oe-news-sources--oe-news--full.html.twig
```diff
@@ -0,0 +1,36 @@
...
+              }
+            } %}
+          {% else %}
+            {{ item.content['#title'] }}
+          {%- endif -%}
+        </div>
+      </div>
+    </article>
+  {% endfor %}
+</div>
```
sergepavle, line 36: Looks like it would be more suitable to reuse the `list_item` pattern like we did for related_links field.
See \oe_theme_preprocess_field function inside oe_theme.theme.

tests/Functional/ContentNewsRenderTest.php
```diff
@@ -85,6 +85,42 @@ public function testNewsRendering(): void {
     $this->assertEquals('http://www.example.org', $header_link->getAttribute('href'));
     $this->assertEquals('http://www.example.org', $header_link->getText());
 
+    // Assert Sources field content.
```
sergepavle, line 88: I suppose by switching to implementation based on the `list_item` pattern, we can use the appropriate `assertPattern` method.

templates/field/field--node--oe-news-sources--oe-news--full.html.twig
sergepavle: Unfortunately, we can't, for now, use the `list_item` pattern according to fact that we can't control the link's icon. I think it would be the more preferable solution anyway to use a pattern but with additional parameter for controlling the link's type (for adding external icon if needed).
I suppose it can be impelmented in follow-up ticket.

sergepavle (dismissed):

nagyad (dismissed):

nagyad (approved):

sergepavle (approved):

PR #913, opened 2021-08-08
[EWPP-1360] EWPP-1388: Add icon optional parameter to list item pattern.

22Alexandra:


templates/patterns/list_item/pattern-list-item.html.twig
```diff
@@ -53,6 +53,11 @@
               type: 'standalone',
               label: title,
               path: url
+            },
+            icon: {
+              'path': ecl_icon_path,
+              'name': icon|default(''),
```
sergepavle, line 59: Looks like missed mentoned `icon` in list of parameters.

sergepavle (changes requested):

templates/patterns/list_item/pattern-list-item.html.twig
sergepavle: Looks like missed mentoned `icon` in list of parameters.

sergepavle (dismissed):

templates/patterns/list_item/pattern-list-item.html.twig
```diff
@@ -53,6 +54,11 @@
               type: 'standalone',
               label: title,
               path: url
+            },
+            icon: {
```
sergepavle: I still have doubts that it is safe. We rely on `ecl-link` ECL component implementation where we have to check that the icon name is not empty. But what if we will have a change in implementation. I don't sure that it is possible, but the chance still exists. It would be simpler to keep the empty icon element if we don't have the `icon` name from our side for explicitly declaring that link without an icon.

sergepavle (commented):
Some more questions for discussion.

oe_theme.theme
```diff
@@ -441,6 +445,9 @@ function oe_theme_preprocess_field(&$variables, $hook): void {
           'title' => $item['content']['#title'],
         ],
       ];
+      if ($variables['element']['#bundle'] === 'oe_news') {
+        $variables['items'][$delta]['content']['#fields']['icon'] = 'external';
```
nagyad: IF news has related links field this will be applied on those too. That is a non desired case I believe, so let's fix that and cover with test that related links should not have icon for now, if you are using this approach.

nagyad (changes requested):

nagyad (dismissed):

sergepavle (approved):

nagyad (approved):

PR #925, opened 2021-08-20
EWPP-1449: Add additional suggestions for block templates based on co…

sergepavle:
…mponent library or/and branding.

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -1636,6 +1636,12 @@ function oe_theme_theme_suggestions_block_alter(array &$suggestions, array $vari
   }
   $suggestions[] = 'block__region_' . $block->getRegion();
   $suggestions[] = 'block__region_' . $block->getRegion() . '__' . $variables['elements']['#id'];
+
+  // Add support of template override which depending on
+  // ECL branding and ECL component library.
+  $branding = theme_get_setting('branding') ?? 'core';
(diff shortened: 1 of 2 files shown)


oe_theme.theme
```diff
@@ -1636,6 +1636,12 @@ function oe_theme_theme_suggestions_block_alter(array &$suggestions, array $vari
   }
   $suggestions[] = 'block__region_' . $block->getRegion();
   $suggestions[] = 'block__region_' . $block->getRegion() . '__' . $variables['elements']['#id'];
+
+  // Add support of template override which depending on
+  // ECL branding and ECL component library.
+  $branding = theme_get_setting('branding') ?? 'core';
+  $suggestions[] = $variables['theme_hook_original'] . '__' . $variables['elements']['#plugin_id'] . '__ecl_component_library_' . theme_get_setting('component_library');
```
imanoleguskiza, line 1643: I'm not entirely against this approach but it worry it might be a bit overkill. Unless there are other instances where it is usefull I am not sure if this is the correct approach. Then again, it is hard to predict if it will be useful or not in the future so I am conflicted.
A simpler solution could be to just check the "component_library" here in the breadcrumb template and add a white or blue background (https://github.com/ec-europa/ewcms/pull/1320/files#diff-fe0dbec8c437f4d659cd7e869aa7bb66b47c2ee91d68020f9f52034f40cc9f4aR11)

imanoleguskiza (changes requested):

PR #933, opened 2021-08-31
EWPP-1362: Reorganise publication date extra field class.

nagyad:


ademarco (changes requested):
The new extra_field_oe_theme_content_news_date field is missing from the hidden section on `./modules/oe_theme_content_news/config/install/core.entity_view_display.node.oe_news.teaser.yml`. It should be enough to re-export it.

ademarco (approved):

modules/oe_theme_content_news/config/install/core.entity_view_display.node.oe_news.teaser.yml
```diff
@@ -65,14 +70,20 @@ content:
     third_party_settings: {  }
 hidden:
   body: true
+  extra_field_oe_theme_content_news_date: true
```
yenyasinn, line 73: Do we need also update teaser in update path?

Also I'm not sure that we need  
```
  $display->save();
```
in oe_theme_content_news_post_update_00009 because teaser and full displays should be there and if they were removed it was done intentionally. Looks like it is just copy paste from oe_theme_content_news_post_update_00003.

yenyasinn (approved):

PR #945, opened 2021-09-10
EWPP-1486: Theme publication collection related fields.

nagyad:


modules/oe_theme_content_publication/oe_theme_content_publication.post_update.php
```diff
@@ -206,3 +206,23 @@ function oe_theme_content_publication_post_update_00010(): void {
...
+  $storage = \Drupal::entityTypeManager()->getStorage('entity_view_display');
+
+  $view_display = EntityViewDisplay::load($display_values['id']);
+  if ($view_display) {
+    $display = $storage->updateFromStorageRecord($view_display, $display_values);
+    $display->save();
+    return;
+  }
+
+  $display = $storage->createFromStorageRecord($display_values);
```
yenyasinn: I know that it is copy/paste but do we need to create full view? It has to be already in the system but if it was removed then site administrator doesn't need it.

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
```diff
@@ -0,0 +1,154 @@
...
+    $referencing_entities = [];
+
+    // Load all entities that have a reference to the given entity.
+    $entity_type_storage = $this->entityTypeManager->getStorage('node');
+    $query = $entity_type_storage->getQuery('OR');
+    $query->condition('oe_publication_publications.target_id', $entity->id());
+
+    $ids = $query->execute();
+    if ($ids) {
+      $referencing_entities = array_merge($referencing_entities, $entity_type_storage->loadMultiple($ids));
```
yenyasinn:
```suggestion
      return $entity_type_storage->loadMultiple($ids));
```

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    return $referencing_entities;
```
yenyasinn:
```suggestion
    return [];
```

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
```diff
@@ -0,0 +1,154 @@
...
+   * Get the publication collections that are referencing other publications.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The publication entity.
+   *
+   * @return array
+   *   The loaded publication entities.
+   */
+  protected function getReferencingCollections(EntityInterface $entity): array {
+    $referencing_entities = [];
```
yenyasinn:
```suggestion
```

We don't need $referencing_entities variable.

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * Get the publication collections that are referencing other publications.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The publication entity.
+   *
+   * @return array
+   *   The loaded publication entities.
+   */
+  protected function getReferencingCollections(EntityInterface $entity): array {
```
yenyasinn, line 160: It's ok for now, but @ademarco  might be we need list of referencing collections somewhere else? In this case it could be to placed in Drupal\oe_content_publication\PublicationNodeWrapper.

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
```diff
@@ -0,0 +1,154 @@
...
+    $referencing_entities = [];
+
+    // Load all entities that have a reference to the given entity.
+    $entity_type_storage = $this->entityTypeManager->getStorage('node');
+    $query = $entity_type_storage->getQuery('OR');
```
yenyasinn:
```suggestion
    $query = $entity_type_storage->getQuery();
```
Let's make it simpler, we don't need "or" condition here.

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
```diff
@@ -0,0 +1,154 @@
...
+    else {
+      $this->label = $this->t('Part of collection');
+    }
+
+    $cache = CacheableMetadata::createFromRenderArray($build);
+
+    $items = [];
+    foreach ($collections as $collection) {
+      $cache->addCacheableDependency($collection);
+      $items[] = [
```
yenyasinn, line 126:
```
      if ($collection->isPublished()) {
        $items[] = [
          'content' => [
            '#type' => 'link',
            '#title' => $collection->label(),
            '#url' => $collection->toUrl(),
            '#attributes' => [
              'class' => ['ecl-link', 'ecl-link--standalone'],
            ],
          ]
        ];
      }
```
I think we shouldn't show unpublished nodes here.
Using `content` allow to use field--multiple-values--pipe-separated.html.twig.
`$collection->toUrl(),` is simpler than `Url::fromRoute('entity.node.canonical', ['node' => $collection->id()]),`

modules/oe_theme_content_publication/templates/oe-theme-content-publication-collections.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
+/**
+ * @file
+ * Theme template for Publication collection extra field.
+ *
+ * @see \Drupal\oe_theme_content_publication\Plugin\ExtraField\Display\PublicationCollection::viewElements()
+ */
+#}
+{% for item in items %}{{ item }}{% if not loop.last %} | {% endif %}{% endfor %}
```
yenyasinn, line 9:
```suggestion
{% include '@oe_theme/field/field--multiple-values--pipe-separated.html.twig' %}
```
to have single template for all fields separated by pipe

tests/Functional/ContentPublicationRenderTest.php
```diff
@@ -364,6 +367,85 @@ public function testPublicationRendering(): void {
...
+    $this->drupalGet($node->toUrl());
+    $details_expected_values['items'][0] = [
+      'label' => 'Part of collection',
+      'body' => 'Test Publication collection node 1',
+    ];
+    $first_element = [
+      'label' => 'Identification',
+      'body' => 'ID 1, ID 2',
+    ];
+    array_unshift($details_expected_values['items'], $first_element);
```
yenyasinn:
```suggestion
    $details_expected_values = [
```
let's define `$details_expected_values` again, currently it is very hard to read and understand why we need array_unshift for.

tests/Functional/ContentPublicationRenderTest.php
```diff
@@ -364,6 +367,85 @@ public function testPublicationRendering(): void {
...
+        'target_id' => $node->id(),
+      ],
+      'oe_publication_date' => [
+        'value' => '2020-04-15',
+      ],
+      'oe_subject' => 'http://data.europa.eu/uxp/1000',
+      'oe_author' => 'http://publications.europa.eu/resource/authority/corporate-body/AASM',
+      'oe_content_content_owner' => 'http://publications.europa.eu/resource/authority/corporate-body/COMMU',
+      'uid' => 0,
+      'status' => 1,
```
yenyasinn, line 464: Also we can cover the case when collection is published/unpublished.

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
nagyad: I discussed with him and for now we leave it here. Once we need it somewhere else we remove.

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
nagyad: Yes, but we are in extra field context and not drupal field that template was made for. We can keep the custom template instead of hack the render array to be able to use the field--multiple-values--pipe-separated.html.twig.
On the other hand, we need an url object there, and that line I took from core entity reference revisions label formatter.

modules/oe_theme_content_publication/templates/oe-theme-content-publication-collections.html.twig
nagyad: Let's keep here this as custom template instead of forcing the field template on a custom extra field.

tests/Functional/ContentPublicationRenderTest.php
nagyad: Very good point.

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
yenyasinn: `toUrl()` also returns Url object and handle edge cases in node.
As I see https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/lib/Drupal/Core/Field/Plugin/Field/FieldFormatter/EntityReferenceLabelFormatter.php#L68 `$entity->toUrl()` is used and wrapped by try/catch.

modules/oe_theme_content_publication/oe_theme_content_publication.post_update.php
yenyasinn: Any comments?

modules/oe_theme_content_publication/src/Plugin/ExtraField/Display/PublicationCollection.php
```diff
@@ -0,0 +1,159 @@
...
+      $access = $collection->access('view', NULL, TRUE);
+      $cache->addCacheableDependency($access);
+      if (!$access->isAllowed()) {
+        continue;
+      }
+
+      $items[] = [
+        '#type' => 'link',
+        '#title' => $collection->label(),
+        '#url' => $collection->toUrl(),
```
yenyasinn, line 129: Would be better to wrap toUrl() by try/catch https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/lib/Drupal/Core/Field/Plugin/Field/FieldFormatter/EntityReferenceLabelFormatter.php#L68

nagyad: In the case you linked it is a generic approach for a formatter. In our case we are theming an extra field that holds nodes, so we no exactly that those have link template, we don't need to wrap it.

yenyasinn (dismissed):

ademarco (changes requested):
Translation is not loaded when printing a collection's title:
image

ademarco (dismissed):

yenyasinn (dismissed):

ademarco (dismissed):

yenyasinn (dismissed):

upchuk (approved):
Approving on behalf of @ademarco

yenyasinn (approved):

PR #973, opened 2021-10-26
EWPP-1609: Add badges paramater to list_item and featured_item patterns.

22Alexandra:


templates/patterns/file_translation/pattern-file-translation.html.twig
```diff
@@ -62,5 +67,7 @@
     },
     'description': more_info|default(''),
     'items': _translations,
-  }
+  },
+  'variant': 'thumbnail',
```
imanoleguskiza: The problem here is that this is not "backwards compatible", in the sense that if someone is using this pattern and does not add a thumbnail will get nothing. I propose we check whether we have a thumbnail or not and set the variant accordingly? cc @ademarco

imanoleguskiza (changes requested):

templates/patterns/file_translation/file_translation.ui_patterns.yml
```diff
@@ -48,3 +55,9 @@ file_translation:
       label: "More information"
       description: "Informative text about the available translations."
       preview: "Looking for another language which is not on the list? Find out why."
+    highlighted:
```
imanoleguskiza: We agreed with @ademarco we won't add this style of property and instead we will use the same "Badges" approach as we use in other patterns

imanoleguskiza (dismissed):

imanoleguskiza (approved):

ademarco (approved):

PR #993, opened 2021-12-01
EWPP-1779: Make Event CT content labels consistent with labels in other CTs

mferran-europe:


mferran-europe:
Discard, based on incorrect branch.

PR #997, opened 2021-12-06
EWPP-1787: Add extra field displaying event status messages.

22Alexandra:


modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
```diff
@@ -0,0 +1,204 @@
...
+        // Cache it by the event end date and add the timezone cache contexts.
+        $build['#title'] = $this->t('The livestream has ended, but the event is ongoing.');
+        $this->applyHourTag($build, $event->getEndDate());
+        $cacheable->addCacheContexts(['timezone']);
+        $cacheable->applyTo($build);
+        return $build;
+      }
+
+      // The event is ongoing but the livestream is yet to start.
+      if ($this->requestDateTime >= $event->getStartDate()->getPhpDateTime() && $this->requestDateTime < $event->getOnlineStartDate()->getPhpDateTime()) {
```
yenyasinn: We will re-use condition often, so EventNodeWrapper is a good place to store it. You can use isOnlinePeriodYetToCome from https://github.com/openeuropa/oe_content/pull/498.

`$this->requestDateTime >= $event->getStartDate()->getPhpDateTime()` also can be stored in EventNodeWrapper.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
```diff
@@ -0,0 +1,204 @@
...
+
+    // If the event is not over then apply time-based tags, so that it can be
+    // correctly invalidated once the event is over.
+    if (!$event->isOver($this->requestDateTime)) {
+      $this->applyHourTag($build, $event->getEndDate());
+    }
+
+    // If we have a link for livestream the livestream messages have priority
+    // over the event status ones.
+    if (!$entity->get('oe_event_online_link')->isEmpty()) {
```
yenyasinn: In the spec there is a point that livestream info should be visible if Online type != None. So it is checked in EWPP-1802. Would be inconsistent to show message here but not to show livestream due to implementation in EWPP-1802.
Screenshot 2021-12-07 at 19 50 38

Or might be we need a single method in EventNodeWrapper that will check all required fields to enable livestream feature.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
```diff
@@ -0,0 +1,204 @@
...
+        // Cache it by the event end date and its livestream start date.
+        $this->applyHourTag($build, $event->getEndDate());
+        $this->applyHourTag($build, $event->getOnlineStartDate());
+        $livestream_date = $this->dateFormatter->format($entity->get('oe_event_online_dates')->start_date->getTimestamp(), 'oe_event_date_hour_timezone');
+        $build['#title'] = $this->t('This event has started. The livestream will start at @date.', ['@date' => $livestream_date]);
+        return $build;
+      }
+
+      // If the livestream is ongoing, cache it by its end date.
+      if ($this->requestDateTime >= $event->getOnlineStartDate()->getPhpDateTime()) {
```
yenyasinn: Also `isOnlinePeriodOver()` can be used.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
```diff
@@ -0,0 +1,204 @@
...
+      // Use the 'warning' variant with the 'livestream' icon.
+      $build['#variant'] = 'warning';
+      $build['#icon'] = 'livestreaming';
+
+      // The livestream is over, but the event is still ongoing.
+      if ($event->isOnlineOver($this->requestDateTime) && $this->requestDateTime >= $event->getStartDate()->getPhpDateTime()) {
+        // Cache it by the event end date and add the timezone cache contexts.
+        $build['#title'] = $this->t('The livestream has ended, but the event is ongoing.');
+        $this->applyHourTag($build, $event->getEndDate());
+        $cacheable->addCacheContexts(['timezone']);
```
yenyasinn:
```suggestion
```
it was added by $this->applyHourTag()

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * Date formatter service.
+   *
+   * @var \Drupal\Core\Datetime\DateFormatterInterface
+   */
+  protected $dateFormatter;
+
+  /**
+   * ProgrammeExtraField constructor.
```
nagyad: Needs to be fixed.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
```diff
@@ -0,0 +1,202 @@
...
+    // Render event status messages using the ECL Message component.
+    $event = new EventNodeWrapper($entity);
+    $build = ['#theme' => 'oe_theme_content_event_status_message'];
+
+    $cacheable = CacheableMetadata::createFromRenderArray($build);
+
+    // If the event is not over then apply time-based tags, so that it can be
+    // correctly invalidated once the event is over.
+    if (!$event->isOver($this->requestDateTime)) {
+      $this->applyHourTag($build, $event->getEndDate());
```
nagyad, line 98: Logic issue, I set the event as planned with the event dates matching the online dates with event cancelled, the message is correct, but when I set it to Rescheduled, the event shows the message "This event has started. You can also watch it via livestream." even when I set the event state to cancelled or postponed, the message doesn't change, even when I'm logged in as admin or logged out. This is due to the fact that you don't exit here, and it continues until the line 147. Please make sure such case is covered by test after fixing it.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    // If we have an online type set the livestream messages have priority
+    // over the event status ones.
+    if ($event->hasOnlineType()) {
```
nagyad, line 119: I think the entity itself should be added as cacheable dependency. Currently the entity is not there.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
```diff
@@ -0,0 +1,202 @@
...
+    $build['#icon'] = 'warning';
+    if ($event->isPostponed()) {
+      $build['#title'] = $this->t('This event has been postponed.');
+    }
+    if ($event->isCancelled()) {
+      $build['#title'] = $this->t('This event has been cancelled.');
+    }
+    if ($event->isRescheduled()) {
+      $build['#title'] = $this->t('This event has been rescheduled.');
+    }
```
nagyad: These will be never reached if we have online type and dates set even though the event status is set to one of these.

nagyad (changes requested):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
22Alexandra: The livestream messages have priority over these ones, as requested.

Regarding online type/dates/link, according to Evegenii's [comment](https://github.com/openeuropa/oe_theme/pull/997#discussion_r764186721) to avoid inconsistency, I'll check the Online type instead of the Online link as requested in the ticket.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
22Alexandra: We'll need confirmation from PO.

nagyad (changes requested):
I have clarified the logic with the PO. As the ticket says, only "As planned" messages get overruled, not the "postponed", "cancelled" or "rescheduled" cases. Please refactor the logic accordingly.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
22Alexandra: Updated the logic.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/StatusMessage.php
22Alexandra: Fixed the logic displaying event status messages (if it's not as planned) over livestream ones.

sergepavle (dismissed):

imanoleguskiza (dismissed):

sergepavle (dismissed):

nagyad (dismissed):

sergepavle (approved):

imanoleguskiza (approved):

PR #999, opened 2021-12-07
EWPP-1802: Improve live stream display.

yenyasinn:


sergepavle (dismissed):

22Alexandra (dismissed):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/LivestreamExtraField.php
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(ContentEntityInterface $entity) {
+    $event = EventNodeWrapper::getInstance($entity);
+
+    if (!$event->hasOnlineType()
+      || !$event->hasOnlineLink()
+      || !$event->hasOnlineDates()
+      || !$event->isOnlinePeriodYetToCome($this->requestDateTime)) {
```
ademarco: It's ok to return empty if type, link and dates are empty (as, due to custom validation logic, filling one requires the other two to be filled as well), but we should isolate the isOnlinePeriodYetToCome check, so that we apply cache tags when type, link and dates are filled, but it's still not time to show the field.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/OnlineDescriptionExtraField.php
```diff
@@ -0,0 +1,62 @@
...
+    if (!$event->hasOnlineType() || !$event->hasOnlineLink() || !$event->hasOnlineDates()) {
+      // All online fields have to be filled to show online information.
+      return [];
+    }
+
+    $build = [];
+
+    if ($event->isOnlinePeriodYetToCome($this->requestDateTime)) {
+      // Invalidate cache tags when livestream will be started.
+      $this->applyHourTag($build, $event->getOnlineStartDate());
```
ademarco, line 45: This is correct and it's basically what outlined in the comment above, on the other field, but we should consider marking the field as empty, instead of returning and empty array, as per extra field docs:

image

ademarco (changes requested):

ademarco (dismissed):
We shouldn't show the link until the time is right:

image

nagyad (dismissed):

sergepavle (dismissed):

nagyad (dismissed):

sergepavle (dismissed):

nagyad (dismissed):

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -568,7 +572,16 @@ public function testEventRendering(): void {
     $this->cronRun();
     $this->drupalGet($node->toUrl());
 
-    $description_summary = $this->assertSession()->elementExists('css', '.ecl', $details_content);
+    // It is time when livestream is active.
```
imanoleguskiza: What does this mean?

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/LivestreamExtraField.php
```diff
@@ -0,0 +1,119 @@
...
+    // render the date only.
+    if ($event->isOnlinePeriodYetToCome($this->requestDateTime)) {
+      $this->applyHourTag($build, $event->getOnlineStartDate());
+    }
+    $build['#date'] = $this->t('Starts on @date', [
+      '@date' => $this->dateFormatter->format($event->getOnlineStartDate()->getTimestamp(), 'oe_event_long_date_hour'),
+    ]);
+    // If the livestream is ongoing, we add the livestream link.
+    if ($event->isOnlinePeriodActive($this->requestDateTime)) {
+      $link = $entity->get('oe_event_online_link')->first();
```
imanoleguskiza, line 110: We need to cache by the end date

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

nagyad (dismissed):

sergepavle (approved):

nagyad (approved):

PR #1003, opened 2021-12-22
EWPP-1830: Added optional 'size' field for items of icons_with_text p…

mferran-europe:


nagyad (dismissed):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DetailsExtraField.php
```diff
@@ -98,10 +98,12 @@ public function viewElements(ContentEntityInterface $entity) {
         'items' => [
           [
             'icon' => 'file',
+            'size' => 's',
```
yenyasinn:
```suggestion
            'size' => 'xs',
```

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DetailsExtraField.php
```diff
@@ -98,10 +98,12 @@ public function viewElements(ContentEntityInterface $entity) {
...
             'text' => $this->getRenderableSubject($entity),
           ],
           [
             'icon' => 'calendar',
+            'size' => 's',
```
yenyasinn:
```suggestion
            'size' => 'xs',
```

templates/patterns/icons_with_text/icons_with_text.ui_patterns.yml
```diff
@@ -8,11 +8,13 @@ icons_with_text:
       description: "List of items, each having an 'icon' and a 'text' field"
       preview:
         - icon: "file"
+          size: "s"
```
yenyasinn: Might be we don't need this line. Size `m` will be used by default.

templates/patterns/icons_with_text/icons_with_text.ui_patterns.yml
```diff
@@ -8,11 +8,13 @@ icons_with_text:
...
           text: "Culture | Innovation"
         - icon: "calendar"
           text: "15-16 November 2019, 08:00 AM"
         - icon: "location"
           text: "Brussels, Belgium"
+          size: "s"
```
yenyasinn:
```suggestion
          size: "xs"
```
as in Event to show difference with default size.

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -225,9 +225,11 @@ public function testEventRendering(): void {
         [
           'icon' => 'file',
           'text' => 'Financing',
+          'size' => 's',
```
yenyasinn:
```suggestion
          'size' => 'xs',
```

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -225,9 +225,11 @@ public function testEventRendering(): void {
...
         ], [
           'icon' => 'calendar',
           'text' => '28 February 2020, 01:00',
+          'size' => 's',
```
yenyasinn:
```suggestion
          'size' => 'xs',
```

tests/src/Kernel/fixtures/rendering.yml
```diff
@@ -2670,12 +2670,14 @@
           text: "Culture | Innovation"
         -
           icon: "calendar"
+          size: "s"
```
yenyasinn, line 2673:
```suggestion
          size: "xs"
```

templates/patterns/icons_with_text/icons_with_text.ui_patterns.yml
22Alexandra: Let's leave it so we see the diferent sizes on the /components page.

tests/src/Kernel/fixtures/rendering.yml
22Alexandra: No need for this change since this is testing the pattern.

**EDIT:** Sorry, my bad. I updated the rendering for the next item https://github.com/openeuropa/oe_theme/pull/1003/files#diff-c393fd6edf31364b5ddc27ffd11f690f3ece540c7b78d6e3c26a2b3dcc5a4ee3R2680

templates/patterns/icons_with_text/icons_with_text.ui_patterns.yml
22Alexandra: No need for this either, is just the pattern.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DetailsExtraField.php
```diff
@@ -98,10 +98,12 @@ public function viewElements(ContentEntityInterface $entity) {
         'items' => [
           [
             'icon' => 'file',
+            'size' => 'xs',
```
yenyasinn: Screenshot 2022-01-07 at 13 14 33

Icons have size `xs` in Full and Teaser views both. Ticket EWPP-1830 asks to change Teaser only.
Looks like icon size should be set in Drupal\oe_theme_content_event\Plugin\ExtraField\Display\TeaserDetailsExtraField and be passed to addRenderableLocation and addRenderableOnlineType as an optional argument.

templates/patterns/icons_with_text/icons_with_text.ui_patterns.yml
yenyasinn: Yes, so I suggest to show `xs` since it's used in the Event. Would be nice to group 2 items with icon `xs` and then 2 items with default site to see different difference. It's just a proposal...

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DetailsExtraField.php
22Alexandra: I agree that the ticket says only Teaser but Teaser and Full view modes are using the same pattern for rendering this information, are we sure we want this change only in Teaser?

yenyasinn: You can pass different sizes of icons to pattern as an option.
Image from spec: https://webgate.ec.europa.eu/fpfis/wikis/download/attachments/416223976/Event%20CT%20-%[…]esktop.png?version=3&modificationDate=1598608568914&api=v2
Icon as is in full, and small in teaser

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/TeaserDetailsExtraField.php
```diff
@@ -42,6 +42,9 @@ public function viewElements(ContentEntityInterface $entity) {
 
     $this->addRenderableLocation($build, $entity);
     $this->addRenderableOnlineType($build, $entity);
+    // Override default icon size.
+    $build['#fields']['items'][0]['size'] = 'xs';
```
yenyasinn: Since Venue and Online type are optional fields, $build['#fields']['items'] can be empty. It has no sense to set icon size for such case.

```
  foreach ($build['#fields']['items'] as $key => $item) {
      $build['#fields']['items'][$key]['size'] = 'xs';
  }
```
foreach won't process empty array ^^.

yenyasinn (approved):

nagyad (approved):

PR #1011, opened 2022-01-19
EWPP-1888: Handle non-eu logos provided by ECL 3.2.

yenyasinn:


upchuk (changes requested):
We said in the refinement we will check for the existence of the file on the system and then default to EN. i understand these two lang codes have different codes and we need a mapping but what about other languages (like catalan)? Or future languages that are added to the system and have a matching correct code.

oe_theme.theme
```diff
@@ -1242,11 +1242,27 @@ function oe_theme_preprocess_block__system_branding_block(&$variables) {
...
+      'tr' => 'tu',
+    ];
+    if (isset($language_icon_mapping[$language_code])) {
+      // Set correct icon for Norwegian and Turkish languages.
+      $variables['current_language_id'] = $language_icon_mapping[$language_code];
+    }
+    else {
+      // Check whether ECL provides a logo for the language.
+      $ecl_logo = DRUPAL_ROOT . '/' . drupal_get_path('theme', 'oe_theme') . '/dist/' . $variables['ecl_component_library'] . '/images/logo/';
+      $ecl_logo .= $variables['ecl_component_library'] === 'eu' ? 'standard-version/positive/' : '';
```
22Alexandra, line 1261: No need for this. Check oe_theme_preprocess() function that already builds the correct logo path.

oe_theme.theme
```diff
@@ -1242,11 +1242,27 @@ function oe_theme_preprocess_block__system_branding_block(&$variables) {
...
   if (!EuropeanUnionLanguages::hasLanguage($language_code)) {
-    $variables['current_language_id'] = 'en';
+    $language_icon_mapping = [
+      'nb' => 'no',
+      'tr' => 'tu',
+    ];
+    if (isset($language_icon_mapping[$language_code])) {
+      // Set correct icon for Norwegian and Turkish languages.
+      $variables['current_language_id'] = $language_icon_mapping[$language_code];
+    }
```
22Alexandra: Instead of mapping the language codes here, can we move the logic in the twig filter that currently just returns the exact same language code if it's not an EU language?\Drupal\oe_theme_helper\TwigExtension\TwigExtension::toInternalLanguageId

22Alexandra (changes requested):

oe_theme.theme
yenyasinn: toInternalLanguageId supports European languages only. We can reorganise it as a part of EWPP-1502 but currently it is out of scope. We will be able to use toInternalLanguageId for Norwegian language, but Turkish language uses "tr" language code but ECL team decided to name logo with "tu" suffix that we don't use anywhere. The best option is to ask ECL team to rename Turkish logo with "tr" in logo name but I suppose it will be long. So imho, for now, it is proper solution.

oe_theme.theme
yenyasinn: oe_theme_preprocess build path based on docroot (/build/dist/...), here we need absolute path (/var/www/.../build), so need to build again. 
Path DRUPAL_ROOT . $variables['ecl_image_logo'] looks like /var/www/.../build/build/...  that can't be used

22Alexandra (dismissed):

oe_theme.theme
```diff
@@ -1242,11 +1242,27 @@ function oe_theme_preprocess_block__system_branding_block(&$variables) {
...
+    else {
```
sergepavle, line 1258: I don't think that we should have here `else`, otherwise we will drop checking if related logo.svg exists for Turkish and Norwegian languages.
These 2 languages should be checked (file presence) as other languages non-eu languages.

sergepavle (changes requested):

oe_theme.theme
yenyasinn: We don't test that logos for European languages exist because we know that they are.
So the same logic that logos for Norwegian and Turkish exist - ECL provides them.
`else` is used as a fallback since site can have any other language that isn't determined as European or supported non-European.

sergepavle (dismissed):

oe_theme.theme
sergepavle: In this case, we will rely on ECL but anyway exist a chance that logo file can be removed/renamed intentionally or by mistake. 
Probably it is fine to leave as is.

22Alexandra (approved):

sergepavle (approved):

upchuk (approved):

PR #1014, opened 2022-01-24
EWPP-1564: Theme Timeline's paragraph introduction field.

22Alexandra:


templates/paragraphs/paragraph--oe-timeline.html.twig
```diff
@@ -4,8 +4,12 @@
  * Theme override to display the 'timeline' type paragraph.
  */
 #}
+{% if content.field_oe_text_long|field_value|render is not empty %}
+  {% set introduction = content.field_oe_text_long|field_value|render %}
+{% endif %}
```
nagyad: Looking around I think we can follow the example to put this into a variable and check that if it's not empty and then assign ti to the render array. My concern here is we call render twice here.

nagyad (changes requested):

templates/patterns/timeline/pattern-timeline.html.twig
```diff
@@ -26,6 +26,11 @@
     {{ title }}
   </h2>
 {% endif %}
+{% if introduction %}
+  <div id="timeline-introduction">
```
nagyad: I think we should apply text related ecl classes because now the text is black. Should follow a normal paragraph text styling.

nagyad (changes requested):

templates/patterns/timeline/pattern-timeline.html.twig
22Alexandra: That's how we render that particular paragraph field everywhere.

yenyasinn:
```suggestion
  <div class="timeline-introduction">
```
"id" attribute has to be unique on the page. Since it is a paragraph, then few elements with the same id can be presented on the page.
```suggestion
  <div class="ecl">
```
Need to check how we show text in other paragraphs. As I see `#404040` color is used in .ecl class of wrapper in other paragraphs.

22Alexandra: It's not needed. Please check how we render this paragraph field.

nagyad: Once we apply the text formats, the corresponding wrapper with the ecl class will be applied on render.

nagyad (dismissed):

templates/patterns/timeline/pattern-timeline.html.twig
upchuk: Not sure it's a good idea to use a id/class for testing purposes only. Definitely not an ID.

Also, maybe we need an ECL class because i am seeing the text not wrapped in the same class as when i use this same field in a rich text paragraph. And yes, this is in EWCMS with the text format on the field:

Screenshot 2022-01-27 at 11 46 04

upchuk (changes requested):

upchuk (approved):

nagyad (approved):

PR #1031, opened 2022-02-09
EWPP-1973: Remove usage of t() in test assertions.

yenyasinn:


upchuk (dismissed):
This is going severely out of scope IMO.

tests/features/site-branding.feature
```diff
@@ -72,3 +72,41 @@ Feature: Site branding
     Then I should see the "page header" element in the "page"
     When I am on "the user registration page"
     Then I should see the "page header" element in the "page"
+
+  @javascript @enable-non-eu-language
```
yenyasinn, line 76: This is a refactoring of the test from EWPP-1854. Since we need to ensure correctness of the data in the logo in different languages we need to pass them into the Behat context.

tests/src/Behat/OeThemeTestContext.php
```diff
@@ -337,44 +368,20 @@ public function setCorporateLibrary(string $component_library): void {
...
-   * @param bool $presence
-   *   Wheres presence (TRUE) or absence (FALSE) should be asserted.
+   * The ECL doesn't provide a way to determine type of the footer except as
+   * to check the existence of the logo.
+   * 'European Union' has logo, 'European Commission' doesn't.
+   *
+   * @return string
+   *   'European Union' or 'European Commission'.
    */
-  public function assertFooter(string $component_library, bool $presence): void {
```
yenyasinn, line 347: assertFooter includes 3 functions - checks EU footer, checks EC footer, ensures that footer doesn't exist.
For logic simplification I've rewritten it, otherwise it is mess of conditions.

tests/features/site-branding.feature
```diff
@@ -72,3 +72,41 @@ Feature: Site branding
...
+    And I open the language switcher dialog
+    And I click "<language_selector>"
+    Then the header logo should contain accessibility attributes with link "<link>" label "<label>" alt "<alt>" title "<title>"
+
+    Examples:
+      | language_selector | link                                      | label                 | alt                 | title          |
+      | English           | https://european-union.europa.eu/index_en | Home - European Union | European Union flag | European Union |
+      | български         | https://european-union.europa.eu/index_bg | Home - European Union | European Union flag | European Union |
+      # Non-EU language.
+      | Icelandic         | https://european-union.europa.eu          | Home - European Union | European Union flag | European Union |
```
22Alexandra, line 93: Label, alt and title are exactly the same for each language so why can't we just adapt the existing scenarios testing the logos for every language already? Now we have two different scenarios for each component library testing the exact same thing the only difference being the URL.

tests/src/Behat/OeThemeTestContext.php
```diff
@@ -400,32 +407,17 @@ public function assertEuMobileLogo(string $language): void {
...
-    $img = $link->find('css', 'img');
-    Assert::assertEquals($img_alt, $img->getAttribute('alt'));
-    Assert::assertEquals($img_title, $img->getAttribute('title'));
+    $logo_link = $this->getSession()->getPage()->find('css', $link_selector);
+
+    Assert::assertEquals($link, $logo_link->getAttribute('href'));
+    Assert::assertEquals($label, $logo_link->getAttribute('aria-label'));
+    $img = $logo_link->find('css', 'img');
+    Assert::assertEquals($alt, $img->getAttribute('alt'));
+    Assert::assertEquals($title, $img->getAttribute('title'));
```
22Alexandra, line 420: Why do we need to change the existing assertions since the info is the same for each language?

tests/src/Behat/OeThemeTestContext.php
```diff
@@ -284,19 +284,50 @@ protected function getDemoPageByTitle(string $title): NodeInterface {
...
+    $img = $logo_link->find('css', 'img');
+    Assert::assertEquals($img_alt, $img->getAttribute('alt'));
+    Assert::assertEquals($img_title, $img->getAttribute('title'));
   }
 
   /**
    * Assert given corporate footer absence on page.
    *
-   * @Then I should not see the :component_library footer( instead)
+   * @Then I should not see the :component_library footer
```
22Alexandra: This has a reason to be here for making sense in cases such as:
```
Then I should not see the "European Commission" footer
But I should see the "European Union" footer instead
```

tests/src/Behat/OeThemeTestContext.php
```diff
@@ -337,44 +368,20 @@ public function setCorporateLibrary(string $component_library): void {
...
-      throw new \Exception("The {$component_library} corporate footer block was expected to be {$expectation} but it is not.");
-    }
-
-    if ($logo && $presence) {
-      $link = $page->find('css', '.ecl-footer-' . $this->getEclBranding() . '__logo-link');
-      Assert::assertEquals('https://european-union.europa.eu/index_en', $link->getAttribute('href'));
-      Assert::assertEquals(t('Home - European Union')->__toString(), $link->getAttribute('aria-label'));
-      $img = $link->find('css', 'img');
-      Assert::assertEquals(t('European Union flag')->__toString(), $img->getAttribute('alt'));
-      Assert::assertEquals(t('European Union')->__toString(), $img->getAttribute('title'));
```
22Alexandra, line 376: I don't think the whole refactoring of this method + 2 other methods should be done in this PR since the scope of the issue is to simply assert the string which you now pass in different scenarios (being the same for each language).

22Alexandra (dismissed):

tests/src/Behat/OeThemeTestContext.php
yenyasinn: There is a comment above 
```
assertFooter includes 3 functions - checks EU footer, checks EC footer, ensures that footer doesn't exist.
For logic simplification I've rewritten it, otherwise it is mess of conditions.
```
I couldn't pass to assertFooter any data, so I  needed to rewrite it.

tests/features/site-branding.feature
yenyasinn: As usual, there are different possible ways of implementation.

Actually I've rewritten existing test completely because it doesn't work as expected. But Danny said that it is out of scope so I put tests for footers separately.  I've left a todo there `Test should be rewritten since it tests European Union style logo on the English homepage and European Commision style logo on the English registration page for each iteration.`

`Label, alt and title are exactly the same for each language`. Yes, it is, but URLs are different. Also label, alt, and title can be changed in admin interface. So I've provided common interface to assert footer and it doesn't matter what data we pass to it.

Additionally, all these strings are translatable. Translations are imported in EWCMS, so we can reuse this context in EWCMS as here https://github.com/ec-europa/ewcms/pull/1627/files with any data.

tests/src/Behat/OeThemeTestContext.php
yenyasinn: Yep, but it isn't used. 
Returned back.

tests/src/Behat/OeThemeTestContext.php
yenyasinn: Because we shouldn't rewrite context every time when strings are changed in the footer. 
I've provided an interface where any strings can be passed, on any language. So just need to provide correct expected strings in the test.

tests/src/Behat/OeThemeTestContext.php
22Alexandra: It is used and I in fact gave you an example from existing code.

22Alexandra (dismissed):

nagyad (approved):

22Alexandra (approved):

PR #1039, opened 2022-02-15
EWPP-1983: Replace event date fields with timezone ones.

nagyad:


sergepavle (commented):
Some findings:

1. Seems like for Livestream field we don't have timezone visible:
image

2. is it expected that we are doing duplication of timezone for start and end date (on node edit page we are selecting timezone only for both dates):

image

nagyad:
We have duplicate because we are using php format, and that is applied to each date one by one.

sergepavle (dismissed):

imanoleguskiza (approved):

yenyasinn (approved):

22Alexandra (approved):
blind

PR #1042, opened 2022-02-18
EWPP-1996: Fix DefaultNodeMetadataEventSubscriber.

22Alexandra:


modules/oe_theme_helper/src/EventSubscriber/DefaultNodeMetadataEventSubscriber.php
```diff
@@ -75,7 +75,7 @@ public function getNode(NodeMetadataEvent $event): void {
 
     if ($this->currentRouteMatch->getRouteName() === 'entity.node.revision') {
       $node_revision = $this->currentRouteMatch->getParameter('node_revision');
-      $node = $this->entityTypeManager->getStorage('node')->loadRevision($node_revision);
+      $node = $this->entityTypeManager->getStorage('node')->loadRevision($node_revision->getRevisionId());
```
upchuk: Please see why this is not failing any test with 9.2 since with that version I believe we used to get an ID and not an entity object as a route parameter. if you confirm there is a difference between 9.2 and 9.3, you should check here if it's an entity or just an ID and put a comment wy.

upchuk (changes requested):

upchuk (dismissed):

upchuk (approved):

yenyasinn (approved):

nagyad (approved):

PR #1051, opened 2022-03-01
EWPP-2016: Improve display of person teaser.

imanoleguskiza:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,8 @@
         "openeuropa/rdf_skos": "^1.0",
         "openeuropa/task-runner-drupal-project-symlink": "^1.0-beta5",
         "phpspec/prophecy-phpunit": "^2",
-        "symfony/dom-crawler": "^4.4.12"
+        "symfony/dom-crawler": "^4.4.12",
+        "phpmd/phpmd": "2.11.1"
     },
     "scripts": {
         "post-install-cmd": "./vendor/bin/run drupal:site-setup",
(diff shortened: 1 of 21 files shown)


modules/oe_theme_content_person/oe_theme_content_person.module
```diff
@@ -11,6 +11,20 @@ use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\oe_theme\ValueObject\ImageValueObject;
 
+/**
+ * Implements hook_theme().
+ */
+function oe_theme_content_person_theme() {
```
sergepavle:
```suggestion
function oe_theme_content_person_theme(): array {
```

sergepavle (dismissed):

templates/field/field--node--oe-departments--oe-person--teaser.html.twig
```diff
@@ -6,4 +6,4 @@
  * @see ./core/themes/stable/templates/field/field.html.twig
  */
 #}
-{%- include 'field--multiple-values--pipe-separated.html.twig' -%}
+{%- include 'field--multiple-values--comma-separated.html.twig' -%}
```
22Alexandra, line 9: This change has to be done for all the departments fields according to the ticket.

22Alexandra (changes requested):
This request from the ticket is missing: `Point 1. 2. and 3. should apply to all instances of the content item pattern, not only on person teaser.`

So we need to hide the contact name everywhere we render contacts and separate departments by comma instead of pipe too.

22Alexandra (dismissed):

composer.json
```diff
@@ -58,7 +58,8 @@
         "openeuropa/rdf_skos": "^1.0",
         "openeuropa/task-runner-drupal-project-symlink": "^1.0-beta5",
         "phpspec/prophecy-phpunit": "^2",
-        "symfony/dom-crawler": "^4.4.12"
+        "symfony/dom-crawler": "^4.4.12",
+        "phpmd/phpmd": "2.11.1"
```
yenyasinn: Do we still need it? If not, can be removed before merge.

yenyasinn (dismissed):

yenyasinn (dismissed):

yenyasinn (approved):

22Alexandra (approved):

PR #1059, opened 2022-03-07
EWPP-1978: Enable failure of tests on javascript errors.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/tests/modules/oe_theme_js_test/src/Controller/UiPatterns.php
+++ b/tests/modules/oe_theme_js_test/src/Controller/UiPatterns.php
@@ -76,15 +76,6 @@ class UiPatterns extends ControllerBase {
       ],
     ];
 
-    $script = <<<EndOfScript
-window.onerror = function(error, url, line) {
-  if (typeof(window.collectedErrors) != 'object') {
-    window.collectedErrors = [];
-  }
-  window.collectedErrors.push({type:'error', data:'ERR:'+error+' URL:'+url+' L:'+line});
-}
-EndOfScript;
-
     $build['#attached']['html_head'][] = [
       [
         '#tag' => 'script',
--- a/tests/modules/oe_theme_js_test/src/Controller/UiPatterns.php
+++ b/tests/modules/oe_theme_js_test/src/Controller/UiPatterns.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Drupal\oe_theme_js_test\Controller;
 
 use Drupal\Core\Controller\ControllerBase;
-use Drupal\Core\Render\Markup;
 
 /**
  * Returns responses for UI Patterns test routes.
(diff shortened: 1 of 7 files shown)


22Alexandra (dismissed):

tests/modules/oe_theme_js_test/src/Controller/UiPatterns.php
```diff
@@ -76,15 +76,6 @@ public function contextNav(): array {
       ],
     ];
 
-    $script = <<<EndOfScript
```
yenyasinn, line 79: There is the code below on line 85 `'#value' => Markup::create($script),`. `$script` is undefined there after removing it here.

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
```diff
@@ -31,6 +34,15 @@ class JavascriptBehavioursTest extends WebDriverTestBase {
    */
   protected $defaultTheme = 'stark';
 
+  /**
+   * {@inheritdoc}
+   */
+  protected function tearDown(): void {
+    $this->failOnJavascriptErrors();
```
yenyasinn, line 41: Looks like failOnJavascriptErrors in this place doesn't catch js errors.

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
```diff
@@ -196,23 +208,7 @@ public function testEclDatePicker(): void {
...
-});
-return result;
-})()
-EndOfScript;
-    $errors = $this->getSession()->evaluateScript($script);
-    if ($errors) {
-      throw new \RuntimeException('Javascript error: ' . str_replace('[NLS]', PHP_EOL, $errors));
-    }
-
+    $this->failOnJavascriptErrors();
```
yenyasinn: It catches errors related to context_nav. But what is about other tests? For example issues in js/ecl_datepicker.js can be caught by testEclDatePicker().
Might be we can implement something like:
```
  /**
   * @inheritdoc
   */
  protected function drupalGet($path, array $options = [], array $headers = []) {
    $out = parent::drupalGet($path, $options, $headers = []);

    $this->failOnJavascriptErrors();
    return $out;
  }
```
It will catch js errors in all tests in JavascriptBehavioursTest.

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
sergepavle: Exposing of js error in this place doesn't work if we had some failure before.
Also, it will not throw error if previously did `$this->failOnJavascriptErrors()`.

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
sergepavle: Good idea. I've used overriding of this method.

yenyasinn (dismissed):

22Alexandra (dismissed):

22Alexandra (approved):

yenyasinn (approved):

PR #1066, opened 2022-03-23
EWPP-2071: Add external icon to external custom footer links.

nagyad:


tests/src/Functional/CorporateFooterRenderTest.php
```diff
@@ -712,7 +770,9 @@ public function testCorporateFooterRendering(): void {
       'label' => 'Custom contact 1',
       'href' => 'http://example.com/custom-contact-1',
     ];
-    $this->assertListLink($actual, 'standardised', $expected);
+    $this->assertListLink($actual, 'standardised', $expected, TRUE);
+    // We should have the external icon present.
+    $this->assertNotEmpty($actual->find('css', 'svg.ecl-icon.ecl-icon--xs.ecl-link__icon'));
```
22Alexandra: All the assertions ```// We should have the external icon present.
    $this->assertNotEmpty($actual->find('css', 'svg.ecl-icon.ecl-icon--xs.ecl-link__icon'));``` are not actually checking for the external icon but for any icon. We need to assert that the `svg` tag contains or not the actual `#external` icon for each case.

tests/src/Functional/CorporateFooterRenderTest.php
```diff
@@ -802,11 +862,14 @@ protected function assertFooterPresence(string $branding, int $expected_section_
    *   Ecl branding, core/standardised.
    * @param array $expected
    *   The expected data.
+   * @param bool $icon
+   *   If the link should have an icon, defaults to FALSE.
    */
-  protected function assertListLink(NodeElement $actual, string $branding, array $expected): void {
+  protected function assertListLink(NodeElement $actual, string $branding, array $expected, bool $icon = FALSE): void {
```
22Alexandra, line 905: This method shouldn't be changed because the fact that there is an icon doesn't ensure that is the external one and the core branding doesn't support site specific footer links anyway.

We can revert the changes for this method and simply assert the external icon is present or not in the test method.

22Alexandra (changes requested):

oe_theme.theme
```diff
@@ -1706,6 +1713,36 @@ function oe_theme_preprocess_oe_corporate_blocks_eu_footer(array &$variables) {
...
+
+  foreach ($variables['site_specific_footer']['other_links'] as $section => $links) {
+    if (empty($links['links'])) {
+      continue;
+    }
+
+    foreach ($links['links'] as $index => $link) {
+      if (empty($link['href'])) {
+        continue;
+      }
```
22Alexandra: The URL is always required, when can it be empty?

tests/src/Functional/CorporateFooterRenderTest.php
```diff
@@ -600,7 +639,32 @@ public function testCorporateFooterRendering(): void {
...
+    $this->assertEquals('Section 1', $actual->getText());
+
+    $actual = $subsection->find('css', 'ul li:nth-child(1) > a');
+    $expected = [
+      'label' => 'Custom link altered',
+      'href' => '/build/',
+    ];
     $this->assertListLink($actual, 'standardised', $expected);
+    // We should not have the external icon present.
+    $this->assertEmpty($actual->find('css', 'svg.ecl-icon.ecl-icon--xs.ecl-link__icon'));
```
22Alexandra:
```suggestion
    // We should not have any icon present.
    $assert->elementNotExists('css', 'svg.ecl-icon.ecl-icon--xs.ecl-link__icon', $actual);
```

tests/src/Functional/CorporateFooterRenderTest.php
```diff
@@ -681,9 +751,11 @@ public function testCorporateFooterRendering(): void {
     $actual = $subsection->find('css', 'ul li:nth-child(1) > a');
     $expected = [
       'label' => 'Custom link altered',
-      'href' => 'http://example.com/custom-link-altered',
+      'href' => '/build/',
     ];
     $this->assertListLink($actual, 'standardised', $expected);
+    // We should not have the external icon present.
+    $this->assertEmpty($actual->find('css', 'svg.ecl-icon.ecl-icon--xs.ecl-link__icon'));
```
22Alexandra:
```suggestion
    // We should not have any icon present.
    $assert->elementNotExists('css', 'svg.ecl-icon.ecl-icon--xs.ecl-link__icon', $actual);
```

22Alexandra (changes requested):

22Alexandra (approved):

22Alexandra (changes requested):
There is a Twig function (`ecl_footer_links`) currently used in the footer templates that is supposed to add the external icon but it doesn't seem working. @nagyad can you please check if we can fix it instead of adding the helper function for the preprocess?

oe_theme.theme
```diff
@@ -1706,6 +1713,36 @@ function oe_theme_preprocess_oe_corporate_blocks_eu_footer(array &$variables) {
...
+    if (empty($links['links'])) {
+      continue;
+    }
+
+    foreach ($links['links'] as $index => $link) {
+      if (!$link['href'] instanceof Url) {
+        continue;
+      }
+
+      if ($link['href']->isExternal()) {
```
22Alexandra, line 1742: We need to consider links under `europa.eu` domain ([reqs page](https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?pageId=550930772)) as internal so we shouldn't add the external icon.

europa eu

22Alexandra (changes requested):

22Alexandra (dismissed):

yenyasinn (approved):

22Alexandra (approved):

PR #1069, opened 2022-03-24
EWPP-2069: External link icon for patterns rendering links.

22Alexandra:


oe_theme.theme
```diff
@@ -213,9 +213,11 @@ function oe_theme_theme_suggestions_paragraph_alter(array &$suggestions, array $
 function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void {
   // Massage data to be compliant with ECL links block component data structure.
   foreach (Element::children($variables['content']['field_oe_links']) as $index) {
+    $url = $variables['content']['field_oe_links'][$index]['#url'];
     $variables['links'][] = [
       'label' => $variables['content']['field_oe_links'][$index]['#title'],
-      'url' => $variables['content']['field_oe_links'][$index]['#url'],
+      'url' => $url,
+      'is_external' => ($url->isExternal() && !str_contains($url->getUri(), 'europa.eu')),
```
nagyad: europa.eu could be present in another section of the url as well, I think it's better to parse the url and check for the path, so we are sure that the domain is what we check against.

oe_theme.theme
```diff
@@ -213,9 +214,11 @@ function oe_theme_theme_suggestions_paragraph_alter(array &$suggestions, array $
 function oe_theme_preprocess_paragraph__oe_links_block(array &$variables): void {
   // Massage data to be compliant with ECL links block component data structure.
   foreach (Element::children($variables['content']['field_oe_links']) as $index) {
+    $url = $variables['content']['field_oe_links'][$index]['#url'];
     $variables['links'][] = [
       'label' => $variables['content']['field_oe_links'][$index]['#title'],
-      'url' => $variables['content']['field_oe_links'][$index]['#url'],
+      'url' => $url,
+      'is_external' => ($url->isExternal() && !strpos(UrlHelper::parse($url->toString())['path'], 'europa.eu')),
```
nagyad: Why are we using is_external and external_link names? Can't we define only one of these in the following components?

oe_theme.theme
```diff
@@ -1415,11 +1426,17 @@ function oe_theme_preprocess_pattern_link(array &$variables): void {
...
   if ($variables['url'] instanceof Url) {
     $attributes = (array) $variables['url']->getOption('attributes');
+    if (isset($attributes['class'])) {
+      $variables['classes'] = implode(' ', $attributes['class']);
+      unset($attributes['class']);
+    }
+    foreach ($attributes as $name => $value) {
+      $variables['extra_attributes'][] = [
+        'name' => $name,
+        'value' => $value,
```
nagyad, line 1438: We might need to check if we add string here and not other value, to have valid html.

templates/patterns/link/pattern-link.html.twig
```diff
@@ -4,4 +4,24 @@
...
+{% if external_link %}
+  {% set _link = _link|merge({
+    icon: {
+      'path': ecl_icon_path,
+      'name': 'external',
+      'size': 's',
+    }
+  }) %}
+{% endif %}
+{% apply spaceless %}
```
nagyad: Do we need this here?

templates/patterns/link_block/pattern-link-block.html.twig
```diff
@@ -14,6 +14,15 @@
       type: 'standalone'
     }
   } %}
+  {% if link.is_external %}
+    {% set _link = _link|merge({
+      'icon': {
+        'path': ecl_icon_path,
+        'name': "external",
+        'size': "xs"
```
nagyad, line 22: Why are we using xs here and s other places?

templates/patterns/list_item/pattern-list-item--variant-date.html.twig
```diff
@@ -54,13 +54,23 @@
...
+              path: url,
             }
-          } only %}
+          } %}
+          {% if external_link %}
+            {% set _link = _link|merge({
+              icon: {
+                'path': ecl_icon_path,
+                'name': 'external',
+                'size': 's',
```
nagyad: This should be xs and at the other variant too, personal opinion. Do as you see fit.
image

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -461,6 +461,7 @@ public function testBanner(): void {
     $this->assertEquals('Banner', trim($crawler->filter('div.ecl-hero-banner__content div.ecl-hero-banner__title')->text()));
     $this->assertEquals('Description', trim($crawler->filter('div.ecl-hero-banner__content p.ecl-hero-banner__description')->text()));
     $this->assertCount(1, $crawler->filter('div.ecl-hero-banner__content a.ecl-link.ecl-link--cta.ecl-link--icon.ecl-link--icon-after'));
+    $this->assertEquals('<use xlink:href="/themes/custom/oe_theme/dist/ec/images/icons/sprites/icons.svg#corner-arrow"></use>', $crawler->filter('svg.ecl-icon.ecl-icon--xs.ecl-link__icon')->html());
```
nagyad, line 464: According to https://european-union.europa.eu/index_en the icon should not be present, no?

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -476,6 +477,10 @@ public function testBanner(): void {
 
     // Variant - image / Modifier - hero_left / Full width - No.
     $paragraph->get('field_oe_banner_type')->setValue('hero_left');
+    $paragraph->set('field_oe_link', [
+      'uri' => 'https://example.com',
+      'title' => 'Example',
+    ]);
```
nagyad, line 483: After this we should assert the icon is not present.

tests/src/Kernel/fixtures/patterns/organisation_teaser_rendering.yml
```diff
@@ -63,3 +63,10 @@
       a.ecl-link[href="mailto:me@example.com"]: 1
     equals:
       a.ecl-link svg.ecl-icon: '<use xlink:href="/themes/custom/oe_theme/dist/ec/images/icons/sprites/icons.svg#external"></use>'
+- fields:
+    contact_page_url: "https://european-union.europa.eu/index_en"
+  assertions:
+    count:
+      a.ecl-link[href="https://european-union.europa.eu/index_en"]: 1
+    equals:
+      a.ecl-link svg.ecl-icon: '<use xlink:href="/themes/custom/oe_theme/dist/ec/images/icons/sprites/icons.svg#corner-arrow"></use>'
```
nagyad, line 72: europa.eu domains should be considered internal, right?

nagyad (changes requested):

oe_theme.theme
22Alexandra: Some patterns have a `url` field and an `items` field. The `items` field also contains a link so in this cases we need `external_link` field for the main url and `is_external` for each item.

templates/patterns/link_block/pattern-link-block.html.twig
22Alexandra: We can adapt to either `xs` or `s`. Whichever fits better. Let me know if you think `s` should be used here.

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
22Alexandra: I'm asserting the internal icon, no? Which was there before.

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
22Alexandra: In fact, the external icon should replace now the internal one after updating the link.

tests/src/Kernel/fixtures/patterns/organisation_teaser_rendering.yml
22Alexandra: And isn't that what I'm asserting, the internal icon?

oe_theme.theme
22Alexandra: Why? The attributes should be set by a text editor and they are valid, no?

templates/patterns/link/pattern-link.html.twig
22Alexandra: Well, we usually use `spaceless` in the majority of our templates, is it wrong to have it?

templates/patterns/list_item/pattern-list-item--variant-date.html.twig
22Alexandra: Could we leave the icon size to the consideration of the PO when testing? We can adapt if you think it's needed but i would rather let him decide after testing which icon size fits better for each case. What do you think?

tests/src/Kernel/Paragraphs/ParagraphsTest.php
```diff
@@ -552,20 +560,36 @@ public function testContextualNavigation(): void {
...
+    $actual = $link3->text();
+    $this->assertEquals('Internal link under eu domain', trim($actual));
+    $actual = $link3->attr('href');
+    $this->assertEquals('http://ec.europa.eu/info', trim($actual));
+
+    $link4 = $crawler->filter('nav.ecl-contextual-navigation ul.ecl-contextual-navigation__list a.ecl-contextual-navigation__link')->eq(3);
+    $actual = $link4->text();
+    $this->assertEquals('Internal link', trim($actual));
+    $actual = $link4->attr('href');
+    $this->assertEquals('/', trim($actual));
```
nagyad: Here let's check the icon cannot be found

nagyad (changes requested):

nagyad (dismissed):

nagyad (dismissed):

oe_theme.theme
```diff
@@ -128,6 +129,7 @@ function oe_theme_preprocess_menu__main(array &$variables): void {
       'label' => $item['title'],
       'href' => $item['url'],
       'is_current' => $item['in_active_trail'],
+      'external' => ($item['url']->isExternal() && !strpos(UrlHelper::parse($item['url']->toString())['path'], 'europa.eu')),
```
imanoleguskiza: We need to centralize this logic, if we need to update/increase the domains that we consider "internal" it will be quite hard to do. On top of this, sites might want to define themselves what constitutes an "external" link

imanoleguskiza (changes requested):

imanoleguskiza (approved):

oe_theme.theme
nagyad: Very good idea.

nagyad (approved):

PR #1084, opened 2022-04-19
[EWPP-2106] EWPP-2109: Theme Carousel paragraph.

22Alexandra:


templates/patterns/carousel/pattern-carousel.html.twig
```diff
@@ -0,0 +1,34 @@
...
+    link: _link,
+    image: item.image|default(''),
+    variant: item.variant|default('primary'),
+    centered: item.centered is defined and item.centered == false ? false : true,
+  }]) %}
+  {% set _link = {} %}
+{% endfor %}
+{% include '@ecl-twig/carousel' with {
+  'items': _items,
+  'counter_label': counter_label|default('of'),
```
sergepavle:
```suggestion
  'counter_label': counter_label|default('of'|t),
```

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -922,4 +927,152 @@ public function testIframe(): void {
...
+      ->setEnabled('media', 'image', TRUE);
+    // Make the image field translatable.
+    $field_config = $this->container->get('entity_type.manager')
+      ->getStorage('field_config')
+      ->load('media.image.oe_media_image');
+    $field_config->set('translatable', TRUE)->save();
+    $this->container->get('router.builder')->rebuild();
+
+    // Create English files.
+    $en_file_1 = file_save_data(file_get_contents(\Drupal::service('extension.list.theme')->getPath('oe_theme') . '/tests/fixtures/example_1.jpeg'), 'public://example_1_en.jpeg');
```
sergepavle, line 945: file_save_data() already deprecated function. Should be used Use \Drupal\file\FileRepositoryInterface::writeData(). True for other places.

tests/src/Kernel/fixtures/rendering.yml
```diff
@@ -3243,3 +3243,125 @@
     equals:
       'div.ecl-list-illustration__item:nth-child(1) svg.ecl-icon': '<use xlink:href="/themes/custom/oe_theme/dist/ec/images/icons/sprites/icons.svg#growth"></use>'
       'div.ecl-list-illustration__item:nth-child(2) svg.ecl-icon': '<use xlink:href="/themes/custom/oe_theme/dist/ec/images/icons/sprites/icons.svg#budget"></use>'
+- array:
```
sergepavle, line 3246: Why cannot use single tests with all these items?
Also looks like we didn't test the 'counter_label' option in the pattern.

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
22Alexandra: It doesn't work on lowest versions: https://drone.fpfis.eu/openeuropa/oe_theme/7246/33 (try locally)

tests/src/Kernel/fixtures/rendering.yml
22Alexandra: I'll merge 2 of the test array but we still need to keep 2 for testing the default value for the `counter_label` and `full_width`

sergepavle (approved):

imanoleguskiza (approved):

PR #1089, opened 2022-04-26
OEL-1362: ImageValueObject doesn't inherit cache information from image file.

escuriola:
### Description

In oe_bootstrap_theme we updated the ImageValueObject. The idea is align both ImageValueObject class and tests.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/src/ValueObject/ImageValueObject.php
+++ b/src/ValueObject/ImageValueObject.php
@@ -123,7 +123,7 @@ class ImageValueObject extends ValueObjectBase implements ImageValueObjectInterf
    * @return $this
    */
   public static function fromImageItem(ImageItem $image_item): ValueObjectInterface {
-    $image_file = $image_item->get('entity')->getTarget();
+    $image_file = $image_item->get('entity')->getValue();
 
     $image_object = new static(
       file_create_url($image_file->get('uri')->getString()),
@@ -145,13 +145,13 @@ class ImageValueObject extends ValueObjectBase implements ImageValueObjectInterf
    *   The image style name.
    *
    * @return $this
-   *   A image value object instance.
+   *   An image value object instance.
    *
    * @throws \InvalidArgumentException
    *   Thrown when the image style is not found.
    */
   public static function fromStyledImageItem(ImageItem $image_item, string $style_name): ValueObjectInterface {
-    $image_file = $image_item->get('entity')->getTarget();
+    $image_file = $image_item->get('entity')->getValue();
 
     $style = \Drupal::entityTypeManager()->getStorage('image_style')->load($style_name);
     if (!$style) {
--- a/tests/src/Kernel/ValueObject/ImageValueObjectTest.php
+++ b/tests/src/Kernel/ValueObject/ImageValueObjectTest.php
@@ -96,7 +96,7 @@ class ImageValueObjectTest extends KernelTestBase {
     $this->assertEquals('This is an alternative title', $object->getAlt());
     $this->assertStringContainsString('/styles/main_style/public/example_1.jpg', $object->getSource());
 
-    // Test that all the cache tags have present and bubbled up.
+    // Test that all the cache tags are present and have bubbled up.
     $this->assertEqualsCanonicalizing([
       'config:image.style.main_style',
       'file:1',
@@ -116,7 +116,7 @@ class ImageValueObjectTest extends KernelTestBase {
     $this->assertEquals('This is an alternative title', $object->getAlt());
     $this->assertStringContainsString('/files/example_1.jpg', $object->getSource());
 
-    // Test that all the cache tags have present and bubbled up.
+    // Test that all the cache tags are present and have bubbled up.
     $this->assertEqualsCanonicalizing([
       'file:1',
     ], $object->getCacheTags());
--- a/tests/src/Kernel/ValueObject/ImageValueObjectTest.php
+++ b/tests/src/Kernel/ValueObject/ImageValueObjectTest.php
@@ -17,7 +17,7 @@ use Drupal\Tests\token\Kernel\KernelTestBase;
  *
  * @group batch2
  */
-class ImageTest extends KernelTestBase {
+class ImageValueObjectTest extends KernelTestBase {
 
   /**
    * {@inheritdoc}
@@ -28,6 +28,13 @@ class ImageTest extends KernelTestBase {
     'entity_test',
     'field',
   ];
+
+  /**
+   * Entity object.
+   *
+   * @var \Drupal\Core\Entity\EntityInterface
+   */
+  protected $entity;
 
   /**
    * {@inheritdoc}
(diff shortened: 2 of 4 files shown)


oe_theme.theme
```diff
@@ -813,10 +813,7 @@ function oe_theme_preprocess_paragraph__oe_list_item(array &$variables): void {
     /** @var \Drupal\file\Plugin\Field\FieldType\FileItem $image_item */
     $image_item = $paragraph->get('field_oe_image')->first();
     $file = $image_item->get('entity')->getValue();
-    $variables['image'] = [
-      'src' => file_url_transform_relative(file_create_url($file->getFileUri())),
-      'alt' => $image_item->get('alt')->getValue(),
-    ];
+    $variables['image'] = ImageValueObject::fromImageItem($image_item);
```
drishu: the pattern signature doesn't ask for an ImageValueObject, but I agree with the change

src/ValueObject/ImageValueObject.php
```diff
@@ -170,4 +171,22 @@ public static function fromStyledImageItem(ImageItem $image_item, string $style_
     return $image_object;
   }
 
+  /**
+   * Transforms the value object into an image render array.
+   *
+   * @return array
+   *   An image render array.
+   */
+  public function toRenderArray(): array {
```
drishu: if we don't use if what's the point ?

tests/src/Kernel/ValueObject/ImageValueObjectTest.php
```diff
@@ -56,44 +63,63 @@ protected function setUp(): void {
...
+    /** @var \Drupal\image\ImageStyleInterface $style */
+    $style = ImageStyle::create(['name' => 'main_style']);
+    $style->save();
+
+    $object = ImageValueObject::fromStyledImageItem($this->entity->get('field_image')->first(), $style->getName());
+    $this->assertEquals('This is a title', $object->getName());
+    $this->assertEquals('This is an alternative title', $object->getAlt());
     $this->assertStringContainsString('/styles/main_style/public/example_1.jpg', $object->getSource());
 
+    // Test that all the cache tags have present and bubbled up.
```
drishu:
```suggestion
    // Test that all the cache tags are present and have bubbled up.
```
same for testFromImageItem()

tests/src/Kernel/ValueObject/ImageValueObjectTest.php
```diff
@@ -56,44 +63,63 @@ protected function setUp(): void {
...
     $invalid_image_style = $this->randomMachineName();
     $this->expectExceptionObject(new \InvalidArgumentException(sprintf('Could not load image style with name "%s".', $invalid_image_style)));
-    ImageValueObject::fromStyledImageItem($entity->get('field_image')->first(), $invalid_image_style);
+    ImageValueObject::fromStyledImageItem($this->entity->get('field_image')->first(), $invalid_image_style);
+  }
+
+  /**
+   * Tests the ::fromImageItem method.
+   */
+  public function testFromImageItem() {
```
drishu, line 113: I think this one makes more sense, but then it makes the one from the unit test redundant

drishu (changes requested):

drishu (commented):
can we update also the PR title and have a better PR description, thx

src/ValueObject/ImageValueObject.php
donquixote: I agree, let's not add unused stuff now.
We can sync more if/when we create a shared package for this stuff.

tests/src/Kernel/ValueObject/ImageValueObjectTest.php
```diff
@@ -56,44 +63,63 @@ protected function setUp(): void {
...
-    // Create a test style.
-    /** @var \Drupal\image\ImageStyleInterface $style */
-    $style = ImageStyle::create(['name' => 'main_style']);
-    $style->save();
-
     // Create an image item.
-    $alt = $this->randomString();
-    $title = $this->randomString();
-    $entity = EntityTest::create([
+    $this->entity = EntityTest::create([
```
donquixote, line 74: To clarify: This reuses the entity create code, but it does not reuse the entity - setUp() is run again for each test method.

Another way to reuse the entity create code is to have a createEntity() method, and then use a local variable where we need it. At least this way we only run it for those test methods that need it (currently all, but still)

donquixote: Currently there is no big difference.
But if we follow this as a best practice, it gives us more flexibility: We can have parameters on these methods, we can call them only when we need them.

In general, I prefer local variables, params and return values over properties.

brummbar (changes requested):
Blocking the PR so that I review it.

templates/patterns/list_item/list_item.ui_patterns.yml
```diff
@@ -84,7 +84,7 @@ list_item:
         end: 1546250400
         timezone: "Europe/Brussels"
     image:
-      type: "array"
+      type: "ImageValueObject"
```
brummbar: We cannot do this, we are not in charge of the API and we are breaking theirs. Let's just fix the tests.

oe_theme.theme
brummbar: Is this a needed change? I don't think so.

tests/src/Unit/ValueObject/ImageValueObjectTest.php
```diff
@@ -67,7 +67,7 @@ protected function setUp(): void {
     $this->imageFile->get('uri')->willReturn($this->imageUri->reveal());
 
     $this->imageEntity = $this->prophesize(EntityReference::class);
-    $this->imageEntity->getTarget()->willReturn($this->imageFile->reveal());
+    $this->imageEntity->getValue()->willReturn($this->imageFile->reveal());
```
brummbar: We can drop all of the mocked stuff.

brummbar (changes requested):

brummbar:
I wrote to drop all the mocked parts, not just that line. Obviously now the PR is broken. 
I'm taking over. I'll also drop the commits that change and restore the pattern signature. These back-and-forth commits make longer the git blame process and have 0 value.

tests/src/Kernel/ValueObject/ImageValueObjectTest.php
drishu: I agree as long as we can keep the random strings, we could pass them as parameters to the create 
```php
   $alt = $this->randomString();
   $title = $this->randomString();
   $entity = $this->createEntity($title, $alt);
```

brummbar: There's no need to have random strings, so moving the code to create the entity is a bit useless in this case. I'm a fan of the approach, but a `$this->createEntity()` in this case has no advantages.

brummbar: We discussed and we want to make sure that the class returns the correct alt and title so we will add the randomisation.

donquixote: Earlier I talked with @drishu and we identified 3 ways of structuring this to be able to reuse the entity creation code.
This is a recurring "problem" that we also see in other tests, so it is a good idea to agree on some guidelines.

1. One big test method to test both static factories. The entity can be created inside this method. Personally I don't like it as much.
2. Create the entity in `setUp()`, store in object property, then have two separate test methods both using `$this->entity`.
3. Create the entity in a separate method, have two separate test methods that use `$entity = $this->createEntity()`.

Personally I prefer option 3, because I generally like local variables and tend to avoid object properties.
This also guarantees that the creation only happens for those test methods where we need it.
(We don't save anything by having it in `setUp()`, because that is called again for each test method.)

I don't really care for the random values. In an earlier talk with @brummbar I thought he was the one who wants them :) For me, literal values are fine, perhaps better.

For literal values, I find it a bit weird if we set them explicitly inside a `testSomething()` method, but then we have the same hard-coded values in a separate assertion method.

donquixote: But, the above are only weak preferences, no strong opinions.

brummbar (approved):

tests/src/Kernel/ValueObject/ImageValueObjectTest.php
drishu: I agreed that random doesn't bring much to this test so current implementation is ok

drishu (approved):

PR #1093, opened 2022-05-06
EWPP-2069: Use configuration to define internal domains.

imanoleguskiza:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_helper/config/install/oe_theme_helper.internal_domains.yml
```diff
@@ -0,0 +1 @@
+internal_domain: ''
```
imanoleguskiza, line 1: EWCMS should set the value to a regex that matches the constraints set by COMM. SOmething like: 
http(s)?:\/\/[a-z]*\.?europa.eu(\/.*)*

PR #1086, opened 2022-04-25
EWPP-2149: Fix collecting of headers for Content row paragraph with inpage navigation variant.

sergepavle:
…npage navigation variant.

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

PR #1094, opened 2022-05-07
EWPP-2149: Allow inpage navigation generation to support multiple instances.

brummbar:
This PR has a different approach than #1086.
The reasoning behind this rework are:

- The above PR forces to specify where the generated items need to be added. If we have a template has very special elements that are rendered and need to act as titles (e.g. `span.fw-bold`), we need to somehow fetch the ID generated in the inpage content row paragraph so that we can pass it to the element. The idea behind the `data-inpage-navigation-source-area` was to be able to declare any template as source area, without having to do any extra steps.
- This PR reworks the JS to be more similar to a standalone library/plugin. Aside from the name (which we cannot change if the users are referencing its old methods), all the ECL specific selectors are gone from `Drupal.eclInPageNavigation`.


js/inpage_navigation.js
```diff
@@ -19,63 +19,41 @@
...
-        // The :scope pseudo-class is needed to make sure that the selectors are applied inside the parent.
-        // @see https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll#user_notes
-        Array.prototype.forEach.call(area.querySelectorAll(':scope ' + selectors), function (element) {
-          element.setAttribute('data-inpage-navigation-source-element', '');
-        });
-      });
+      let containers = [].slice.call(document.querySelectorAll('.inpage-navigation-container'));
+      // If no specifically defined containers are present, we use the main content as one.
+      if (containers.length === 0) {
+        containers.push(document.querySelector('#main-content'));
```
sergepavle, line 30: As we plan to use js not related to some specific theme, It, probably, makes sense not to hardcode selector of the main content.

js/inpage_navigation.js
```diff
@@ -19,63 +19,41 @@
...
-          Drupal.eclInPageNavigation.handleEmptyInpageNavigation(block);
+          Drupal.eclInPageNavigation.handleEmptyInpageNavigation(nav);
           return;
         }
 
-        block.querySelector('ul').innerHTML = items_markup;
-        var instance = new ECL.InpageNavigation(block);
+        nav.querySelector('ul').innerHTML = items_markup;
+
+        const instance = new ECL.InpageNavigation(nav);
```
sergepavle, line 53: Does it make sense to move initialization in the separate method of `Drupal.eclInPageNavigation` class as we could expect that this script can be used outside of ECL.

sergepavle (dismissed):

js/inpage_navigation.js
brummbar: The idea was to make reusable the items generation, but most likely all the remaining Drupal-ish logic will stay in here.

js/inpage_navigation.js
brummbar: Same as above.

js/inpage_navigation.js
sergepavle: good for me.

sergepavle (approved):

22Alexandra (approved):

PR #1096, opened 2022-05-12
EWPP-2166: Fix access check for event programme items.

nagyad:


modules/oe_theme_content_event/src/Plugin/ExtraField/Display/ProgrammeExtraField.php
```diff
@@ -113,10 +113,13 @@ public function getLabel() {
...
   /**
    * {@inheritdoc}
+   *
+   * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+   * @SuppressWarnings(PHPMD.NPathComplexity)
    */
   public function viewElements(ContentEntityInterface $entity) {
     if ($entity->get('oe_event_programme')->isEmpty()) {
-      return;
+      return [];
```
22Alexandra: We shouldn't return empty array but use the `$isEmpty` flag of the `ExtraFieldDisplayFormattedBase`.

22Alexandra (changes requested):

22Alexandra (dismissed):

22Alexandra (approved):

upchuk (approved):

PR #1103, opened 2022-05-23
EWPP-2225: Project content type fixes.

nagyad:


modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
```diff
@@ -106,26 +106,31 @@ public function viewElements(ContentEntityInterface $entity) {
...
    * @param float $part
-   *   The percentage value.
+   *   The contribution part of the total value.
    *
-   * @return float
+   * @return string
    *   Percentage value.
    */
-  protected function getPercentage(float $total, float $part): float {
+  protected function getPercentage(float $total, float $part): string {
```
22Alexandra: We should keep the float type.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
```diff
@@ -106,26 +106,31 @@ public function viewElements(ContentEntityInterface $entity) {
...
 
     if ($total > 0 && $part > 0) {
-      $percentage = round(100 * $part / $total, 0);
+      $percentage = 100 * $part / $total;
     }
 
-    return $percentage;
+    // If our number is not round we will limit the decimals to one.
+    if (intval($percentage) != $percentage) {
+      return number_format($percentage, 1, '.', ' ');
```
22Alexandra: We shouldn't have to provide any separators for a percentage and I think the whole if statement is not needed.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
```diff
@@ -106,26 +106,31 @@ public function viewElements(ContentEntityInterface $entity) {
...
+    }
+
+    return (string) $percentage;
```
22Alexandra: By doing this:
```suggestion
    return round($percentage, 2);
```
We would get a more accurate percentage:
2 decimals

But we could also use only one decimal and get the same result we currently get with the if statement above that could be replaced by this:
```suggestion
    return round($percentage, 1);
```
1 decimal

tests/src/Kernel/ProjectRenderTest.php
```diff
@@ -172,7 +172,7 @@ public function testProjectTeaser(): void {
           'items' => [
             [
               'label' => 'Project locations',
-              'body' => '1000 Brussels, Belgium',
+              'body' => 'Belgium',
```
22Alexandra, line 175: Try adding more project locations so we make sure the rendering works as expected including the separator, not just that it renders only the country.

modules/oe_theme_content_project/oe_theme_content_project.post_update.php
```diff
@@ -54,3 +54,13 @@ function oe_theme_content_project_post_update_00002(): void {
   $display = $storage->createFromStorageRecord($display_values);
   $display->save();
 }
+
+/**
+ * Updates the budget field's 'thousand' separator to space.
+ */
+function oe_theme_content_project_post_update_00003(): void {
+  $view_display = EntityViewDisplay::load('node.oe_project.full');
```
22Alexandra: We should also update the teaser view mode address formatter, no?

22Alexandra (changes requested):

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
nagyad: The logic there is if the result is a number without decimals, we use that, if it has decimals, then we make it round with 1 decimal number to be more accurate.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
nagyad: round can't be applied in all cases, see my previous comment. That's why that logic was implemented.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
nagyad: We don't need float, we are not using it for anything.

tests/src/Kernel/ProjectRenderTest.php
nagyad: That is tested in the dedicated test for the fomatter. see AddressCountryInlineFormatterTest
Here we don't have to re-test it an iterate through this big test if you don't mind.

modules/oe_theme_content_project/oe_theme_content_project.post_update.php
nagyad: Yes, let's do that.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
nagyad: It will apply the decimal even if the final number is round, like 10% becomes 10.0% and this is not what we want.

tests/src/Kernel/ProjectRenderTest.php
22Alexandra: Fine.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
22Alexandra: Did you check if it's the case? Because it doesn't seems so:
1 decimal
no decimal

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
22Alexandra: Updating this as I suggested using `round()` method the returned value will be a float.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
22Alexandra: Please check the reply on the comment below.

modules/oe_theme_content_project/oe_theme_content_project.post_update.php
22Alexandra: And you didn't...

modules/oe_theme_content_project/oe_theme_content_project.post_update.php
```diff
@@ -54,3 +54,15 @@ function oe_theme_content_project_post_update_00002(): void {
...
+
+/**
+ * Updates the budget field's 'thousand' separator to space.
+ */
+function oe_theme_content_project_post_update_00003(): void {
+  foreach (['teaser', 'full'] as $display) {
+    $view_display = EntityViewDisplay::load("node.oe_project.$display");
+    $view_display->getComponent('oe_project_budget');
+    $component['settings']['thousand_separator'] = ' ';
+    $view_display->setComponent('oe_project_budget', $component)->save();
```
22Alexandra: This is not good.
We need to update this separately. 
`oe_project_budget` is hidden on teaser view mode so there is no point for updating the `thousand_separator` setting.

Address should show the Country separated by comma only on teaser view mode. 

Could you plase fix the update path?

22Alexandra (changes requested):

modules/oe_theme_content_project/oe_theme_content_project.post_update.php
nagyad: Sorry bad conflict resolve.

modules/oe_theme_content_project/oe_theme_content_project.post_update.php
```diff
@@ -54,3 +54,23 @@ function oe_theme_content_project_post_update_00002(): void {
...
+
+/**
+ * Update the formatter plugin for locations field.
+ */
+function oe_theme_content_project_post_update_00004(): void {
+  $view_display = EntityViewDisplay::load('node.oe_project.teaser');
+  $view_display->getComponent('oe_project_locations');
+  $component['type'] = 'oe_theme_helper_address_country_inline';
+  $view_display->setComponent('oe_project_locations', $component)->save();
+}
```
22Alexandra: By separately I only meant there shouldn't be any loop, not really splitting the update path because it doesn't make sense. The rest of the PR seems fine.

22Alexandra (changes requested):

22Alexandra (dismissed):

yenyasinn (dismissed):

yenyasinn (dismissed):

22Alexandra (approved):

yenyasinn (approved):

PR #1105, opened 2022-05-23
EWPP-2283: Integrate new ECL Page header component.

yenyasinn:
PR based on the branch with ECL 3.3.0.


behat.yml.dist
```diff
@@ -33,7 +33,7 @@ default:
               logo: "a.ecl-site-header-core__logo-link .ecl-site-header-core__logo-image, a.ecl-site-header-standardised__logo-link .ecl-site-header-standardised__logo-image"
               footer logo: "a.ecl-footer-core__logo-link .ecl-footer-core__logo-image-desktop, a.ecl-footer-standardised__logo-link .ecl-footer-standardised__logo-image-desktop"
               media container: ".ecl-media-container"
-              page header: ".ecl-page-header-core"
+              page header: ".ecl-page-header"
```
22Alexandra, line 36: On the `ec` component with `core` branding we have an additional class, so we should target the page header like this:
```suggestion
              page header: ".ecl-page-header, .ecl-page-header--negative"
```

behat.yml.dist
```diff
@@ -72,17 +72,14 @@ default:
         header with menu: ".ecl-site-header-core[data-ecl-has-menu], .ecl-site-header-standardised[data-ecl-has-menu]"
         language dialog: ".ecl-dialog"
         language switcher: ".ecl-language-list"
-        language page switcher: ".ecl-page-header-core .ecl-lang-select-page"
         navigation: ".ecl-menu"
         page: "body"
-        page header: ".ecl-page-header-core, .ecl-page-header-standardised"
+        page header: ".ecl-page-header"
```
22Alexandra, line 77: Same as above.

modules/oe_theme_helper/tests/src/Functional/PageHeaderBlockTest.php
```diff
@@ -66,20 +66,20 @@ public function testRendering(): void {
...
     $assert_session->elementsCount('css', '.ecl-page-header__intro', 0);
     $assert_session->elementsCount('css', '.ecl-meta--header .ecl-meta__item', 0);
 
     // Test another route.
     $this->drupalGet('/user/login');
-    $assert_session->elementsCount('css', '.ecl-page-header-core', 1);
-    $header = $this->getSession()->getPage()->find('css', '.ecl-page-header-core');
-    $this->assertEquals('Log in', trim($header->find('css', '.ecl-page-header-core__title')->getText()));
-    $assert_session->elementsCount('css', '.ecl-page-header-core__description', 0);
+    $assert_session->elementsCount('css', '.ecl-page-header', 1);
```
22Alexandra: The negative page header class should be present so please assert that.

package.json
```diff
@@ -23,6 +23,7 @@
     "@ecl/theme-eu": "3.3.0",
     "@ecl/twig-component-accordion": "3.3.0",
     "@ecl/twig-component-blockquote": "3.3.0",
+    "@ecl/twig-component-breadcrumb": "3.3.0",
```
22Alexandra: Do we need to add the new Breadcrumb component in this PR?

package.json
```diff
@@ -47,6 +48,7 @@
     "@ecl/twig-component-message": "3.3.0",
     "@ecl/twig-component-news-ticker": "3.3.0",
     "@ecl/twig-component-page-banner": "3.3.0",
+    "@ecl/twig-component-page-header": "3.3.0",
```
22Alexandra: Please remove the old components as we shouldn't use them anymore.

sass/style-ec.scss
```diff
@@ -4,11 +4,9 @@
 // TODO: The help form component is not available on ECL v3.
 //@use '@ecl/ec-component-form-help-block/ec-component-form-help-block';
 @use "./compositions/content_item";
-@use "./compositions/content_language_switcher";
```
22Alexandra, line 7: Why removing the language selector?

sass/style-eu.scss
```diff
@@ -7,13 +7,9 @@
 //@use '@ecl/eu-component-form-help-block/eu-component-form-help-block';
 @use '@ecl/theme-dev/maps/color';
 @use "./compositions/content_item";
-@use "./compositions/content_language_switcher" with (
-  $_core_color: map.get(theme.$color, 'blue-100'),
-);
```
22Alexandra, line 12: Why removing the language selector?

tests/src/Functional/ContentCallForProposalsRenderTest.php
```diff
@@ -75,7 +75,7 @@ public function testProposalRendering(): void {
     $this->drupalGet($node->toUrl());
 
     // Assert page header - metadata.
-    $page_header = $this->assertSession()->elementExists('css', '.ecl-page-header-core');
+    $page_header = $this->assertSession()->elementExists('css', '.ecl-page-header');
```
22Alexandra: Assert the negative class everywhere the core class should have been present.

22Alexandra (dismissed):

package.json
yenyasinn: Yes, we need - new page header component uses new breadcrumb component https://github.com/ec-europa/europa-component-library/blob/v3-dev/src/implementations/twig/components/page-header/page-header.html.twig

behat.yml.dist
yenyasinn: "page header" region is used to take the header. ".ecl-page-header-core, .ecl-page-header-standardised" means ".ecl-page-header-core" or ".ecl-page-header-standardised". Since we have the same class for both headers we don't need to separate them anymore.

behat.yml.dist
yenyasinn: It is a selector to find the header. 
In your example "".ecl-page-header, .ecl-page-header--negative" browser will always use ".ecl-page-header" since both headers have this selector. Second selector is redundant. 
The type of the header is tested in tests/src/Kernel/Patterns/PageHeaderRenderingTest.php.

package.json
yenyasinn: Good point

sass/style-ec.scss
yenyasinn: Styles from this files aren't used anymore. Please check https://github.com/openeuropa/oe_theme/pull/1105/files#diff-2ff0353a221baabf5fe9662567a9de59fc9b3304137bb5b219a0cbd7ef25fbf3.
Language switcher isn't included in the page header in ECL v.3

sass/style-eu.scss
yenyasinn: Styles from this files aren't used anymore. Please check https://github.com/openeuropa/oe_theme/pull/1105/files#diff-2ff0353a221baabf5fe9662567a9de59fc9b3304137bb5b219a0cbd7ef25fbf3.
Language switcher isn't included in the page header in ECL v.3

modules/oe_theme_helper/tests/src/Functional/PageHeaderBlockTest.php
yenyasinn: Ok

tests/src/Functional/ContentCallForProposalsRenderTest.php
yenyasinn: The type of the header is tested in tests/src/Kernel/Patterns/PageHeaderRenderingTest.php.
"Negative" class is also can be tested in page header related test `modules/oe_theme_helper/tests/src/Functional/PageHeaderBlockTest.php`.

In this test we just ensure that header exists. It doesn't matter for the page which exactly type of the header exists.

sass/style-ec.scss
yenyasinn: Alexandra insists that this file is required but I don't understand why. Could someone else check this point.

22Alexandra (approved):

nagyad (approved):

PR #1119, opened 2022-06-13
EPIC-EWPP-2022-External-links

22Alexandra:


modules/oe_theme_helper/src/ExternalLinks.php
```diff
@@ -0,0 +1,57 @@
...
+   *   The configuration factory.
+   */
+  public function __construct(ConfigFactoryInterface $config_factory) {
+    $this->configFactory = $config_factory;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isExternalLink($url): bool {
```
ademarco: We should cover this with a dedicated test, where we throw at it all sorts of different values, of different types, and check the output.

modules/oe_theme_helper/src/ExternalLinks.php
```diff
@@ -0,0 +1,57 @@
...
+    else {
+      $external = UrlHelper::isExternal($url);
+      $path = UrlHelper::parse($url)['path'];
+    }
+    if (!$external) {
+      return $external;
+    }
+
+    // If it's an external link, make sure its domain is not internal.
+    $internal_domain_expression = $this->configFactory->get('oe_theme_helper.internal_domains')->get('internal_domain');
```
ademarco: We can get the expression in the constructor, as we really don't need the config factory for anything else in this service.

Let's have a property called $internalDomainExpression, and we simply fill it in in the constructor.

ademarco (dismissed):

imanoleguskiza (approved):

nagyad (approved):

PR #1125, opened 2022-06-17
EWPP-2022: Provide test coverage for the external links service.

22Alexandra:


modules/oe_theme_helper/tests/src/Kernel/ExternalLinksTest.php
```diff
@@ -0,0 +1,37 @@
...
+ * Provides test coverage for the ExternalLinks service.
+ *
+ * @group batch1
+ */
+class ExternalLinksTest extends AbstractKernelTestBase {
+
+  /**
+   * Covers isExternalLink method.
+   */
+  public function testIsExternalLink(): void {
```
ademarco, line 20: Let's also add some tests asserting to `true`, if we don't do this we might as well have a method that always returns false, for any combinations of inputs.

ademarco (changes requested):

ademarco (approved):

nagyad (approved):

PR #1159, opened 2022-08-25
EWPP-2539: Fix hreflang attribute in file_translation and file_teaser patterns.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra (changes requested):
We should fix file_teaser pattern as well because now it still shows pt-pt:

file teaser

22Alexandra (dismissed):

22Alexandra (dismissed):

PR #1174, opened 2022-10-04
EWPP-2646: Fix publication thumbnail translation render.

nagyad:


sergepavle (approved):

sergepavle (dismissed):

tests/src/Functional/ContentPublicationRenderTest.php
```diff
@@ -496,4 +496,87 @@ public function testPublicationRendering(): void {
     $field_list_assert->assertPattern($details_expected_values, $content_items[0]->getHtml());
   }
 
+  /**
+   * Tests that the Publication thumbnail media renders the translated media.
+   */
+  public function testPublicationThumbnailMediaTranslation(): void {
+    // Make publication node and image media translatable.
+    \Drupal::service('content_translation.manager')->setEnabled('node', 'oe_publication', TRUE);
```
22Alexandra: The bundle is already translatable because this is done in the `setUp()` method.

tests/src/Functional/ContentPublicationRenderTest.php
```diff
@@ -496,4 +496,87 @@ public function testPublicationRendering(): void {
     $field_list_assert->assertPattern($details_expected_values, $content_items[0]->getHtml());
   }
 
+  /**
+   * Tests that the Publication thumbnail media renders the translated media.
+   */
+  public function testPublicationThumbnailMediaTranslation(): void {
```
22Alexandra, line 506: Why can't we test this in the previous test method? The title field is set to be translatable in the `setUp()` and then the translation of the node is being tested. We can just do all the configuration needed in the `setUp()` and then provide a `Thumbnail` for the publication node so we can test this scenario too.

22Alexandra (changes requested):

tests/src/Functional/ContentPublicationRenderTest.php
nagyad: Because that method is over 400 lines. I don't like shovelling everything under one test case. Here we test a specific scenario and this scenario is separate for event as well.

22Alexandra (dismissed):

sergepavle (dismissed):

22Alexandra (approved):

hernani (approved):

sergepavle (approved):

PR #1191, opened 2022-10-18
EWPP-2506: Add classes to identify the last item per column.

22Alexandra:


sergepavle (dismissed):

nagyad (dismissed):

sergepavle (dismissed):

templates/patterns/list_item/pattern-list-item.html.twig
```diff
@@ -153,9 +153,9 @@
         }]) %}
       {% endif %}
     {% endfor %}
-    {# Always use the horizontal variant of ECL Description list component for this pattern. #}
+    {# Always use the vertical variant of ECL Description list component for this pattern. #}
```
imanoleguskiza: This comment is not very correct I would say. Shouldn't it be something like "Use the horizontal variant of ECL Description list component by default"?

imanoleguskiza (dismissed):
Currently we are not using the tags display in link lists, we need to make sure we keep tthe same display as before.

sergepavle (approved):

yenyasinn (approved):

nagyad (approved):

PR #1192, opened 2022-10-20
EWPP-2718: Update ECL to 3.5.0 version.

nagyad:


imanoleguskiza (dismissed):

22Alexandra (changes requested):
We need to rename the patch files.

22Alexandra (dismissed):

22Alexandra (dismissed):

composer.json
```diff
@@ -19,6 +19,7 @@
         "behat/behat": "^3.10",
         "behat/mink-extension": "^2.3.1",
         "composer/installers": "~1.5",
+        "consolidation/annotated-command": "4.5.6",
```
upchuk: Is this still needed?

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
```diff
@@ -123,7 +123,7 @@ public function testEclMultiSelect(): void {
 
     // Assert all options are visible.
     $options = [
-      'Select all',
+      'Select all (4)',
```
upchuk, line 126: Why is this happening? We cannot just change the test cause drone fails. There is a reason we have tests and that is to check if we break stuff. Not that if we break stuff, we just adapt the test. So please confirm what this is.

upchuk (changes requested):
Please

upchuk (approved):

22Alexandra (approved):

PR #1194, opened 2022-10-25
EWPP-2706: Make sure featured media title is not printed when field is empty.

nagyad:


modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DescriptionExtraField.php
```diff
@@ -88,13 +88,16 @@ public function getLabel() {
    * {@inheritdoc}
    */
   public function viewElements(ContentEntityInterface $entity) {
+    // If we have no renderable text wee don't need a title.
+    $text = $this->getRenderableText($entity);
+    $title = !empty($text[0]['#text']) ? $this->getRenderableTitle($entity) : '';
     // Display event description using "text_featured_media" pattern.
```
sergepavle: Probably we have to hide if all fields are empty. There missed Title but the image is available:
image

sergepavle (changes requested):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DescriptionExtraField.php
```diff
@@ -88,13 +88,16 @@ public function getLabel() {
    * {@inheritdoc}
    */
   public function viewElements(ContentEntityInterface $entity) {
+    // If we have no renderable text and image wee don't need a title.
+    $text = $this->getRenderableText($entity);
+    $title = !empty($text[0]['#text']) || !$entity->get('oe_event_featured_media')->isEmpty() ? $this->getRenderableTitle($entity) : '';
```
sergepavle: As I can see in \Drupal\oe_theme_content_event\Plugin\ExtraField\Display\DescriptionExtraField::addFeaturedMediaThumbnail implementation we have to check not only if the field is empty but also if the media still exists and the user still has access to media. Maybe it makes sense to implement a separate method for checking if the image can be exposed.

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -496,6 +496,9 @@ public function testEventRendering(): void {
     $this->assertSession()->linkNotExistsExact('Main link for media items');
     $this->assertSession()->pageTextNotContains('More media links');
 
+    // Assert "Description" title is not rendered unless there is a body text.
+    $this->assertSession()->pageTextNotContains('Description');
+
     // Assert "Full text", "Featured media", "Featured media legend" fields
```
sergepavle: Probably tests should be slightly updated because the following behavior is deprecated.

sergepavle (changes requested):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DescriptionExtraField.php
```diff
@@ -92,13 +92,13 @@ public function viewElements(ContentEntityInterface $entity) {
     $build = [
       '#type' => 'pattern',
       '#id' => 'text_featured_media',
-      '#fields' => [
-        'title' => $this->getRenderableTitle($entity),
-        'text' => $this->getRenderableText($entity),
-      ],
     ];
-
+    // If we have no renderable text and image wee don't need a title.
```
sergepavle:
```suggestion
    // If we have no renderable text and image we don't need a title.
```

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -496,14 +496,44 @@ public function testEventRendering(): void {
     $this->assertSession()->linkNotExistsExact('Main link for media items');
     $this->assertSession()->pageTextNotContains('More media links');
 
-    // Assert "Full text", "Featured media", "Featured media legend" fields
-    // (these fields have to be filled all together).
+    // Assert "Description" title is not rendered unless there is a body text.
+    $this->assertSession()->pageTextNotContains('Description');
+
+    // Assert "Full text", "Featured media", "Featured media legend" fields..
```
sergepavle:
```suggestion
    // Assert "Full text", "Featured media", "Featured media legend" fields.
```

sergepavle (dismissed):

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -496,14 +496,44 @@ public function testEventRendering(): void {
...
+
+    $description_content = $this->assertSession()->elementExists('css', 'article > div > div:nth-child(3)');
+    $text_featured = new TextFeaturedMediaAssert();
+    $text_featured_expected_values = [
+      'title' => 'Description',
+      'text' => 'Event full text',
+    ];
+    $text_featured->assertPattern($text_featured_expected_values, $description_content->getHtml());
+
+    $node->set('body', '');
```
22Alexandra: Please adapt the test coverage without the changes in oe_content because having a media without description will break the layout and that's fixed in EWPP-2714. This ticket should only fix the issue regarding having a heading for the description field group when no field is filled in.
So, only the first assertion is needed as a change to make sure that without having any of the fields we don't render a text with featured media pattern.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DescriptionExtraField.php
```diff
@@ -92,13 +92,13 @@ public function viewElements(ContentEntityInterface $entity) {
...
     $this->addFeaturedMediaThumbnail($build, $entity);
+    $text = $this->getRenderableText($entity);
+    $title = !empty($text[0]['#text']) || !empty($build['#fields']['image']) ? $this->getRenderableTitle($entity) : '';
+    $build['#fields']['title'] = $title;
+    $build['#fields']['text'] = $text;
```
22Alexandra, line 112: This is not good. When there is no media or description we're rendering an empty pattern with cache tags.
text field

If we don't have a media or text we should simply return an empty array instead of rendering a pattern template that's empty.
empty markup

22Alexandra (changes requested):

22Alexandra (dismissed):

sergepavle (dismissed):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DescriptionExtraField.php
```diff
@@ -88,16 +88,25 @@ public function getLabel() {
...
       '#type' => 'pattern',
       '#id' => 'text_featured_media',
-      '#fields' => [
-        'title' => $this->getRenderableTitle($entity),
-        'text' => $this->getRenderableText($entity),
-      ],
     ];
+    // If we have no renderable text and image we don't need a title.
+    $text = $this->getRenderableText($entity);
+    $this->addFeaturedMediaThumbnail($build, $entity);
```
upchuk, line 98: You end up calling this twice..

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DescriptionExtraField.php
```diff
@@ -88,16 +88,25 @@ public function getLabel() {
...
+    $title = !empty($text[0]['#text']) || isset($build['#fields']['image']) ? $this->getRenderableTitle($entity) : '';
+
+    // If we don't have a title we do not render anything because there is
+    // no text and no image.
+    if (empty($title)) {
+      return [];
```
upchuk: `getRenderableText()` and `getRenderableTitle()` add cache tags. So here you essentially kill all those cache tags.

upchuk (changes requested):

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DescriptionExtraField.php
```diff
@@ -88,17 +88,23 @@ public function getLabel() {
...
-
+    // If we have no renderable text and image we don't need a title.
+    $text = $this->getRenderableText($entity);
     $this->addFeaturedMediaThumbnail($build, $entity);
+    $title = !empty($text[0]['#text']) || isset($build['#fields']['image']) ? $this->getRenderableTitle($entity) : '';
+
+    // If we don't have a title we do not render anything because there is
+    // no text and no image.
+    if (empty($title)) {
+      return $build;
```
upchuk, line 108: I wrote in the original review: 

```
getRenderableText() and getRenderableTitle() add cache tags. So here you essentially kill all those cache tags.
```
So you now return $build but without those cache tags from `getRenderableText` and `getRenderableTitle`

upchuk (changes requested):

22Alexandra (approved):

sergepavle (approved):

upchuk (approved):

PR #1210, opened 2022-11-14
EWPP-2734: Fix WysiwygTableTest with removing workaround.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/tests/src/Behat/MinkContext.php
+++ b/tests/src/Behat/MinkContext.php
@@ -190,6 +190,9 @@ class MinkContext extends DrupalExtensionMinkContext {
       throw new \InvalidArgumentException(sprintf('Could not not find link: "%s"', $link));
     }
     $link->mouseOver();
+    // Add wait on showing menu sub-items implemented in ECL CSS:
+    // https://github.com/ec-europa/europa-component-library/blob/v3-dev/src/implementations/vanilla/components/menu/_menu.scss#L850.
+    $this->getSession()->wait(1000);
   }
 
 }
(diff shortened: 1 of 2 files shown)


tests/src/Behat/MinkContext.php
```diff
@@ -190,6 +190,8 @@ public function iHoverOverTheLink($link): void {
       throw new \InvalidArgumentException(sprintf('Could not not find link: "%s"', $link));
     }
     $link->mouseOver();
+    // Add wait for allowing possible delay on showing menu sub-items.
+    $this->getSession()->wait(1000);
```
22Alexandra, line 195: The scenario runs locally without the wait so maybe we should not add it because of drone and see why it fails in drone.

22Alexandra (changes requested):

tests/src/Behat/MinkContext.php
sergepavle: With using local docker I also can reproduce failure during the execution of behat tests `Scenario: The demo site navigation menu features dropdown menus`.
It is explicitly implemented delay on the show/hide of menu sub-items.

22Alexandra (dismissed):

yenyasinn (dismissed):

22Alexandra (approved):

yenyasinn (approved):

PR #1220, opened 2022-12-02
EWPP-2861: Add publication date field to file translation pattern.

22Alexandra:


nagyad (dismissed):

sergepavle (commented):
Should the file pattern be with meta (publication date)? As I understand from https://ec.europa.eu/component-library/ec/components/file/code/ such a file pattern is connected with the files component without translation. Is it because this pattern is not used?

sergepavle (dismissed):

sergepavle (approved):

nagyad (approved):

PR #1240, opened 2023-02-08
EWPP-3069: Update to ECL 3.7.0.

22Alexandra:


oe_theme.theme
```diff
@@ -2106,7 +2106,7 @@ function oe_theme_preprocess_paragraph__oe_carousel(array &$variables): void {
       'centered' => FALSE,
       // The paragraph has a required image field, so we need to set a default
       // image variant in order to render the image.
-      'variant' => 'image-gradient',
+      'variant' => 'text-highlight',
```
sergepavle, line 2111: As I understand we don't have any more 'gradient' variant for banner component but is it confirmed replacing to highlight variant (why not overlay for instance)?

patches/@ecl/twig-component-file+3.7.0.patch
```diff
@@ -0,0 +1,265 @@
...
+   </div>
+-{% if _translation is not empty %}
+-  {% include '@ecl/file/file-translations.html.twig' with _translation|merge({
+-    variant: _variant
+-  }) only %}
+-{% endif %}
+-</div>
+ 
+-{% endapply %}
++  {% endapply %}
```
sergepavle: It would be nice to leave only real changes without change only in indentation otherwise it is hard to review and maintain this patch in the future.

templates/patterns/carousel/carousel.ui_patterns.yml
```diff
@@ -5,22 +5,22 @@ carousel:
...
+          variant: "text-box"
           centered: false
           credit: "© Copyright first item"
         - title: "Duis vitae pulvinar turpis"
           description: "Integer quis lorem tellus. Nullam sollicitudin suscipit diam, ac blandit ipsum tempor consectetur"
           url: "https://example.com"
           url_text: "CTA"
           image: "https://placeimg.com/640/480/arch"
-          variant: "image-gradient"
+          variant: "text-highlight"
```
sergepavle, line 23: It can be a tricky moment for other non-EWPP sites if they will try to use not existing anymore variants.
Does it make sense somehow deprecate variants (and in a nutshell show different variant)?

sergepavle (commented):
I've left a few comments for consideration.

oe_theme.theme
22Alexandra: I replied in the ticket.

templates/patterns/carousel/carousel.ui_patterns.yml
22Alexandra: I agree but we're rendering the Carousel paragraph using the ECL Carousel component directly. I mentioned in the Carousel pattern description that the items have to be compatible with the ECL Banner component so non-EWPP sites can check that.

For the Banner paragraph there is no issue for now because the deprecated components i.e Page and Hero Banners can still be used and for the Carousel paragraph, the variant cannot be changed and we set a default value.

We will deal with the new Banner component and variants changes in a follow up.

patches/@ecl/twig-component-file+3.7.0.patch
22Alexandra: Fixed.

nagyad (dismissed):

sergepavle (approved):

upchuk (approved):
For adam

PR #1264, opened 2023-03-29
EWPP-2735: Row Header in Tables - Inconsistent text alignment.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
```diff
@@ -46,6 +46,15 @@ public function process($text, $langcode) {
         $table->removeAttribute('data-striped');
       }
 
+      // Processing of the table with any cell spans over multiple columns or
+      // rows produce visual issues in mobile/tablet view. For avoiding such
+      // problems we can use 'simple display' instead of ECL enhanced one for
+      // mobile/tablet.
+      $span_cells = $xpath->query('.//*[self::th or self::td][(@colspan and @colspan > 1) or (@rowspan and @rowspan > 1)]', $table);
```
imanoleguskiza: I don't think we need this fix, my tests didn't any issue with the mobile tables. In fact, adding the simple table display made the tables not adjust to the mobile view and to be much wider than the rest of the contents of the page.

imanoleguskiza (changes requested):

modules/oe_theme_helper/src/Plugin/Filter/FilterEclTable.php
sergepavle: Yes indeed. Mobile/tablet view looks as it should be. We may keep responsive behavior.

imanoleguskiza (approved):

22Alexandra (approved):

PR #1270, opened 2023-05-11
EWPP-3296: Theme the Media paragraph.

22Alexandra:


oe_theme.theme
```diff
@@ -2204,3 +2204,91 @@ function oe_theme_preprocess_oe_theme_helper_in_page_navigation__node__oe_person
...
+function oe_theme_preprocess_paragraph__oe_av_media(array &$variables): void {
+  /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
+  $paragraph = $variables['paragraph'];
+
+  // Bail out if there is no media present.
+  if ($paragraph->get('field_oe_media')->isEmpty()) {
+    return;
+  }
+
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
```
nagyad, line 2220: Why to create this here if you bail out after when the media entity is not available anymore in line 2226?

oe_theme.theme
```diff
@@ -2204,3 +2204,91 @@ function oe_theme_preprocess_oe_theme_helper_in_page_navigation__node__oe_person
...
+
+  // If it's an image media, render it and assign it to the image variable.
+  if ($source instanceof MediaAvPortalPhotoSource || $source instanceof Image) {
+    $thumbnail = $media->get('thumbnail')->first();
+    $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_medium_no_crop');
+    $cacheability->applyTo($variables);
+    return;
+  }
+
+  $cacheability->applyTo($variables);
```
nagyad: You already apply the cache of the $variables in line 2275 and after that you don't add anything to it, so this is not needed.

oe_theme.theme
```diff
@@ -2204,3 +2204,91 @@ function oe_theme_preprocess_oe_theme_helper_in_page_navigation__node__oe_person
...
+      $variables['ratio'] = str_replace('_', '-', $ratio);
+    }
+    return;
+  }
+
+  // If it's an image media, render it and assign it to the image variable.
+  if ($source instanceof MediaAvPortalPhotoSource || $source instanceof Image) {
+    $thumbnail = $media->get('thumbnail')->first();
+    $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_theme_medium_no_crop');
+    $cacheability->applyTo($variables);
```
nagyad: You already apply the cache of the $variables in line 2275 and after that you don't add anything to it, so this is not needed.

nagyad (changes requested):

oe_theme.theme
22Alexandra: It's created after making sure a media item exists. Why would it be created before?

oe_theme.theme
22Alexandra: Not really, that's for video media items. After we finish preprocessing the video, we return on 2282.

oe_theme.theme
22Alexandra: Same as below, on 2275 we apply the cache for the video item and then we return. 
Here we apply the cache that contains only the translated media and it's access which means we don't have a video or an image anymore.

nagyad (dismissed):

oe_theme.theme
```diff
@@ -2204,3 +2204,88 @@ function oe_theme_preprocess_oe_theme_helper_in_page_navigation__node__oe_person
...
+    return;
+  }
+
+  // Get the media source.
+  $source = $media->getSource();
+
+  if ($source instanceof MediaAvPortalVideoSource || $source instanceof OEmbed || $source instanceof Iframe) {
+    if ($source instanceof MediaAvPortalVideoSource) {
+      // Default video aspect ratio is set to 16:9 for AV Portal Video.
+      $variables['ratio'] = '16:9';
```
imanoleguskiza: This seems to be messing up with the height of avportal v ideos

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

nagyad (dismissed):

composer.json
```diff
@@ -50,7 +50,7 @@
         "openeuropa/oe_corporate_countries": "^2.0.0-alpha8",
         "openeuropa/oe_media": "^1.23.1",
         "openeuropa/oe_multilingual": "^1.13",
-        "openeuropa/oe_paragraphs": "^1.19",
+        "openeuropa/oe_paragraphs": "dev-EWPP-3295",
```
nagyad: Needs merge first.

nagyad (changes requested):

nagyad (approved):

hernani (approved):

PR #1275, opened 2023-05-28
EWPP-3324: Add new fields to gallery pattern.

22Alexandra:


templates/patterns/gallery/gallery.ui_patterns.yml
```diff
@@ -109,3 +109,31 @@ gallery:
...
+      description: "Disables the overlay functionality."
+      preview: true
+      escape: false
+    expandable:
+      type: "boolean"
+      label: "Expandable"
+      description: "Whether the 'See all' link is shown or all items are shown."
+      preview: true
+      escape: false
+    url:
```
upchuk: Is this how we handle links with patterns? `url` and `label` as separate fields? Shouldn't it be a single array field with label and url? I'm asking as I don't know well. For sure, `label` and `url` are not good names for something that is in fact the "Footer URL"

upchuk (changes requested):

templates/patterns/gallery/gallery.ui_patterns.yml
22Alexandra: Yes, that's how we use links because we need to pass a Link object to the ECL component so we build that in the pattern's template.

upchuk: And what if we need to have 2 links that do 2 different things? How do we handle that? In any case `url` and `label` are not good names. Should be prefixed with `footer_link_*` no?

nagyad (dismissed):

nagyad (dismissed):

nagyad (approved):

upchuk (approved):

PR #1277, opened 2023-06-01
EWPP-3315: Remove projecy budget fallback logic.

nagyad:


modules/oe_theme_content_project/src/Plugin/ExtraField/Display/OverallBudgetExtraField.php
```diff
@@ -74,7 +74,7 @@ public function getLabel() {
    * {@inheritdoc}
    */
   public function viewElements(ContentEntityInterface $entity) {
-    if ($entity->get('oe_project_budget')->isEmpty() && $entity->get('oe_project_eu_budget')->isEmpty()) {
+    if ($entity->get('oe_project_eu_budget')->isEmpty()) {
```
22Alexandra, line 77: We should not return empty arrays but use the isEmpty property.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
```diff
@@ -89,24 +89,19 @@ public function viewElements(ContentEntityInterface $entity) {
...
     // Return only EU contribution if budget is empty.
-    if ($entity->get('oe_project_budget')->isEmpty() && $entity->get('oe_project_eu_budget')->isEmpty()) {
+    if ($entity->get('oe_project_eu_budget')->isEmpty()) {
       return $build;
     }
 
     // Compute budget percentage field value.
-    $budget = $entity->get('oe_project_eu_budget')->isEmpty() ? $entity->get('oe_project_budget_eu')->value : $entity->get('oe_project_eu_budget')->value;
+    $budget = $entity->get('oe_project_eu_budget')->value;
+    $budget_eu = $entity->get('oe_project_eu_contrib')->value;
```
22Alexandra: We're still rendering 0% contribution which I don't think is needed.

22Alexandra (changes requested):

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/OverallBudgetExtraField.php
22Alexandra: Ok this is good now but we still need to `return;` when there is nothing else to do.

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
```diff
@@ -74,8 +74,8 @@ public function getLabel() {
    * {@inheritdoc}
    */
   public function viewElements(ContentEntityInterface $entity) {
-    if ($entity->get('oe_project_budget_eu')->isEmpty() && $entity->get('oe_project_eu_contrib')->isEmpty()) {
-      return [];
+    if ($entity->get('oe_project_eu_budget')->isEmpty() && $entity->get('oe_project_eu_contrib')->isEmpty()) {
+      $this->isEmpty = TRUE;
```
22Alexandra, line 78: Same as above.

22Alexandra (changes requested):

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/OverallBudgetExtraField.php
22Alexandra: Nvm.

22Alexandra (dismissed):

modules/oe_theme_content_project/src/Plugin/ExtraField/Display/PercentageExtraField.php
```diff
@@ -89,28 +89,23 @@ public function viewElements(ContentEntityInterface $entity) {
...
       return $build;
     }
 
-    // Compute budget percentage field value.
-    $budget = $entity->get('oe_project_eu_budget')->isEmpty() ? $entity->get('oe_project_budget_eu')->value : $entity->get('oe_project_eu_budget')->value;
-    $percentage = $this->getPercentage((float) $budget, (float) $budget_eu);
-    $build[] = [
-      '#markup' => '<div class="ecl-u-mt-m">' . $this->t("@percentage% of the overall budget", ["@percentage" => $percentage]) . '</div>',
-    ];
+    if (!$entity->get('oe_project_eu_contrib')->isEmpty() && !$entity->get('oe_project_eu_contrib')->isEmpty()) {
```
sergepavle: Repeated condition

sergepavle (changes requested):

sergepavle (approved):

22Alexandra (approved):

22Alexandra (approved):

PR #1324, opened 2023-09-22
EWPP-3633: Add support for CKEditor 5

brummbar:


modules/oe_theme_helper/tests/src/FunctionalJavascript/Ckeditor5TableTest.php
```diff
@@ -0,0 +1,435 @@
...
+  public function setUp(): void {
+    parent::setUp();
+
+    FilterFormat::create([
+      'format' => 'test_format',
+      'name' => 'Test format',
+      'weight' => 1,
+      'roles' => [RoleInterface::AUTHENTICATED_ID],
+      'filters' => [
+        'filter_ecl_table' => [
```
sergepavle, line 63: Should we add test coverage to assert that plugins are available if enabled **filter_ecl_table** and table is available in the toolbar?

modules/oe_theme_helper/oe_theme_helper.ckeditor5.yml
```diff
@@ -0,0 +1,50 @@
...
+        contentToolbar: [toggleTableSimple]
+  drupal:
+    label: Table simple
+    library: oe_theme_helper/internal.oe_theme_helper.ckeditor5.table_simple
+    elements:
+      - <table data-simple>
+    conditions:
+      plugins:
+        - ckeditor5_table
+      filter: filter_ecl_table
```
sergepavle: For some reason, the condition doesn't work. It looks like conditions work with OR logic.

brummbar: I found out that you actually need a specific order of conditions to apply them correctly.
Whenever a "plugins" condition is present, the check bails out and doesn't evaluate the other condition types.
This because the plugins can be evaluated only when everything else has been evaluated.
But using the correct order (moving plugins always below) it will apply them correctly.

modules/oe_theme_helper/tests/src/FunctionalJavascript/Ckeditor5TableTest.php
brummbar: Absolutely.

sergepavle (dismissed):

modules/oe_theme_helper/webpack.config.js
```diff
@@ -0,0 +1,66 @@
...
+function getDirectories(srcpath) {
+  return fs
+    .readdirSync(srcpath)
+    .filter((item) => fs.statSync(path.join(srcpath, item)).isDirectory());
+}
+
+module.exports = [];
+// Loop through every subdirectory in src, each a different plugin, and build
+// each one in ./build.
+getDirectories('./js/ckeditor5_plugins').forEach((dir) => {
```
nagyad, line 16: Shouldn't we do this in a try-catch if readdirSync throws an error?

modules/oe_theme_helper/webpack.config.js
```diff
@@ -0,0 +1,66 @@
...
+            },
+          },
+          test: /\.js(\?.*)?$/i,
+          extractComments: false,
+        }),
+      ],
+      moduleIds: 'named',
+    },
+    entry: {
+      path: path.resolve(
```
nagyad, line 35: I think `path:` is not needed there, since webpack entry takes the path as a string or an array of strings. You can just simply add path.resolve(....

modules/oe_theme_helper/webpack.config.js
```diff
@@ -0,0 +1,66 @@
+const path = require('path');
+const fs = require('fs');
+const webpack = require('webpack');
+const { styles, builds } = require('@ckeditor/ckeditor5-dev-utils');
+const TerserPlugin = require('terser-webpack-plugin');
+
+function getDirectories(srcpath) {
```
nagyad, line 7: This function should have an annotation describing its purpose.

brummbar: This file is the same as the webpack.config.js present in core, with modifications provided by the ckeditor5_dev module (maintained by core people). I would not introduce changes to it, for ease of updates in the future and consistency. This is valid for the points above too

brummbar: For reference: https://git.drupalcode.org/project/ckeditor5_dev/-/blob/1.0.x/ckeditor5_plugin_starter_template/webpack.config.js

nagyad (dismissed):

sergepavle (approved):

PR #1331, opened 2023-10-26
EWPP-3678: Biuld ECL from scratch and use its artifacts.

ademarco:


ademarco:
Closing this, I'll open a new one where I'll add the full build parts, instead of trying to replace the current build method.

PR #1343, opened 2023-12-04
EWPP-3768: Add inverted styles for EC Footer links.

imanoleguskiza:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra (changes requested):
We need to set `inverted: true` for the Accessibility link in `oe-corporate-blocks-ec-footer--standardised.html.twig` to avoid this:
ec standardised - accessibility link

22Alexandra (dismissed):

nagyad (approved):

22Alexandra (approved):

PR #1344, opened 2023-12-04
EWPP-3767: Update File teaser pattern.

22Alexandra:


imanoleguskiza (dismissed):

.env.dist
```diff
@@ -1 +1 @@
-ECL_BUILD_REF=v4-dev
+ECL_BUILD_REF=EWPP-3767
```
nagyad: We will have to change this to the epic branch before merge, right?

nagyad (dismissed):

.env.dist
22Alexandra: We'll use the v4-oe_theme branch from the fork.

imanoleguskiza (approved):

nagyad (approved):

PR #1357, opened 2023-12-15
EWPP-3826: Remove spaceless filter from long text fields.

nagyad:


templates/field/field--text-long.html.twig
```diff
@@ -6,18 +6,16 @@
...
-      label_display == 'visually_hidden' ? 'visually-hidden',
-    ]
-    %}
-    <div{{ title_attributes.addClass(title_classes) }}>{{ label }}</div>
-  {% endif %}
-  <div class="ecl">
-    {%- include 'field--bare.html.twig' -%}
-  </div>
-{% endapply %}
+{% if not label_hidden %}
```
donquixote:
```suggestion
{%- if not label_hidden -%}
```
Instead of the spaceless we can use the "-" in this and more places.
Not sure how many we need.
The purpose, from what I understand, is to remove whitespace caused by the template itself.

donquixote:
Also a test could be useful:
- Verify that spaces inside the content don't get removed.
- Verify that whitespace from the template is removed (if we care about that).

templates/field/field--text-long.html.twig
nagyad: Good point, thanks!

22Alexandra (approved):

imanoleguskiza (approved):

22Alexandra (approved):

imanoleguskiza (approved):

PR #1358, opened 2023-12-15
EWPP-3826: Remove spaceless filter

donquixote:
Remove spaceless filter, use "-" instead.

I don't know if all of the "-" are really needed. See what happens with or without.

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


donquixote:
oops, duplicate. #1357

PR #1366, opened 2024-01-08
EWPP-3771: Fix link block style.

imanoleguskiza:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/templates/compositions/ec-component-link-block/link-block.html.twig
+++ b/templates/compositions/ec-component-link-block/link-block.html.twig
@@ -24,7 +24,6 @@
 {% set _links = links|default([]) %}
 {% set _css_class = 'ecl-link-block' %}
 {% set _extra_attributes = '' %}
-{% set _color = ecl_component_library == 'ec' ? 'neutral-20' : 'primary-5' %} %}
 
 {# Internal logic - Process properties #}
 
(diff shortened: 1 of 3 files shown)


templates/compositions/ec-component-link-block/link-block.html.twig
```diff
@@ -24,6 +24,7 @@
 {% set _links = links|default([]) %}
 {% set _css_class = 'ecl-link-block' %}
 {% set _extra_attributes = '' %}
+{% set _color = ecl_component_library == 'ec' ? 'neutral-20' : 'primary-5' %} %}
```
22Alexandra:
```suggestion
{% set _color = ecl_component_library == 'ec' ? 'neutral-20' : 'primary-5' %}
```

sergepavle: Why here use `neutral-20` color? Do we have any reference on what exactly color should be used? Why not just `neutral` for example or `neutral-40` for example?
It is discussable question.

sergepavle (dismissed):

templates/compositions/ec-component-link-block/link-block.html.twig
imanoleguskiza: To be honest we can discuss and decide I think, I chose that one pretty much at random. But I will use the twig method created by @22Alexandra so it is centralized if we want to change it in the future

22Alexandra (approved):

sergepavle (approved):

PR #1370, opened 2024-01-10
EWPP-3765: Update contextual navigation composition.

22Alexandra:


sass/components/_contextual_navigation.scss
```diff
@@ -1,8 +1,8 @@
 @use "sass:map";
```
fmmribeiro: We could remove this line because we dont use map anymore 
```suggestion
```

fmmribeiro (dismissed):

sergepavle (approved):

fmmribeiro (approved):

PR #1374, opened 2024-01-14
EWPP-3875: Update News CT rendering tests.

22Alexandra:


sergepavle (approved):

sergepavle (dismissed):

templates/field/field--node--oe-news-sources--oe-news--full.html.twig
```diff
@@ -6,10 +6,11 @@
  * @see ./core/themes/stable/templates/field/field.html.twig
  */
 #}
+{% set _color = ecl_component_library == 'ec' ? 'neutral-40' : 'primary-10' %}
 <h3 class="ecl-u-type-heading-3 ecl-u-mt-3xl">{{ 'Sources'|t }}</h3>
 {% for item in items %}
   {% set bubble_cache = item.content|render %}
-  <div class="ecl-u-border-bottom ecl-u-border-color-grey-15 ecl-u-pt-m ecl-u-pb-m">
+  <div class="ecl-u-border-bottom ecl-u-border-color-{{ _color }} ecl-u-pt-m ecl-u-pb-m">
```
ademarco: Since we repeat this 5 times only in this PR, we could turn this into a Twig function that, depending on the library , prints out the right class, as `ecl-u-border-color-{{ _color }}`, like this:

```suggestion
  <div class="ecl-u-border-bottom {{ ecl_class_border_color() }} ecl-u-pt-m ecl-u-pb-m">
```

ademarco (changes requested):

ademarco (approved):

sergepavle (approved):

PR #1379, opened 2024-01-19
EWPP-3912: Adapt site and page header templates.

22Alexandra:


tests/src/Functional/SiteHeaderTest.php
```diff
@@ -156,15 +155,20 @@ public function testSiteHeader(): void {
...
       $assert->elementAttributeExists('css', 'a.ecl-site-header__language-selector', 'href');
       $assert->elementAttributeExists('css', 'a.ecl-site-header__language-selector', 'data-ecl-language-selector');
       $this->assertEquals('button', $language_switcher_button->getAttribute('role'));
       $this->assertEquals('Change language, current language is English', $language_switcher_button->getAttribute('aria-label'));
       $this->assertEquals('language-list-overlay', $language_switcher_button->getAttribute('aria-controls'));
-      $icon = $language_switcher_button->find('css', "span.ecl-site-header__language-icon svg.ecl-icon.ecl-icon--s.ecl-site-header__icon[focusable='false'][aria-hidden='true']");
-      $this->assertEquals('<use xlink:href="/build/themes/custom/oe_theme/dist/' . $component . '/images/icons/sprites/icons.svg#global" xmlns:xlink="http://www.w3.org/1999/xlink"></use>', $icon->getHtml());
-      $this->assertEquals('English', $language_switcher_button->getText());
+      file_put_contents('test.html', $this->getSession()->getPage()->getOuterHtml());
```
nagyad: Debug code to be removed.

nagyad (changes requested):

templates/patterns/navigation_menu/navigation_menu.ui_patterns.yml
```diff
@@ -17,10 +17,14 @@ navigation_menu:
...
+      type: "array"
       label: "Close button"
-      description: "Close button label."
-      preview: "Close"
+      description: "Close button label and whether only an icon should be displayed."
+      preview:
+        label: 'Close'
+        icon_name: 'close-filled'
+        icon_size: 'm'
+        hide_label: false
```
ademarco: This changes the API of this particular pattern, could we, instead, hide the change in the pattern implementation, so that preview stays `Close` and then we add that array in the pattern itself. The button is just the close button anyway, we can only offer the possibility to change its label.

templates/patterns/navigation_menu/pattern-navigation-menu.html.twig
```diff
@@ -72,7 +72,13 @@
       }
     },
     close: {
-      label: close|default('Close'|t),
+      label: close.label|default('Close'|t),
+      icon: {
+        path: ecl_icon_path,
+        name: close.icon_name,
+        size: close.icon_size,
```
ademarco: See comment above, here just hard code values for the close button and leave only the label to be customizable.

tests/src/Functional/SiteHeaderTest.php
ademarco: @22Alexandra yep, let's remove this if not needed.

ademarco (dismissed):

ademarco (approved):

nagyad (approved):

PR #1383, opened 2024-01-24
EWPP-3874: Update Event CT.

22Alexandra:


modules/oe_theme_content_event/templates/oe-theme-content-event-contacts.html.twig
```diff
@@ -13,7 +13,7 @@
       <div id="event-contacts-general">
         <h3 class="ecl-u-type-heading-3 ecl-u-mt-none ecl-u-mb-m ecl-u-mb-m-l">{{ 'General contact'|t }}</h3>
         {% for contact in general %}
-          <div{{ not loop.last ? ' class="ecl-u-mb-xl ecl-u-pb-xl ecl-u-border-bottom ecl-u-border-color-grey-20"'}}>
+          <div{{ not loop.last ? ' class="ecl-u-mb-xl ecl-u-pb-xl ecl-u-border-bottom {{ ecl_class_border_color() }}"'}}>
```
imanoleguskiza: We have two prints inside one another ({{ ecl_class_border_color() }} is inside another {{ }}) and it just prints it as a string. The resulting div is: <div class="ecl-u-mb-xl ecl-u-pb-xl ecl-u-border-bottom {{ ecl_class_border_color() }}">

imanoleguskiza: I think this is not tested either in tests/src/Functional/ContentEventRenderTest.php so maybe a good time to add an assertion for it.

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

ademarco:
@22Alexandra looks good but we have an issue with the calendar icon on mobile devices, see GIF below:

Peek 2024-01-30 11-26

22Alexandra:
> @22Alexandra looks good but we have an issue with the calendar icon on mobile devices, see GIF below:
> 
> Peek 2024-01-30 11-26

Unfortunately, there is not much to be fixed regarding this. This happens for any inline icon with long text in comparison with the inline icons with shorter text because it will need to fit the text on two lines. Please take a look the following video:

https://github.com/openeuropa/oe_theme/assets/22908988/4e66a56b-52d7-4cea-8d7b-6ad98e7eafb1

**UPDATE:** The same behaviour on v3 in production: https://comm-ewcms-demo-ec.prd.fpfis.tech.ec.europa.eu/event-collection-demo/event-demo-2022-06-01_en

ademarco (dismissed):

ademarco:
I see, it's an upstream problem, we should open a ticket on ECL then, I'll approve this for the time being.

upchuk (approved):

sergepavle (approved):

PR #1384, opened 2024-01-26
EWPP-3878: Update Person CT.

22Alexandra:


sergepavle (commented):
I've noticed that we have 2 div's wrappers. Could be that it is an old issue.
image

sergepavle (dismissed):

sergepavle (approved):

upchuk (approved):

PR #1386, opened 2024-01-29
EWPP-3877: Reactivate page content type tests.

nagyad:


tests/src/Kernel/LegacyContentRenderTest.php
```diff
@@ -56,7 +55,7 @@ public function testPage(): void {
     // Related links.
     $related_links_heading = $crawler->filter('.ecl-u-type-heading-2');
     $this->assertStringContainsString('Related links', $related_links_heading->text());
-    $related_links = $crawler->filter('div.ecl-u-border-bottom.ecl-u-border-color-grey-15.ecl-u-pt-m.ecl-u-pb-m a');
+    $related_links = $crawler->filter('div.ecl-u-border-bottom.ecl-u-border-color-neutral-40.ecl-u-pt-m.ecl-u-pb-m a');
```
22Alexandra, line 58: This needs to be changed in the field template as well.
Also, on line 66 the selector for the ECL Content item component doesn't use an h1 tag for the title anymore, I believe it's a div so it should be fixed too.

22Alexandra (changes requested):

22Alexandra (approved):

22Alexandra (approved):

sergepavle (approved):
for @imanoleguskiza

PR #1394, opened 2024-01-30
EWPP-3880: Fix content type "Project" tests.

fmmribeiro:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/sass/components/_description_list.scss
+++ b/sass/components/_description_list.scss
@@ -3,11 +3,11 @@
  * Add a featured modifier to ECL description list style.
  */
 @use '@ecl/vanilla-utility-border/border' as *;
+$background-color: var(--ecl-color-neutral-40) !default;
+$border-color: var(--ecl-color-neutral-40) !default;
 
 .ecl-description-list--featured {
-  @extend .ecl-u-border-all !optional;
-  @extend .ecl-u-border-width-2 !optional;
-  @extend .ecl-u-border-color-blue-25 !optional;
-  @extend .ecl-u-bg-blue-5 !optional;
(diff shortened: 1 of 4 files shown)


22Alexandra (changes requested):
We need to fix our custom css because they don't seem to look right.
EC:
EU
EU:
EC

I didn't find the featured variant of the description list component in ECL v4, perhaps they removed it. However, we have our own custom css where we apply border and background color (check `_description_list.scss`). We need to fix the colours on both EC and EU components because for now they look the same since the colours we apply in that file do not exist anymore.
Check the `eclBorderColor` and `eclBackgroundColor` twig functions where we define colours based on the component library so the same ones should be used for this case.

sass/style-ec.scss
```diff
@@ -18,7 +18,11 @@
 @use "./components/toolbar";
 @use "./components/wysiwyg";
 @use "./components/global";
-@use "./components/description_list";
+@use "./components/description_list" with (
+  $background-color: var(--ecl-color-neutral-40),
+  $border-color: var(--ecl-color-neutral-40),
+);
+
```
22Alexandra: No need to override the variable values since you're setting the default ones.
```suggestion
@use "./components/description_list";
```

22Alexandra (changes requested):

22Alexandra (approved):

nagyad (approved):

PR #1396, opened 2024-02-07
EWPP-4032: Add Highlighted list pattern and composition.

22Alexandra:


templates/patterns/highlighted_list/highlighted_list.ui_patterns.yml
```diff
@@ -0,0 +1,83 @@
...
+        secondary_meta: "3 minutes read"
+        picture:
+          src: "https://loremflickr.com/1200/900/arch"
+          alt: "Alternative text for image"
+    items:
+      type: "array"
+      label: "Items"
+      description: "List of items compatible with the ECL Content item component."
+      preview:
+        - title: "Item 2"
```
upchuk: Is there a missing `url` from `items`? Also, is there "external link" handling?

templates/patterns/highlighted_list/pattern-highlighted-list.html.twig
```diff
@@ -0,0 +1,164 @@
...
+{% endif %}
+
+{# Prepare the title as a link or simple string for the highlighted_item. #}
+{% if highlighted_item.url is defined and highlighted_item.url is not empty %}
+  {% set highlighted_item_title = {
+    link: {
+      type: 'standalone',
+      label: highlighted_item.title,
+      path: highlighted_item.url,
+      external: highlighted_item.url|is_external_url,
```
upchuk: `is_external_url` crashes here if we pass a Url object to `url`. Check everywhere where it's used

upchuk (changes requested):

templates/patterns/highlighted_list/highlighted_list.ui_patterns.yml
22Alexandra: Is not missing, should be possible to pass title without url as well. The external link is handled when the url is provided.

https://github.com/openeuropa/oe_theme/pull/1396/files#diff-483d6c5e23122fb0e98a232b9507e78f9ed69d3ca7890afdcb174ee3bbf6829bR16

tests/src/PatternAssertions/HighlightedListAssert.php
```diff
@@ -0,0 +1,140 @@
...
+  protected function assertHighlightedItem(?array $expected_highlighted_item, string $selector, Crawler $crawler) {
+    if (is_null($expected_highlighted_item)) {
+      $this->assertElementNotExists($selector, $crawler);
+      return;
+    }
+    $this->assertElementExists($selector, $crawler);
+    $highlighted_wrapper = $crawler->filter($selector);
+    $list_item_assert = new ListItemAssert();
+    $html = $highlighted_wrapper->html();
+    $list_item_assert->assertPattern($expected_highlighted_item, $html);
```
upchuk, line 78: Asserting the ListItemAssert doesn;t work out of the box. I think we have a problem in ` ListItemAssert::assertThumbnailImage()`:

IMO, this:

```
 $variant_class = $variant === 'thumbnail_primary' ? 'picture.ecl-content-item__picture--large.ecl-content-item__picture--left' : 'picture.ecl-content-item__picture--large.ecl-content-item__picture--right';
```

should become this:

```
    $variant_class = 'picture.ecl-content-item__picture--large.ecl-content-item__picture--top';
    switch ($variant) {
      case 'thumbnail_primary':
        $variant_class = 'picture.ecl-content-item__picture--large.ecl-content-item__picture--left';
        break;
      case 'thumbnail_secondary':
        $variant_class = 'picture.ecl-content-item__picture--large.ecl-content-item__picture--right';
        break;
    }
```

However, I noticed a problem. For the second item (the top one in the right column), you process it to show the image as `medium` so the class changes to `ecl-content-item__picture--large` meaning the above will fail. Locally I removed that class from the assertion and now it works. Do you have a better way? I guess we cannot pass dynamically..

tests/src/PatternAssertions/HighlightedListAssert.php
```diff
@@ -0,0 +1,140 @@
...
+        [$this, 'assertElementText'],
+        'div#highlighted-news-block h2.ecl-u-type-heading-2',
+      ],
+      'highlighted_item' => [
+        [$this, 'assertHighlightedItem'],
+        'div#highlighted-news-block div.ecl-row div.ecl-col-l-8.ecl-u-d-flex.ecl-u-flex-column',
+      ],
+      'items' => [
+        [$this, 'assertItems'],
+        'div.highlighted-news-divider div.ecl-row div.ecl-col-l-4.ecl-u-d-flex.ecl-u-flex-column',
```
upchuk: This needs to be
```
div#highlighted-news-block div.ecl-row div.ecl-col-l-4.ecl-u-d-flex.ecl-u-flex-column
```

tests/src/PatternAssertions/HighlightedListAssert.php
```diff
@@ -0,0 +1,140 @@
...
+   * Asserts the detail text from the footer of the highlighted list.
+   *
+   * @param array|null $expected
+   *   The expected description values.
+   * @param string $selector
+   *   The CSS selector to find the detail.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertDescription($expected, string $selector, Crawler $crawler): void {
```
upchuk: Is this needed?

tests/src/PatternAssertions/HighlightedListAssert.php
```diff
@@ -0,0 +1,140 @@
...
+    // Assert the number of items.
+    $items = $items_wrapper->filter('article');
+    self::assertCount(count($expected_items), $items, 'The expected number of items does not correspond with the actual number of items in the list.');
+    // Assert there is no description element for any of the items.
+    $this->assertElementNotExists('.ecl-content-block__description', $items_wrapper);
+
+    // Assert each item's info.
+    $list_item_assert = new ListItemAssert();
+    foreach ($expected_items as $index => $expected_item) {
+      $html = $expected_item->html();
```
upchuk: This should be
```
$html = $items->eq($index)->outerHtml();
```
or it doesn;t make sense :)

upchuk (dismissed):

tests/src/PatternAssertions/HighlightedListAssert.php
22Alexandra: It should be used for the formatted text in the footer, below the See more link.

tests/src/PatternAssertions/HighlightedListAssert.php
22Alexandra: We should be passing it dynamically.

oe_theme.theme
```diff
@@ -2518,3 +2518,78 @@ function oe_theme_page_attachments_alter(array &$attachments) {
 function _oe_theme_sanitize_url($url): mixed {
   return $url instanceof Url ? $url : UrlHelper::filterBadProtocol($url);
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_theme_preprocess_pattern_highlighted_list(array &$variables): void {
```
sergepavle, line 2525: It looks like most of this code can be moved into template.

templates/compositions/ec-component-highlighted-news-block/highlighted-news-block.html.twig
```diff
@@ -0,0 +1,43 @@
...
+  @file
+  Theme implementation for the Highlighted News Block
+  Parameters:
+    - "title" (string) (default: ''): The Highlighted News Block title.
+    - "highlighted" (array) (default: []): The main item of the list compatible with the ECL Content Item component.
+    - "items" (array) (default: []): Highlighted News items compatible with the ECL Content Item component.
+    - "see_more" (array) (default: []): Predefined structure compatible with Link component.
+    - "detail" (string) (default: ''): Additional information to be displayed below the 'See more' link.
+#}
+  <div id="highlighted-news-block">
```
sergepavle: Is it safe to use hard-coded id?

22Alexandra: I'd say so, there is no JS, what would be the concern?

oe_theme.theme
22Alexandra: This would keep the template cleaner, no?

sergepavle: Normally before we tried to move simple manipulation with data inside template. But it is discussable. We could leave like that.

templates/compositions/ec-component-highlighted-news-block/highlighted-news-block.html.twig
sergepavle: Just for keeping validity of html.

oe_theme.theme
22Alexandra: We need to override too much data of the list_item pattern in order to make sure the data is correct for this pattern. I'd keep it in the preprocess.

sergepavle (dismissed):

sergepavle (approved):

fmmribeiro (approved):

PR #1413, opened 2024-02-19
JMSITE-340: Do not force the library favicon if the theme is not conf…

imanoleguskiza:
…igured to do so.

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra:
Hello @imanoleguskiza, we no longer support 4.x. The fix was merged 5.x.

PR #1420, opened 2024-02-27
EWPP-4137: Cleanup deprecated patterns and fields and update README.

22Alexandra:


modules/oe_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -507,13 +507,6 @@ protected function getIconPath(array $context, string $icon): string {
    *   The trimmed output.
    */
   public function smartTrim(Environment $env, $input, $limit = 0) {
-    if ($limit === NULL) {
-      // phpcs:disable Drupal.Semantics.FunctionTriggerError
-      @trigger_error('Using the smart_trim filter with a null limit value is deprecated in oe_theme:3.x and will be removed in oe_theme:4.x releases.', E_USER_DEPRECATED);
-      // phpcs:enable
-      $limit = 0;
-    }
-
```
ademarco, line 516: If we got NULL, in 3.x we were changing it to 0, what should we do now? We set the type to `int` in the argument?

```php
  public function smartTrim(Environment $env, $input, int $limit = 0) {
```

ademarco (changes requested):

ademarco (approved):

nagyad (approved):

PR #1436, opened 2024-03-25
EWPP-4085: Add mega menu variant to navigation menu pattern.

22Alexandra:


modules/oe_theme_helper/templates/oe-theme-helper-site-navigation.html.twig
```diff
@@ -15,4 +15,4 @@
   close: close,
   items: menu_items,
   site_name: site_name,
-}, 'vertical') }}
+}, 'default') }}
```
upchuk: you need to make the variant a variable in hook_theme and allow that to be passed. And it can default to vertical or default or whatever you wanna call it. Check also for other potential uses of this.

https://github.com/openeuropa/oe_theme/blob/4.x/modules/oe_theme_helper/templates/oe-theme-helper-site-navigation.html.twig#L13

upchuk (commented):
The extra_attributes dont work.

templates/patterns/navigation_menu/pattern-navigation-menu--variant-mega-menu.html.twig
```diff
@@ -0,0 +1,160 @@
...
+  {# Process children, if any. #}
+  {% set _children = [] %}
+  {% for child in item.children %}
+    {% set _child = {
+      path: child.href,
+      label: child.label,
+      see_all: child.see_all|default(false),
+      extra_attributes: child.extra_attributes,
+    } %}
+    {% if child.is_current is defined and child.is_current == true%}
```
sergepavle:
```suggestion
    {% if child.is_current is defined and child.is_current == true %}
```

sergepavle (dismissed):

sergepavle (dismissed):

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -95,6 +95,12 @@ function oe_theme_helper_theme($existing, $type, $theme, $path) {
       'variables' => [
         'site_name' => '',
         'menu_items' => [],
+        'back' => '',
+        'see_all_label' => '',
```
upchuk, line 99: As far as I see, the see_all has been moved to the second level so it can go away from here.

imanoleguskiza (dismissed):

upchuk (approved):

PR #1438, opened 2024-04-04
EWPP-4139: Add a variant field to navigation_list pattern.

22Alexandra:


templates/patterns/navigation_list/navigation_list.ui_patterns.yml
```diff
@@ -2,6 +2,11 @@ navigation_list:
   label: "Navigation list"
   description: "Navigation lists allow to structure groups of links together."
   fields:
+    item_variant:
```
imanoleguskiza: We should be using the patterns variant functionality like we have in banner

imanoleguskiza (changes requested):

imanoleguskiza (approved):

fmmribeiro (approved):

PR #1452, opened 2024-05-16
Fix issue https://github.com/openeuropa/oe_theme/issues/1037

### LINKED ISSUE #1037: Node forms that include a Date field and a field with unlimited values cause error: Maximum call stack size exceeded.
Create a node content type with a Date input field and a field with unlimited "Allowed number of values". For example a URL input field.Select view widget for Date field as "Date and time" which applies the Datepicker input for the Date field. 
Make the node access public or authorized in order to use oe_theme in the frontend as a signed-in user.
Next, open the node edit form. Input a date with the Datepicker field and then click on "Add" button to add an extra field input for the multivalue field ( For example add 2 URLs in the node ). 

Result: The form will produce the JS error: ecl-ec.js?r74vg1:2 Uncaught RangeError: Maximum call stack size exceeded.
After a while the form page becomes unresponsive and the browser window crashes. 

In our case, we are using a Publication node type with a publication date input field and a multivalue Paragraph field "Author". 
image
image
image

pmoutsophypertech:
### Description

Provides a fix for issue https://github.com/openeuropa/oe_theme/issues/1037 so that datepicker works for websites having BigPipe enabled too.

### Change log

- Added: Add dependency on "core/once".
- Changed: The added dependency requires at least Drupal 9.2 according to https://www.drupal.org/node/3158256
- Deprecated: -
- Removed:-
- Fixed: https://github.com/openeuropa/oe_theme/issues/1037
- Security: -

### DIFF
--- a/js/ecl_datepicker.js
+++ b/js/ecl_datepicker.js
@@ -5,7 +5,7 @@
 (function (ECL, Drupal) {
   Drupal.behaviors.eclDatepicker = {
     attach: function attach(context, settings) {
-      var elements = document.querySelectorAll('[data-ecl-datepicker-toggle]');
+      var elements = once('ecl-datepicker', document.querySelectorAll('[data-ecl-datepicker-toggle]'));
       for (var i = 0; i < elements.length; i++) {
         var datepicker = new ECL.Datepicker(elements[i], {
           format: settings.oe_theme.ecl_datepicker_format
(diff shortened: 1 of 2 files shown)


22Alexandra:
Hello @pmoutsophypertech, thank you for the contribution. Unfortunately, we do not provide support for 4.x anymore.
We suggest updating to the latest version of 5.x or 6.x (README will be soon updated) and open an issue if you still encounter the bug. The newer versions also use Duet.js for datepickers.

PR #1460, opened 2024-06-03
EWPP-4383: Update gallery to render media name as title parameter.

22Alexandra:


sergepavle (dismissed):

src/ValueObject/GalleryItemValueObject.php
```diff
@@ -65,12 +72,15 @@ class GalleryItemValueObject extends ValueObjectBase {
...
+   * @param string $title
+   *   Title for the gallery item.
    * @param string $caption
    *   Caption for the gallery item.
    * @param string $meta
    *   Meta for the gallery item, such as a copyright note.
    */
-  private function __construct(ImageValueObjectInterface $thumbnail, string $source, string $type, string $caption = '', string $meta = '') {
+  private function __construct(ImageValueObjectInterface $thumbnail, string $source, string $type, string $title = '', string $caption = '', string $meta = '') {
+    $this->title = $title;
```
upchuk, line 83: Let's put the title as the last parameter to maintain some BC and values don't get set wrongly if someone uses this without a title.

upchuk (changes requested):

upchuk (dismissed):

PR #1469, opened 2024-06-19
EWPP-4401: Pass hide_label variable to the social media links pattern.

intelektron:
### Description

EWPP CMS defines a field that allows to hide social media labels, leaving only icons. However, the `hide_label` parameter is not passed to the pattern.

### DIFF
--- a/templates/patterns/social_media_links/pattern-social-media-links.html.twig
+++ b/templates/patterns/social_media_links/pattern-social-media-links.html.twig
@@ -12,6 +12,7 @@
       path: link.url,
       label: link.label,
       icon_position: 'before',
+      hide_label: link.hide_label|default(false),
     }
   } %}
   {% if link.service is not empty %}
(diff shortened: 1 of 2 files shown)


templates/patterns/social_media_links/pattern-social-media-links.html.twig
```diff
@@ -12,6 +12,7 @@
       path: link.url,
       label: link.label,
       icon_position: 'before',
+      hide_label: link.hide_label|default(false),
```
22Alexandra, line 15: Let's update one of the links in `social_media_links.ui_patterns.yml` so it's also visible in the patterns overview page.

22Alexandra (changes requested):

22Alexandra (approved):

22Alexandra (approved):

nagyad (approved):

PR #1474, opened 2024-07-08
EWPP-4509: Update to ECL 4.3.0.

22Alexandra:


sergepavle (dismissed):

composer.json
```diff
@@ -97,7 +97,7 @@
         },
         "patches": {
             "drupal/core": {
-                "https://www.drupal.org/project/drupal/issues/3151609": "https://www.drupal.org/files/issues/2023-12-11/link_internal_as_text_v10-3151609_0.patch"
+                "https://www.drupal.org/project/drupal/issues/3151609": "https://www.drupal.org/files/issues/2024-06-11/link_internal_as_text_v11-3151609.patch"
```
nagyad: EWCMS is still on 10.2, how is this going to apply there?

nagyad (changes requested):

composer.json
nagyad: Currently, the patch applies, which is simply a coincidence, but did you check if it actually works?

nagyad: I tested and it seems to be working fine.

nagyad (dismissed):

PR #1475, opened 2024-07-09
OEL-3051: Update to Drupal 10.3

AaronGilMartinez:


tests/src/Kernel/RenderingTest.php
```diff
@@ -101,14 +101,6 @@ public function submitForm(array &$form, FormStateInterface $form_state) {}
    * @dataProvider renderingDataProvider
    */
   public function testRendering(array $structure, array $assertions): void {
-    // @todo Remove when support for 10.1.x is dropped.
-    if (!empty($structure['core_version'])) {
-      if (!$this->shouldBeTested($structure['core_version'])) {
```
brummbar, line 106: We need to remove the method called here.

brummbar (changes requested):

tests/src/Kernel/RenderingTest.php
AaronGilMartinez: Removed

brummbar (approved):

PR #1477, opened 2024-07-12
EWPP-4496: Add Introduction field to various patterns.

intelektron:
### Description

Add Introduction field to link lists.

### DIFF
--- a/templates/compositions/ec-component-highlighted-news-block/highlighted-news-block.html.twig
+++ b/templates/compositions/ec-component-highlighted-news-block/highlighted-news-block.html.twig
@@ -5,6 +5,7 @@
   Parameters:
     - "columns" (int) (default: 2): The number of columns layout. It can be 2 or 4 and it defaults to 2.
     - "title" (string) (default: ''): The Highlighted News Block title.
+    - "introduction" (string) (default: ''): A brief summary of the block content.
     - "highlighted" (array) (default: []): The main item of the list compatible with the ECL Content Item component.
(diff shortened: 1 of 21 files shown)


templates/compositions/ec-component-highlighted-news-block/highlighted-news-block.html.twig
```diff
@@ -19,6 +20,14 @@
         </div>
       </div>
     {% endif %}
+    {% set _introduction = introduction|default('') %}
```
22Alexandra: Why is this needed? Instead of moving the value from a variable to another, we can just check if `introduction` is not empty.

22Alexandra (changes requested):
We need test coverage for all the patterns that have been changed.
Please check `tests/src/Kernel/fixtures/rendering.yml`

templates/compositions/ec-component-highlighted-news-block/highlighted-news-block.html.twig
intelektron: I just followed the way from other components, I can change it.

intelektron: Fixed

intelektron:
> We need test coverage for all the patterns that have been changed. Please check `tests/src/Kernel/fixtures/rendering.yml`

I've added the introduction field to pattern fixtures

22Alexandra (dismissed):

PR #1479, opened 2024-07-15
EWPP-4515: Add new fields for Gallery pattern.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/templates/patterns/gallery/gallery.ui_patterns.yml
+++ b/templates/patterns/gallery/gallery.ui_patterns.yml
@@ -146,3 +146,26 @@ gallery:
       description: "Number of visible items in an expandable gallery."
       preview: 3
       escape: false
+    grid:
+      type: "boolean"
+      label: "Grid display"
+      description: "Display gallery as a grid. Default is false."
+      preview: true
+      escape: false
+    columns:
+      type: "numeric"
+      label: "Columns number"
+      description: "Number of columns. Default is 2. Only used if the grid display is enabled."
+      preview: 2
+      escape: false
+    ratio:
+      type: "string"
+      label: "Aspect ratio"
+      description: "Image aspect ratio. Default is 3-2. Only used if the grid display is enabled."
+      preview: '3-2'
+    disable_hover:
+      type: "boolean"
+      label: "Disable hover"
+      description: "Hides the item's title on hover. Default is false."
+      preview: false
+      escape: false
--- a/templates/patterns/gallery/pattern-gallery.html.twig
+++ b/templates/patterns/gallery/pattern-gallery.html.twig
@@ -17,6 +17,11 @@
   } %}
 {% endif %}
 
+{% if grid %}
+  {% set _columns = columns|default(2) %}
+  {% set _ratio = ratio|default('3-2') %}
+{% endif %}
+
 {% set _gallery = {
   'overlay': {
     'counter_separator': ' / ',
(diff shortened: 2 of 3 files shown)


templates/patterns/gallery/pattern-gallery.html.twig
```diff
@@ -17,6 +17,11 @@
   } %}
 {% endif %}
 
+{% if grid %}
+  {% set columns = columns|default(2) %}
+  {% set ratio = ratio|default('3-2') %}
```
22Alexandra: If the same name is used for the variables in the template, we rely on the ECL implementation so checking if the `grid` is enabled wouldn't be needed. Instead let's use different variable names to be passed to the ECL component.
```suggestion
  {% set _columns = columns|default(2) %}
  {% set _ratio = ratio|default('3-2') %}
```

templates/patterns/gallery/pattern-gallery.html.twig
```diff
@@ -63,6 +68,10 @@
   'expandable': expandable ?? true,
   'visible_items': visible_items|default(8),
   'footer': footer_link|default([]),
+  'grid': grid|default(false),
+  'column': columns|default(null),
+  'ratio': ratio|default(null),
```
22Alexandra: Considering the comment above, let's rename the passed variables for columns and ratio.
```suggestion
  'column': _columns|default(null),
  'ratio': _ratio|default(null),
```

templates/patterns/gallery/gallery.ui_patterns.yml
```diff
@@ -146,3 +146,27 @@ gallery:
...
+      escape: false
+    columns:
+      type: "numeric"
+      label: "Columns number"
+      description: "Number of columns. Default is 2. Only used if the grid display is enabled."
+      preview: 2
+      escape: false
+    ratio:
+      type: "string"
+      label: "Column number"
```
22Alexandra: It's not the column number.
```suggestion
      label: "Aspect ratio"
```

templates/patterns/gallery/gallery.ui_patterns.yml
```diff
@@ -146,3 +146,27 @@ gallery:
...
+      description: "Image aspect ratio. Default is 3-2. Only used if the grid display is enabled."
+      preview: 2
```
22Alexandra: This value for the preview is wrong and it's not working.

templates/patterns/gallery/gallery.ui_patterns.yml
```diff
@@ -146,3 +146,27 @@ gallery:
...
+      escape: false
```
22Alexandra: Do we need to escape a string?
Let's also add the empty new line at the end of the file.

tests/src/Kernel/fixtures/rendering.yml
```diff
@@ -785,6 +786,14 @@
       '.ecl-gallery .ecl-button.ecl-button--ghost.ecl-gallery__slider-previous': 0
       '.ecl-gallery .ecl-button.ecl-button--ghost.ecl-gallery__slider-next': 0
       '.ecl-gallery__footer-link': 0
+      '.ecl-gallery--grid': 0
+      '.ecl-gallery--col-2': 0
+      '.ecl-gallery--col-3': 0
+      '.ecl-gallery--col-4': 0
+      '.ecl-gallery--col-5': 0
+      '.ecl-gallery--ratio-3-2': 0
+      '.ecl-gallery--ratio-3-1': 0
```
22Alexandra, line 816: Values for `columns` and `ratio` should be provided in order to assert they are not used if grid is not enabled.

22Alexandra (changes requested):
Let's fix the ticket number in the commit message and PR title.

22Alexandra (dismissed):

nagyad (dismissed):

PR #1481, opened 2024-07-19
EWPP-4551: Update ECL version.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


22Alexandra (changes requested):
ECL 4.4.0 kinda removes the Banner component variants but keeps adding the classes to keep some sort of BC. Doing this, the rendering of the plain background variant is wrong: it renders both a plain background + the image if available.
plain background

This should not happen in EWCMS since the Plain background variant doesn't have an image field but we need to fix it in oe_theme. So, let's remove the image if available when rendering this variant. Please also add a test case to ensure this.

22Alexandra (dismissed):

nagyad (dismissed):

22Alexandra (approved):

upchuk (approved):

nagyad (approved):

PR #1486, opened 2024-08-01
EWPP-4580: Allow additional html tags to be rendered in CTs summary field.

22Alexandra:


nagyad (approved):

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
```diff
@@ -110,7 +110,7 @@ protected function getIntroductionMetadata(string $field_name): array {
       // We strip the tags because the ECL component expects only one paragraph
       // of text and the field is using a text format which adds paragraph tags.
       '#type' => 'inline_template',
-      '#template' => '{{ summary|render|striptags("<strong><a><em><svg><use><span>")|raw }}',
+      '#template' => '{{ summary|render|striptags("<strong><a><em><svg><use><span><sub><sup><u><ol><ul><li>")|raw }}',
```
nagyad: The ticket says only `<sub><sup><u>` to be enabled.

nagyad (changes requested):

modules/oe_theme_helper/src/Plugin/PageHeaderMetadata/NodeViewRoutesBase.php
22Alexandra: I know but the reason we're stripping tags is because the ECL doesn't except multiple p tags while it renders all the other tags allowed in the simple rich text editor. Why shouldn't we enable the ordered and unordered list then?

nagyad (approved):

PR #1488, opened 2024-08-02
UCPKN-3176: Accessibility statement link in the footer.

peterserfozo:
### Description

The accessibility statement link logic is in the oe_coporate_blocks module from now. See https://github.com/openeuropa/oe_corporate_blocks/pull/155

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_theme.theme
+++ b/oe_theme.theme
@@ -1840,14 +1840,7 @@ function oe_theme_preprocess_oe_corporate_blocks_ec_footer(array &$variables) {
     $variables['path'] = 'https://commission.europa.eu/index_' . EuropeanUnionLanguages::getInternalLanguageCode($variables['current_language_id']);
   }
   _oe_theme_preprocess_site_specific_footer_links($variables);
-
-  if (function_exists('oe_corporate_blocks_preprocess_set_accessibility_link')) {
-    oe_corporate_blocks_preprocess_set_accessibility_link($variables);
-
-    if (isset($variables['accessibility_link']) && $variables['accessibility_link'] instanceof Url) {
-      $variables['accessibility_link'] = $variables['accessibility_link']->toString();
-    }
-  }
+  _oe_theme_preprocess_accessibility_link($variables);
 }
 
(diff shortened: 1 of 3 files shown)


drishu (dismissed):

drishu (dismissed):

drishu (dismissed):

oe_theme.theme
```diff
@@ -1942,38 +1941,17 @@ function _oe_theme_preprocess_site_specific_footer_links(array &$variables): voi
...
-        // Do not show the button if access to the node is forbidden.
-        return;
-      }
-    }
-  }
-  $url = Url::fromUri($uri)->toString();
 
-  $variables['accessibility_link'] = $url;
-  $cacheability->applyTo($variables);
+  $variables['accessibility_link'] = $variables['accessibility_link']->toString();
```
brummbar: I am not sure I understand the need of manually calling the preprocess function as it's already set in the two theme implementations.
The accessibility_link variable is already there when I debugged the method, and the toString() method is not needed as far as I can see, because it's called by twig when rendering. I have commented out this method and the links were showing ok for me. The only limitation that we need to be sure to put is a constraint in both oe_theme and oe_corporate_blocks to make sure that previous versions of either components are not used with an older version of the other.
Or, alternatively, we keep the code in this method, leave a todo for the next major, and bail out if the accessibility_link is already there.

brummbar (changes requested):

oe_theme.theme
drishu: removed the function entirely and marked the conflict.

AaronGilMartinez (dismissed):

brummbar (dismissed):

drishu (dismissed):

brummbar (approved):

AaronGilMartinez (approved):

PR #1496, opened 2024-09-02
EWPP-4693: Allow 'mark' tag for page header introduction field.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
+++ b/modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
@@ -81,11 +81,14 @@ class NodeViewRoutesMetadataTest extends BrowserTestBase {
    */
   public function testNodeRoutes(): void {
     // Create a published revision for a node.
-    $published_revision_body = $this->randomString() . '<mark>published</mark>';
+    $published_revision_body = '<u>Custom page</u> introduction with <sub>rich</sub><sup>text</sup><mark>highlighted</mark>.';
     $node = $this->drupalCreateNode([
       'type' => 'test',
       'moderation_state' => 'published',
-      'body' => $published_revision_body,
+      'body' => [
+        'value' => $published_revision_body,
+        'format' => 'full_html',
+      ],
     ]);
     // Save the revision url for later access.
     $first_revision_url = $node->toUrl('revision');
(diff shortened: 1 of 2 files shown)


modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
```diff
@@ -68,18 +69,26 @@ public function setUp(): void {
       'view latest version',
       'view any unpublished content',
     ]);
+
+    FilterFormat::create([
```
22Alexandra, line 73: Let's remove this.

modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
```diff
@@ -68,18 +69,26 @@ public function setUp(): void {
...
+    ])->save();
   }
 
   /**
    * Tests that the correct metadata is retrieved on node view routes.
    */
   public function testNodeRoutes(): void {
     // Create a published revision for a node.
-    $published_revision_body = $this->randomString();
+    $published_revision_body = '<u>Custom page</u> introduction with <sub>rich</sub><sup>text</sup><mark>highlighted</mark>.';
```
22Alexandra, line 84: Why are we testing the exact same things as in `\Drupal\Tests\oe_theme_helper\Functional\PageHeaderBlockTest::testRendering`? Could you please just update the mentioned test method and remove all the changes in this class?

22Alexandra (changes requested):

modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
sergepavle: I thought it doesn't work without that. I will recheck

modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
sergepavle: I have found that test in `\Drupal\Tests\oe_theme_helper\Functional\PageHeaderBlockTest::testRendering` doest take in account allowed tags in `\Drupal\oe_theme_helper\Plugin\PageHeaderMetadata\NodeViewRoutesBase::getIntroductionMetadata`

modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
sergepavle: Without creating filter format test is failed.

modules/oe_theme_helper/tests/src/Functional/NodeViewRoutesMetadataTest.php
sergepavle: So it will always green even if we try to put tags which should be stripped.

22Alexandra: You're right 😞

22Alexandra (approved):

22Alexandra (approved):

PR #1502, opened 2024-09-12
EWPP-4750: Add extra_attributes to patterns used to render node templates.

22Alexandra:


nagyad (dismissed):

templates/content/node--oe-call-proposals--teaser.html.twig
```diff
@@ -16,6 +16,8 @@
   {% set additional_information = additional_information|merge([additional_information_items]) %}
 {% endif %}
 
+{% set extra_attributes = [] %}
```
sergepavle: I didn't get why in templates we added `extra_attributes` but before inserting set this empty. Is it intentional?
Because, in this case, external projects cannot set extra_attributes anyway.

sergepavle (dismissed):

PR #1517, opened 2024-10-11
EWPP-4825: Preview extension for file patterns.

intelektron:
### Description

Modification of the patterns to support additional properties.

### DIFF
--- a/templates/patterns/file_teaser/file_teaser.ui_patterns.yml
+++ b/templates/patterns/file_teaser/file_teaser.ui_patterns.yml
@@ -23,6 +23,13 @@ file_teaser:
         size: "140000"
         name: "document.pdf"
         language_code: "en"
+    link_attributes:
+      type: "array"
+      label: "The download link attributes"
+      description: "Any additional attributes to be added to the file download link."
+      preview:
+        - "name": "data-attribute-foo"
+          "value": "bar"
     thumbnail:
       type: "array"
       label: "Thumbnail"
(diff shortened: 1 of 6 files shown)


templates/patterns/file_teaser/file_teaser.ui_patterns.yml
```diff
@@ -23,6 +23,11 @@ file_teaser:
         size: "140000"
         name: "document.pdf"
         language_code: "en"
+    preview_extension:
```
intelektron: At this point we can think about whether `oe_theme` should be aware of the existence of `data-wt-preview`, or whether we should perhaps pass a list of extra parameters defined in the preprocess. So we would have something like this instead:

```php
$variables['download_extra_attributes'] = [
  [ 'name' => 'data-wt-preview', 'value' => $extension ]
];
```

This would make it more elastic, but it might be overengineered.

sergepavle (approved):

templates/patterns/file_translation/file_translation.ui_patterns.yml
```diff
@@ -18,6 +18,13 @@ file_translation:
         size: "150000"
         name: "document.pdf"
         language_code: "en"
+    link_attributes:
```
sergepavle, line 21: Why in the file_teaser pattern we add the possibility to add additional attributes but in the file_translation pattern we don't have a chance to specify extra_attributes for each translation?
I've checked in ewcms but didn't find that it works.
Is it intentional?
image

sergepavle: I've checked and seems like it is possible to add extra_attributes for file_translation items as well.

intelektron: This is limited by the fact, that `file_translation` has a different definition of the `translations` property. It does not allow to pass any additional data along with the translated file (it just takes a list of `FileValueObject`). So I can't pass the `extra_attributes` too. To change it, i would have to refactor the pattern and all references to it.

sergepavle (approved):

fmmribeiro (approved):

PR #1521, opened 2024-10-21
EWPP-4830: Add ecl class to paragraphs and patterns heading so ECL default styling is applied.

22Alexandra:


templates/paragraphs/paragraph--oe-facts-figures.html.twig
```diff
@@ -7,9 +7,11 @@
  */
 #}
 {% if title is not empty %}
-  <h2 class="ecl-u-type-heading-2">
-    {{ title }}
-  </h2>
+  <div class="ecl">
+    <h2 class="ecl-u-type-heading-2">
```
sergepavle, line 11: As I understand .ecl class was dedicated to automatic styling tags (like h2) in the case when we don't have ecl-related classes. Like HTML coming from ckeditor.
Is it from ecl team recommendation?

22Alexandra: It is not a recommendation from the ECL team but yes, the "ecl" was added to style the heading because the margin of the heading:
`heading-margin: var(--s-2xl) 0 var(--s-m)` - 2xl top and m bottom does not apply if the wrapper does not have the "ecl" class.

sergepavle (approved):

intelektron (approved):

PR #1547, opened 2025-01-30
Release-4.5.0: Update drone settings for artifact creation.

nagyad:


.drone.yml
```diff
@@ -112,8 +112,8 @@ pipeline:
       event:
         include: tag
       matrix:
-        CORE_VERSION: "10.2.0"
-        PHP_VERSION: "8.1"
+        CORE_VERSION: "10.4.0"
+        PHP_VERSION: "8.3"
```
sergepavle: I think it is better to use the lowest version of PHP/core anyway, just in case, to avoid theoretical issues with back compatibility.

nagyad: Done

sergepavle (approved):

22Alexandra (approved):

PR #1561, opened 2025-03-12
EWPP-5343: Fix duplicated IDs in select and multiselect facets.

intelektron:
### Description

Fix the select and multiselect labels by de-duplicating "id" attributes and setting proper "for" attributes.

### DIFF
--- a/templates/form/select.html.twig
+++ b/templates/form/select.html.twig
@@ -56,6 +56,7 @@
 {% endfor %}
 
 {% include '@ecl-twig/form-group' with {
+  label: label,
   id: attributes.id,
   name: attributes.name,
   input: {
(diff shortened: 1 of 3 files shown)


tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
```diff
@@ -127,6 +127,16 @@ public function testEclMultiSelect(): void {
     $select_dropdown = $this->getSession()->getPage()->find('css', 'div.ecl-select__multiple-dropdown.ecl-select__container.ecl-select__container--m');
     Assert::assertFalse($this->getSession()->getDriver()->isVisible($select_dropdown->getXpath()));
 
+    // Assert the label is present and it points to the right id.
```
intelektron, line 130: I did not update the rendering test, as there were not many changes in the output HTML.

Also, I wanted to avoid counting elements based on their IDs, the results may not be reliable, as IDs are supposed to be unique.

The rendering test checks for the label existence, which seems to be sufficient.

templates/form/select.html.twig
```diff
@@ -56,6 +56,7 @@
 {% endfor %}
 
 {% include '@ecl-twig/form-group' with {
+  label: label,
```
intelektron, line 59: By setting it like this, we hand over the label handling to ECL.

Without this variable, we would have to change the `for` attribute manually, to point to the multiselect toggle.

The change was done according to the ECL documentation of the component

22Alexandra (approved):

fmmribeiro (approved):

PR #1567, opened 2025-03-26
EWPP-5369: Move the "Report an IT vulnerability" link

intelektron:
### Description

Move the "Report an IT vulnerability" link to another footer column for standarised branding.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -112,6 +112,9 @@
             },
             "drupal/address": {
                 "https://www.drupal.org/project/address/issues/3144823": "https://www.drupal.org/files/issues/2020-11-05/3144823-6.patch"
+            },
+            "openeuropa/oe_corporate_blocks": {
+                "EWPP-5369": "https://github.com/openeuropa/oe_corporate_blocks/compare/5.x...EWPP-5369.diff"
             }
         },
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -112,6 +112,9 @@
             },
             "drupal/address": {
                 "https://www.drupal.org/project/address/issues/3144823": "https://www.drupal.org/files/issues/2020-11-05/3144823-6.patch"
+            },
+            "openeuropa/oe_corporate_blocks": {
+                "EWPP-5369": "https://github.com/openeuropa/oe_corporate_blocks/compare/5.x...EWPP-5369.diff"
```
22Alexandra, line 117: Nothing wrong with this but it might cause issues during the packaging. @nagyad would it be better if the dev branch is used directly instead of a patch?

intelektron: I just didn't want to lock this version of `oe_corporate_blocks` in ewcms too, as there is another definition of this module

nagyad: Normally, this should be fine, but since we are planning to package in the meantime it is better to put the dev branch directly.

22Alexandra (approved):

Issue #1617, opened 2025-06-30
Twig - Flexibility to extend form select & submit, for extra classes + attributes


RaphTbm:
Currently you have to copy-paste those two whole OE templates to add / re-add classes, attributes.
The two templates for select.html.twig and submit.html.twig can't be extended simply to set/re-set attributes and classes.

This is necessary as the current template are too destructive and are removing completely the class on the final element / input / select. This is causing some broken behavior. 
Like for example the WYSIWYG editor with the format choice selector not having the correct classes and data attributes for the editor JS to work properly and load the editor with the related select element.

22Alexandra:
Hi @RaphTbm, thank you for reporting the issue and for the contribution. It will be addressed in #1630.

PR #1618, opened 2025-06-30
Issue #1617: Twig - Flexibility to extend form select & submit, for extra classes + attributes

RaphTbm:
### Description

Related to #1617 :: Twig - Flexibility to extend form select & submit, for extra classes + attributes

This a non breaking change proposal, so people would still have per project to extend the necessary and set those.
But on the long term, we should maybe force the `attributes` and `attributes.class` values/content into the final form element by default, so things can work properly by default for new projects without the need to extend those templates.

### DIFF
--- a/templates/form/input--submit.html.twig
+++ b/templates/form/input--submit.html.twig
@@ -6,13 +6,14 @@
  * @see ./core/themes/stable/templates/form/input.html.twig
  */
 #}
-{% set extraAttributes = [] %}
+{% set extra_attributes = extra_attributes|default([]) %}
 {% for name,value in attributes %}
-  {% set extraAttributes = extraAttributes|merge([{'name': name, 'value': value}]) %}
+  {% set extra_attributes = extra_attributes|merge([{'name': name, 'value': value}]) %}
 {% endfor %}
 
 {% include '@ecl-twig/button' with {
   label: attributes.value,
   variant: variant|default('primary'),
-  extra_attributes: extraAttributes
+  extra_classes: extra_classes|default(''),
+  extra_attributes: extra_attributes
 } %}
--- a/templates/form/select.html.twig
+++ b/templates/form/select.html.twig
@@ -64,6 +64,8 @@
     icon_path: ecl_icon_path,
     id: attributes.id,
     name: attributes.name,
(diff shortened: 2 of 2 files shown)


templates/form/input--submit.html.twig
```diff
@@ -6,13 +6,14 @@
  * @see ./core/themes/stable/templates/form/input.html.twig
  */
 #}
-{% set extraAttributes = [] %}
+{% set extra_attributes = extra_attributes|default([]) %}
 {% for name,value in attributes %}
-  {% set extraAttributes = extraAttributes|merge([{'name': name, 'value': value}]) %}
+  {% set extra_attributes = extra_attributes|merge([{'name': name, 'value': value}]) %}
```
22Alexandra, line 11: These changes are not necessary because the way the template works now is passing the `extraAttributes` built from the existing `attributes` of the form element.

For example, these are the existing attributes of an element:
template preprocess
And these are the attributes rendered:
input--submit template

So, the desired extra attributes can safely be passed through the preprocess on the attributes array.

templates/form/input--submit.html.twig
```diff
@@ -6,13 +6,14 @@
...
 {% endfor %}
 
 {% include '@ecl-twig/button' with {
   label: attributes.value,
   variant: variant|default('primary'),
-  extra_attributes: extraAttributes
+  extra_classes: extra_classes|default(''),
```
22Alexandra, line 17: Indeed, passing extra classes was missing. It's going to be fixed in #1630 and the extra classes can be passed through the preprocess on the `class` array of the `attributes` array.

templates/form/select.html.twig
```diff
@@ -64,6 +64,8 @@
     icon_path: ecl_icon_path,
     id: attributes.id,
     name: attributes.name,
+    extra_classes: extra_classes_input|default(''),
+    extra_attributes: extra_attributes_input|default([]),
```
22Alexandra, line 68: Will be added in #1630.

22Alexandra (commented):
Thank you for the contribution, the changes will be done in #1630.

22Alexandra:
The issue has been fixed and closed. The fix has been included in 5.0.0-beta14.

PR #1639, opened 2025-09-08
EWPP-5805: Use composer-dependent-patches and unlock drupal core.

nagyad:


intelektron (dismissed):

composer.json
```diff
@@ -33,7 +33,7 @@
         "drupal/emr": "^1.3",
         "drupal/entity_browser": "2.13",
         "drupal/entity_reference_revisions": "~1.11",
-        "drupal/extra_field": "~3",
+        "drupal/extra_field": "3.0.0-beta2",
```
22Alexandra: Not that it matters much but this should be beta1.
Tests failing in this PR will be fixed after https://github.com/openeuropa/oe_theme/pull/1641 is merged and this PR gets rebased.

composer.json
```diff
@@ -116,7 +116,8 @@
             "Explicit requirement of lullabot/mink-selenium2-driver v1.6.0 to ensure dragTo() keeps working in tests. Aliased as 1.7.3 because core-dev requires it on that version.",
             "autoload-dev: explicitly mapping EntityReferenceFieldCreationTrait to prevent failures.",
             "Patch for drupal/driver refers to https://github.com/jhedstrom/DrupalDriver/pull/235",
-            "Locked drupal/entity_browser to 2.13 to version because newer release is breaking media avportal entity browser."
+            "Locked drupal/entity_browser to 2.13 to version because newer release is breaking media avportal entity browser.",
```
22Alexandra: The drupal/entity_browser was unlocked in the oe_content PR. So let's unlock the module here as well.

22Alexandra (changes requested):

composer.json
nagyad: That should not be the case, because we have a ticket for it, I will revert in oe_content.

22Alexandra (dismissed):

22Alexandra (dismissed):

intelektron (dismissed):

intelektron (approved):

sergepavle (approved):

PR #1646, opened 2025-09-25
EWPP-6046: Move CI to Gitlab CI.

hernani:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


runner.yml.dist
```diff
@@ -14,75 +14,67 @@ drupal:
...
+                'namespace' => 'Drupal\sparql_entity_storage\Driver\Database\sparql',
+                'driver' => 'sparql',
+              ];
   setup:phpunit9:
     - { task: "process", source: "phpunit9.xml.dist", destination: "phpunit.xml" }
   setup:behat:
     - { task: "process", source: "behat.yml.dist", destination: "behat.yml" }
-
+  toolkit:code-review:
+      - { task: "exec", command: "./vendor/bin/grumphp run --verbose" }
```
donquixote:
```suggestion
    - { task: "exec", command: "./vendor/bin/grumphp run --verbose" }
```
:)

runner.yml.dist
```diff
@@ -14,75 +14,67 @@ drupal:
...
+        }
+        $settings['file_scan_ignore_directories'] = array('vendor', 'build', 'node_modules', 'bower_components');
+        $settings['file_private_path'] = 'sites/default/files/private';
+        $settings['extension_discovery_scan_tests'] = TRUE;
+        $databases['sparql_default']['default'] = [
+          'host' => 'sparql',
+          'port' => '8890',
+          'namespace' => 'Drupal\sparql_entity_storage\Driver\Database\sparql',
+          'driver' => 'sparql',
+        ];
```
donquixote, line 40: Indentation here seems off.

runner.yml.dist
```diff
@@ -14,75 +14,67 @@ drupal:
...
+    - { task: "exec", command: "./vendor/bin/drush config-set system.theme default oe_theme -y" }
+    - { task: "exec", command: "./vendor/bin/drush config-set system.theme admin claro -y" }
+    - { task: "exec", command: "./vendor/bin/drush config-set node.settings use_admin_theme 1 -y" }
+    - { task: "exec", command: "./vendor/bin/drush config-set user.settings register visitors -y" }
+    - { task: "exec", command: "./vendor/bin/drush cr" }
+  additional_settings: |
+        if (isset($config['system.performance']['cache']['page']['max_age'])) {
+          unset($config['system.performance']['cache']['page']['max_age']);
+        }
+        $settings['file_scan_ignore_directories'] = array('vendor', 'build', 'node_modules', 'bower_components');
```
donquixote, line 32: Is this old-school array syntax the result of a var_export()?

docker-compose.yml
```diff
@@ -35,16 +41,6 @@ services:
...
-  node:
-    image: node:22.11.0
-    user: "node"
-    working_dir: /home/node/app
-    environment:
-      - NODE_ENV=${NODE_ENV}
-      - SHELL=bash
-    volumes:
-      - ./:/home/node/app
-    command: "sleep infinity"
```
donquixote, line 47: We are removing the 'node' container, but the README.md still talks about `docker-compose exec -u node node npm install`.
Are we now relying on npm in the web container? If that is sufficient, why did we need this separate container in the past?

runner.yml.dist
22Alexandra: Seems fine to me.

docker-compose.yml
22Alexandra: I agree we should update README to reflect the changes and how oe_theme is supposed to be built from now on.

22Alexandra (commented):
Looking good and the artifacts seem to get correctly created. Let's update the readme.

docker-compose.yml
donquixote: I think there are two changes:
1. In the general setup we no longer need to call the npm steps, they are part of `drupal:site-setup` which is called with composer install and composer update.
2. We could still mention the steps for front-end development.

runner.yml.dist
```diff
@@ -101,3 +93,38 @@ release:
         project: oe_theme
         version: ${release.version}
         datestamp: ${release.timestamp}
+
+toolkit:
+  npm:
+    theme-task-runner: ecl-builder
```
donquixote, line 99: What does this line do? Is it needed?
I see we build npm assets in `site-setup`, that should be sufficient, or not?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
+name: ci
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  automated-tests:
```
donquixote: This job should be named "release" or similar.
```suggestion
  release:
```
or
```suggestion
  build-release-assets:
```

.github/workflows/ci.yml
donquixote, line 1: This file should be named like `release.yml`, not `ci.yml`.

If in the future we move actual tests to github actions, they can live in a separate file `ci.yml`.

donquixote: On the other hand.. there is a lot of overlap with what would be in a CI pipeline for tests.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Run grumphp
+        run: docker compose exec -T web ./vendor/bin/grumphp run
+
+      - name: Components build
+        run: docker compose exec -T web bash ./scripts/build-ecl-pipeline.sh
+
+      - name: Create release archive
+        run: |
+          echo "Tag name from GITHUB_REF_NAME: $GITHUB_REF_NAME"
+          echo "Tag name from github.ref_name: ${{  github.ref_name }}"
```
donquixote, line 49: Did you notice any difference between the two expressions?
Or do you suspect they could be different in some scenarios?
Otherwise, I'd say the env var $GITHUB_REF_NAME is easier to write.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+          docker compose exec -T web apt-get install zip -y
+
+      - name: Update composer
+        run: docker compose exec -T web composer self-update --2
+
+      - name: Build dependencies
+        run: docker compose exec -T web composer require drupal/core:~${{ matrix.core_version }} drupal/core-composer-scaffold:~${{ matrix.core_version }} --update-with-all-dependencies --ansi --no-progress --no-interaction
+
+      - name: Run grumphp
+        run: docker compose exec -T web ./vendor/bin/grumphp run
```
donquixote, line 42: Should we run grumphp for a release? Is this a compromise so we have at least _some_ validation?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
+name: ci
+on:
+  push:
+    tags:
+      - '*'
```
donquixote, line 5: This will work if you create a release and the tag through the UI.

However, if you push a release tag before the release exists, this will run and then the "gh release upload" will fail because there is no release.
Later when you create the release in the UI, using the existing tag, nothing will happen.

It is possible to auto-create a (draft) release with github actions. I found some workflows with code search, but they look quite complicated. Also not sure if we really want to create a release for _every_ tag, seems a bit too trigger-happy.

There is an alternative trigger for release published:
```
on:
  release:
    types: [published]
```
(there is also type = created, but that triggers when you save as draft. Note that draft is not the same as "pre-release", a "pre-release" is fully published. Therefore I would use published trigger here, not created.)

But this would not renew the assets if we force-push a tag. (which we may want to do for draft releases sometimes)

What we could do is trigger on release publish, _and_ on tag push, and for the latter we abort if no release exists.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: |
+          echo "Tag name from GITHUB_REF_NAME: $GITHUB_REF_NAME"
+          echo "Tag name from github.ref_name: ${{  github.ref_name }}"
+          docker compose exec -T web ./vendor/bin/run toolkit:release:create-archive --name=oe_theme --tag=${{  github.ref_name }}
+          docker compose exec -T web ./vendor/bin/run toolkit:release:create-archive --name=oe_theme --tag=${{  github.ref_name }} --zip
+
+      - name: Upload files to release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh release upload "${{  github.ref_name }}" "oe_theme-${{  github.ref_name }}.tar.gz" "oe_theme-${{  github.ref_name }}.zip"
```
donquixote: Where does this extra space come from?
```suggestion
        run: gh release upload "${{ github.ref_name }}" "oe_theme-${{ github.ref_name }}.tar.gz" "oe_theme-${{ github.ref_name }}.zip"
```

donquixote: From what I read (chatgpt), calling "gh release upload" repeatedly will add more and more assets, but not remove or replace the old assets.
This would happen if we force-push the release tag for an existing release, which we might want to do sometimes for draft releases.

runner.yml.dist
donquixote, line 40: This is what I mean.. unless there is a special code style rule for this kind of text block.
```suggestion
  additional_settings: |
    if (isset($config['system.performance']['cache']['page']['max_age'])) {
      unset($config['system.performance']['cache']['page']['max_age']);
    }
    $settings['file_scan_ignore_directories'] = array('vendor', 'build', 'node_modules', 'bower_components');
    $settings['file_private_path'] = 'sites/default/files/private';
    $settings['extension_discovery_scan_tests'] = TRUE;
    $databases['sparql_default']['default'] = [
      'host' => 'sparql',
      'port' => '8890',
      'namespace' => 'Drupal\sparql_entity_storage\Driver\Database\sparql',
      'driver' => 'sparql',
    ];
```

runner.yml.dist
```diff
@@ -14,75 +14,67 @@ drupal:
...
+      file: "build/sites/default/settings.testing.php"
+      text: |
+          <?php
+              $settings['extension_discovery_scan_tests'] = TRUE;
+              $databases['sparql_default']['default'] = [
+                'host' => 'sparql',
+                'port' => '8890',
+                'namespace' => 'Drupal\sparql_entity_storage\Driver\Database\sparql',
+                'driver' => 'sparql',
+              ];
```
donquixote, line 69:
```suggestion
      text: |
        <?php

        $settings['extension_discovery_scan_tests'] = TRUE;
        $databases['sparql_default']['default'] = [
          'host' => 'sparql',
          'port' => '8890',
          'namespace' => 'Drupal\sparql_entity_storage\Driver\Database\sparql',
          'driver' => 'sparql',
        ];
```
This would make more sense to me.
Or is there a special reason I don't know about?

.gitignore
```diff
@@ -4,6 +4,7 @@ build
 !/modules/oe_theme_helper/js/build
 npm-debug.log
 grumphp.yml
+junit-export
```
donquixote, line 7: Is this needed?
I suppose it would only be needed if these files get generated in a local installation.
For me, in oe_bootstrap_theme, I never saw this directory being created. Maybe I need to run phpunit in a specific way?

.github/workflows/ci.yml
donquixote: Actually.. bad info from AI (chatgpt).
I tried this, and also looked at the docs here https://cli.github.com/manual/gh_release_upload

`gh release upload` will fail with "asset under the same name already exists: ...", if the asset already exists.

However, we can use `--clobber` parameter to overwrite.

```suggestion
        run: gh release upload --clobber "${{  github.ref_name }}" "oe_theme-${{  github.ref_name }}.tar.gz" "oe_theme-${{  github.ref_name }}.zip"
```
and multiline can be nice, and cleaning the spaces.
```suggestion
        run: >
          gh release upload --clobber "${{ github.ref_name }}"
          "oe_theme-${{ github.ref_name }}.tar.gz"
          "oe_theme-${{ github.ref_name }}.zip"
```

.gitignore
hernani: Only now

toolkit/src/TaskRunner/Commands/BatchCommands.php
donquixote, line 1: Is there a convention to put toolkit commands in a directory with this name?
And where does the namespace come from, `oe_theme_commands`?

I have nothing against it, just want to know if it is a convention imposed by toolkit or a custom choice.

toolkit/src/TaskRunner/Commands/BatchCommands.php
```diff
@@ -0,0 +1,71 @@
...
+
+    $this->say('Trying to run tests that are not assigned to any batch...');
+    $phpunit_bin = $this->getBin('phpunit');
+    $working_directory = $this->getWorkingDir();
+
+    $unassigned_tests_result_file = $working_directory . '/junit-export/phpunit-unassigned-tests.xml';
+    $collection->addTask($this->taskExec($phpunit_bin)
+      ->option('log-junit', $unassigned_tests_result_file)
+      ->option('exclude-group', implode(',', $batches))
+    );
```
donquixote, line 37: Should this "leftovers" batch not be run in parallel with the named batch?

donquixote: Oh, I can answer this myself.
Having "leftover" tests (outside of the named groups) is considered a mistake.

sergepavle (approved):

nagyad (approved):

PR #1650, opened 2025-10-06
EWPP-6094: Add new variants to List with illustration pattern.

22Alexandra:


patches.lock.json
nagyad: Please add this file to the gitignore.

nagyad (changes requested):

sergepavle (dismissed):

sergepavle (approved):

patches.lock.json
22Alexandra: Done.

PR #1659, opened 2025-11-18
EWPP-6280: Update to ECL 5.0.0-alpha.22.2.

22Alexandra:


intelektron (dismissed):

oe_theme.theme
```diff
@@ -2563,6 +2563,17 @@ function oe_theme_preprocess_select(&$variables) {
  * Implements hook_page_attachments_alter().
  */
 function oe_theme_page_attachments_alter(array &$attachments) {
+  // Add the CDN file required for ECL datepicker to work.
+  $attachments['#attached']['html_head'][] = [
+    [
+      '#tag' => 'script',
+      '#attributes' => [
+        'type' => 'module',
+        'src' => 'https://cdn.jsdelivr.net/npm/@duetds/date-picker@1.4.0/dist/duet/duet.esm.js',
```
intelektron, line 2571: I explored some possibilities of using libraries here, but neither `weight` nor `header: true` options were working for this thing.

We should probably cache it locally somehow, to avoid issues with pipelines and avoid using external scripts inside of ewcms.

oe_theme.theme
```diff
@@ -343,13 +343,13 @@ function oe_theme_preprocess_input__checkbox(array &$variables): void {
...
+  $php_date_format = $variables['element']['#date_date_format'] ?? 'd-m-Y';
   $format_map = [
     // @todo find a cleverer way of mapping PHP date formats to the expected
     // ECL ones.
-    'Y-m-d' => 'YYYY-MM-DD',
+    'd-m-Y' => 'DD-MM-YYYY',
   ];
-  $ecl_format = $format_map[$php_date_format] ?? 'YYYY-MM-DD';
+  $ecl_format = $format_map[$php_date_format] ?? 'DD-MM-YYYY';
   $variables['#attached']['library'][] = 'oe_theme/ecl_datepicker';
```
intelektron: We removed this lib, right?

22Alexandra: Indeed.

oe_theme.theme
22Alexandra: I would postpone this for a follow up ticket in ewcms after the release, if that's ok.

tests/src/FunctionalJavascript/ParagraphsTest.php
```diff
@@ -0,0 +1,86 @@
...
+
+    // Rebuild the ui_pattern definitions to collect the ones provided by
+    // oe_theme itself.
+    \Drupal::service('plugin.manager.ui_patterns')->clearCachedDefinitions();
+  }
+
+  /**
+   * Test Accordion item paragraph form.
+   */
+  public function testAccordionItemParagraph(): void {
```
22Alexandra, line 55: This test had to be ported to JS because the new ECL version uses Duet.js for date pickers and they only work using JS, otherwise the date would not be submitted in the form.

templates/patterns/page_header/page_header.ui_patterns.yml
```diff
@@ -51,8 +51,8 @@ page_header:
       preview: "page-header-pattern-background-preview.jpeg"
     overlay:
       type: "text"
-      label: "Overlay"
-      description: "Optional overlay on top of background image (can be 'light', 'dark'). Only used on EC sites"
+      label: "Overlay - Deprecated and will be removed in 5.x stable version."
+      description: "Optional overlay on top of background image (can be 'light', 'dark'). Not used in ECL 5.x anymore."
```
sergepavle, line 55: Should we do anything on EWPP side? The select box for selecting an overlay no longer works. Do we already have a follow up ticket to fix that (hide this select box or some other adjustments)?

22Alexandra: There is a new feature for R39 that will rework the Page header options and overlay will be removed entirely in ewcms.
You can find additional info [here](https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?pageId=2191951878&spaceKey=EWPP&title=New%2BVisual%2BIdentity%2B2025%2B-%2BNew%2Bpage%2Bheader%2Boptions) and new tickets will be created soon.

sergepavle: It is fine.

sergepavle (approved):

intelektron (approved):

PR #1678, opened 2026-01-19
EWPP-6456: Create the SDC for each pattern.

22Alexandra:


components/text_featured_media/text_featured_media.twig
```diff
@@ -0,0 +1,135 @@
...
+{% if link %}
+  {% set link_highlighted = 'featured' in variant ? true : false %}
+  {% set _icon = get_link_icon(link.path, 'm') %}
+  {% if _icon['name'] == 'corner-arrow' %}
+    {% set _icon = _icon|merge({
+      'name': 'arrow-left',
+      'transform': 'flip-horizontal',
+    }) %}
+  {% endif %}
+  {{ dump(_icon['name']) }}
```
nagyad: This seems to some debug code.

components/banners/banner_video.component.yml
```diff
@@ -0,0 +1,9 @@
+name: 'Banner: video'
+description: 'Is a large web banner video, prominently placed on a web page, generally at the top.'
+slots:
+  video:
+    title: Video
+    description: 'Banner video: uses the media container parameters as defined by the ECL'
+  description:
+    title: Description
+    description: 'Banner description'
```
nagyad, line 9: Aren't we missing extra_classes and extra_attributes that are being used in the banner_video.twig template?

components/link/link.story.twig
```diff
@@ -0,0 +1,11 @@
+{% for _variant in ['standalone', 'cta'] %}
+  <h2 class="ecl-u-type-heading-2">
+    Variant: {{ _variant|render }}
+  </h2>
+  {% include 'oe_theme:link' with {
+    text: 'A simple link',
+    url: '#',
+    external_link: true,
```
nagyad, line 9: We loop through the _variant here, but we never pass it to the 'oe_theme:link', where different variants are used. Is that on purpose?

nagyad (changes requested):
I see we are mixing the story usage here. We should use the same, most components use .story.twig (Twig-based), some still use .preview.story.yml (YAML-based): gallery, file_teaser, file.

22Alexandra:
> I see we are mixing the story usage here. We should use the same, most components use .story.twig (Twig-based), some still use .preview.story.yml (YAML-based): gallery, file_teaser, file.

True, I forgot to open a ticket for it. These patterns are using custom object types which are particular cases and do not work as the other "slots", they might need props. So, they're basically not working in this PR.
Created EWPP-6499 to handle this SDCs.

sergepavle (dismissed):

components/banners/banner_video.component.yml
22Alexandra: We didn't have those fields before, the structure was kept the same.

components/link/link.story.twig
22Alexandra: No, I fixed it.

nagyad (approved):

sergepavle (approved):

PR #1699, opened 2026-03-02
EWPP-6494: Replace usage of patterns with SDCs in templates.

22Alexandra:


components/link/link.twig
```diff
@@ -14,8 +14,7 @@
     label: text,
     path: url,
     external: external_link,
-    icon_path: ecl_icon_path
   },
   extra_attributes: extra_attributes,
-  extra_classes: classes
+  extra_classes: extra_classes
```
sergepavle, line 19: Is it fine like that?
In previous implementation for ui_patterns in \oe_theme_preprocess_pattern_link we had passing attributes from native link object options. It is hard to find example where we rendering link.
Changes in modules/oe_theme_content_event/src/Plugin/ExtraField/Display/OnlineDescriptionExtraField.php is good example of such case when passed URL object attributes is omitted.

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.post_update.php
```diff
@@ -29,3 +29,10 @@ function oe_theme_content_call_proposals_post_update_00001(): void {
   $display = $storage->createFromStorageRecord($display_values);
   $display->save();
 }
+
+/**
+ * Uninstall the ui_patterns_field_group module.
+ */
+function oe_theme_content_call_proposals_post_update_00002(): void {
+  \Drupal::service('module_installer')->uninstall(['ui_patterns_field_group']);
```
sergepavle, line 37: Don't know if this is safe unless oe_theme will have new major release with breaking changes.
I suppose other projects based on oe_theme can implement own ui_patterns.

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
```diff
@@ -238,7 +238,8 @@ public function testEclDatePicker(): void {
    * Tests that contextual navigation pattern is rendered properly.
    */
   public function testContextNavPattern(): void {
-    $this->drupalGet('/oe_theme_js_test/ui_patterns/context_nav');
+    $this->drupalGet('/oe_theme_js_test/single_directory_components/context_nav');
+    file_put_contents('test.html', $this->getSession()->getPage()->getHtml());
```
sergepavle: Debugging leftover

tests/src/Kernel/Patterns/LinkPatternRenderingTest.php
```diff
@@ -21,17 +21,30 @@ class LinkPatternRenderingTest extends AbstractKernelTestBase {
    * @throws \Exception
    */
   public function testLinkPatternRendering() {
+    $url = Url::fromUserInput('/node/add', [
+      'attributes' => [
+        'class' => ['foo'],
+        'foo' => 'bar',
+      ],
+    ]);
+    // Extract attributes from the URL object, as the SDC component expects
```
sergepavle, line 30: Again if this is new approach of work with link component - it is fine and you may ignore previous comment regarding omitted `class` attribute from URL object (in background).

oe_theme.theme
```diff
@@ -2397,7 +2398,7 @@ function oe_theme_preprocess_node__oe_organisation__teaser(array &$variables): v
   if (!empty($variables['content']['extra_field_oe_theme_content_organisation_teaser_details'])) {
     $variables['teaser_details'] = [];
     foreach (Element::children($variables['content']['extra_field_oe_theme_content_organisation_teaser_details']) as $index) {
-      $variables['teaser_details'][] = $variables['content']['extra_field_oe_theme_content_organisation_teaser_details'][$index]['#fields']['items'];
+      $variables['teaser_details'][] = $variables['content']['extra_field_oe_theme_content_organisation_teaser_details'][$index]['#props']['items'];
     }
   }
 }
```
sergepavle, line 2403: In oe_theme.theme I see many `oe_theme_preprocess_pattern_*` hook implementation. As I understand it is left overs and can be removed.

sergepavle (changes requested):

modules/oe_theme_content_call_proposals/oe_theme_content_call_proposals.post_update.php
22Alexandra: Yes, oe_theme will have a new branch 6.x. The SDC epic is a breaking change and will be released in a new major.

components/link/link.twig
22Alexandra: The problem is SDC templates cannot be preprocessed, so the only way I've found was to pass extra_classes like we do for other cases.

tests/src/Kernel/Patterns/LinkPatternRenderingTest.php
22Alexandra: Explained in the previous comment why this change.

sergepavle (dismissed):

modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/SocialMediaIconsListFormatter.php
```diff
@@ -37,11 +37,13 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
     $elements = parent::viewElements($items, $langcode);
     foreach ($items as $delta => $item) {
       $pattern = [
-        '#type' => 'pattern',
-        '#id' => 'social_icon',
-        '#fields' => [
-          'service' => $item->link_type,
+        '#type' => 'component',
+        '#component' => 'oe_theme:social_icon',
+        '#slots' => [
```
nagyad: `label` placed in `#slots` instead of `#props`. 
The social_icon component declares `label` as a prop (line 14 of social_icon.component.yml), not a slot. The formatter passes it via `#slots`, which means it will be silently discarded, and every social icon will render with an empty link label.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/DescriptionExtraField.php
```diff
@@ -111,13 +111,15 @@ public function viewElements(ContentEntityInterface $entity) {
       return $empty_build;
     }
 
-    $build['#fields']['title'] = $title;
-    $build['#fields']['text'] = $text;
+    $build['#slots']['title'] = $title;
+    if (!empty($text)) {
+      $build['#slots']['text'] = $text;
+    }
```
nagyad: `title`, `text`, and `caption` placed in` #slots` instead of `#props`. The text_featured_media component defines `title`, `text`, and `caption` as props (lines 31-54 of text_featured_media.component.yml). The PHP code passes them via `#slots` on lines 114, 116, and 249, so these values will be empty when the event description renders.

modules/oe_theme_content_event/src/Plugin/ExtraField/Display/ProgrammeExtraField.php
```diff
@@ -126,11 +126,13 @@ public function viewElements(ContentEntityInterface $entity) {
...
     $build = [
-      '#type' => 'pattern',
-      '#id' => 'timeline',
-      '#fields' => [
+      '#type' => 'component',
+      '#component' => 'oe_theme:timeline',
+      '#slots' => [
+        'title' => ['#markup' => $this->getLabel()],
+      ],
+      '#props' => [
```
nagyad, line 131: `title` placed in `#slots` instead of `#props`. The timeline component defines `title` as a prop (line 38 of timeline.component.yml). The PHP code passes it via `#slots` on line 132, so the timeline heading will silently disappear.

modules/oe_theme_helper/oe_theme_helper.post_update.php
```diff
@@ -810,3 +810,13 @@ function oe_theme_helper_post_update_50004(): void {
...
 }
+
+/**
+ * Replace the ui_patterns module with sdc_component_library.
+ */
+function oe_theme_helper_post_update_60001(): void {
+  \Drupal::service('module_installer')->uninstall(['ui_patterns_library']);
+  \Drupal::service('module_installer')->uninstall(['ui_patterns']);
+  // Install sdc_component_library.
+  \Drupal::service('module_installer')->install(['sdc_component_library']);
```
nagyad, line 821: This function uninstalls `ui_patterns_library` then ui_patterns, but `ui_patterns_field_group` (which depends on `ui_patterns`) is uninstalled in separate post_update hooks in other modules. Drupal doesn't guarantee cross-module post_update execution order, so it could fail if oe_theme_helper's hook runs first.

nagyad (changes requested):
The PR is massive; it wasn't an easy feat to review this. Next time we need to think about how we could split this up better.

Nicely done, good work.

modules/oe_theme_helper/oe_theme_helper.post_update.php
22Alexandra: I don't think oe_theme_helper post_update can run before the others, unless a project doesn't have the other submodules installed. In that case, the `ui_patterns_field_group` should not be installed either.
post updates

nagyad (approved):

sergepavle (approved):

PR #1712, opened 2026-04-01
EWPP-6539: Add ECL tooltip.

gorsky87:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/tests/modules/oe_theme_js_test/src/Form/TooltipTestForm.php
+++ b/tests/modules/oe_theme_js_test/src/Form/TooltipTestForm.php
@@ -1,0 +1,177 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\oe_theme_js_test\Form;
+
+use Drupal\Core\Form\FormInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+
+/**
+ * Provides an example form with tooltip variants.
+ */
+class TooltipTestForm implements FormInterface {
+
+  use StringTranslationTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId(): string {
+    return 'oe_theme_js_tooltip_test';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state): array {
+    $tooltip = [
+      'data-ecl-tooltip' => 'tooltip',
+    ];
+
+    $tooltip_inverted = [
+      'data-ecl-tooltip-inverted' => 'inverted tooltip',
+    ];
+
+    $form['info'] = [
+      '#type' => 'markup',
+      '#markup' => $this->t('Each field has normal + inverted tooltip version.'),
+    ];
+
+    $form['textfield_normal'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Normal tooltip (Textfield)'),
+      '#attributes' => $tooltip,
+    ];
+
+    $form['textfield_inverted'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Inverted tooltip (Textfield)'),
+      '#attributes' => $tooltip_inverted,
+    ];
+
+    $form['textarea_normal'] = [
+      '#type' => 'textarea',
+      '#title' => $this->t('Normal tooltip (Textarea)'),
+      '#attributes' => $tooltip,
+      '#rows' => 2,
+    ];
+
+    $form['textarea_inverted'] = [
+      '#type' => 'textarea',
+      '#title' => $this->t('Inverted tooltip (Textarea)'),
+      '#attributes' => $tooltip_inverted,
+      '#rows' => 2,
+    ];
+
+    $form['email_normal'] = [
+      '#type' => 'email',
+      '#title' => $this->t('Normal tooltip (Email)'),
+      '#attributes' => $tooltip,
+    ];
+
+    $form['email_inverted'] = [
+      '#type' => 'email',
+      '#title' => $this->t('Inverted tooltip (Email)'),
+      '#attributes' => $tooltip_inverted,
+    ];
+
+    $form['number_normal'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Normal tooltip (Number)'),
+      '#attributes' => $tooltip,
+    ];
+
+    $form['number_inverted'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Inverted tooltip (Number)'),
+      '#attributes' => $tooltip_inverted,
+    ];
+
+    $form['select_normal'] = [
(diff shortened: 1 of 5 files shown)


templates/form/input--checkbox.html.twig
```diff
@@ -12,6 +12,19 @@
  * @see ./core/themes/stable/templates/form/input.html.twig
  */
 #}
+{% set _input_extra_attributes = extra_attributes_input|default([]) %}
+
+{% set _input_extra_attributes = extra_attributes_input|default([]) %}
```
22Alexandra: Probably no need to do this twice.

tests/modules/oe_theme_js_test/src/Form/TooltipTestForm.php
```diff
@@ -0,0 +1,189 @@
...
+      '#type' => 'date',
+      '#title' => $this->t('Normal tooltip (Date)'),
+      '#attributes' => $tooltip,
+    ];
+
+    $form['date_inverted'] = [
+      '#type' => 'date',
+      '#title' => $this->t('Inverted tooltip (Date)'),
+      '#attributes' => $tooltip_inverted,
+    ];
```
22Alexandra: This does not work at all, so why adding it to the form? Please remove it.
Don't we need to add the extra_attributes in `input--date.html.twig`?

templates/form/input--checkbox.html.twig
```diff
@@ -25,6 +38,7 @@
         name: attributes.name,
         label: title,
         icon_path: ecl_icon_path,
+        _input_extra_attributes: _input_extra_attributes,
```
22Alexandra: This is wrong, I don't think it even works. ECL templates are not looking for such parameter `_input_extra_attributes` and the items array shouldn't contain the extra_attributes but the they should be passed to the template directly inside the `extra_attributes` that ECL uses.

templates/form/input--checkbox.html.twig
```diff
@@ -12,6 +12,19 @@
  * @see ./core/themes/stable/templates/form/input.html.twig
  */
 #}
+{% set _input_extra_attributes = extra_attributes_input|default([]) %}
```
22Alexandra: Where is this variable `extra_attributes_input` coming from or who's setting it?
Why are we not using the `extra_attributes` ECL parameter like in other templates that render ECL components?

22Alexandra (changes requested):

templates/form/input--checkbox.html.twig
gorsky87: in
form-group.html.twig

 {% set _input_extra_attributes = _input.extra_attributes|default([]) %}
 
 extra attributes render tooltip in field wrapper, we have to pass it to input and then it's rendeer like extra atribute

tests/modules/oe_theme_js_test/src/Form/TooltipTestForm.php
gorsky87: its not working even when we add    extra_attributes or  _input_extra_attributes,

templates/form/input--checkbox.html.twig
```diff
@@ -12,6 +12,17 @@
  * @see ./core/themes/stable/templates/form/input.html.twig
  */
 #}
+{% set _input_extra_attributes = _input_extra_attributes|default([]) %}
```
22Alexandra: I don't know where this variable comes from and how/why this line makes any sense.

templates/form/input--checkbox.html.twig
```diff
@@ -25,13 +38,14 @@
...
         label: title,
         icon_path: ecl_icon_path,
+        extra_attributes: _input_extra_attributes,
         disabled: attributes.disabled is defined,
       }
     ]
   },
   invalid: invalid,
   disabled: attributes.disabled is defined,
-  extra_attributes: input_attributes|default([]),
```
intelektron, line 34: What happens to `input_attributes` here? It looks like they're lost

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
```diff
@@ -234,6 +234,63 @@ public function testEclDatePicker(): void {
...
+    JS;
+
+      $this->getSession()->executeScript($script);
+      $this->getSession()->wait(500);
+
+      $visible = FALSE;
+
+      foreach ($this->getSession()->getPage()->findAll('css', '.ecl-tooltip') as $tooltip) {
+        if ($this->getSession()->getDriver()->isVisible($tooltip->getXpath())) {
+          $visible = TRUE;
```
intelektron, line 291: Shouldn't we check only the tooltip that is being hovered in this test case?

Otherwise any visible tooltip is considered as passed test

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
```diff
@@ -234,6 +234,63 @@ public function testEclDatePicker(): void {
...
+        el.dispatchEvent(new MouseEvent('mouseover', {
+          bubbles: true,
+          cancelable: true,
+          view: window
+        }));
+      })();
+    JS;
+
+      $this->getSession()->executeScript($script);
+      $this->getSession()->wait(500);
```
intelektron: Could we actually wait for the tooltip here? 500ms seems like arbitrary value

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
```diff
@@ -234,6 +234,63 @@ public function testEclDatePicker(): void {
     $this->assertSession()->pageTextContains('Date 1 is 10 May 2020');
   }
 
+  /**
+   * Test ECL tooltip.
+   */
+  public function testEclTooltip(): void {
+    $this->drupalGet('/oe_theme_js_test/tooltip');
+
+    $test_cases = [
```
intelektron, line 243: Don't we test it for radio? I can see the tooltip there

Image

templates/form/input--checkbox.html.twig
```diff
@@ -12,6 +12,19 @@
  * @see ./core/themes/stable/templates/form/input.html.twig
  */
 #}
+
+{% set _source_extra_attributes = extra_attributes|default([]) %}
```
intelektron: Are these changes taken into account? I noticed your discussion about tooltips not working for checkboxes, and I can't see any tooltip in the test form.

Image

tests/src/FunctionalJavascript/JavascriptBehavioursTest.php
gorsky87: yes but gererally, behavior is wrong

intelektron (approved):

22Alexandra (approved):

PR #1713, opened 2026-04-07
EWPP-6686: Add override of endpoint url for webtools icon provider service.

sergepavle:
…rvice.

### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_theme_helper/src/WebtoolsIconsProvider.php
```diff
@@ -168,24 +177,32 @@ public function getCacheTags(): array {
...
   protected function downloadWebtoolsIcons(): array {
+    $logos_url = Settings::get('webtools_wshape.url') ?: static::ICONS_URL;
     try {
-      $json_response = $this->httpClient->get(static::ICONS_URL);
+      /** @var \Drupal\Core\Url $url */
+      $url = Url::fromUri($logos_url);
+
+      if ($this->iconsFamily !== NULL) {
+        $url->setOption('query', ['family' => $this->iconsFamily]);
+      }
```
22Alexandra: Since `$iconsFamily` is a public property, when a provider sets it to `logos` it will only fetch those icons, which makes sense but if this is done only for the `SpecialDayIconsProvider` from ewcms, I would rather pass it as a parameter to the method that needs it. 

Otherwise, if we go with this approach, document the property in the interface, so it's clear that is meant for later consumers (such as ewcms) and maybe the property should be injected via the service definition.

22Alexandra (changes requested):

modules/oe_theme_helper/src/WebtoolsIconsProvider.php
```diff
@@ -168,24 +177,32 @@ public function getCacheTags(): array {
    *   The array of icons.
    */
   protected function downloadWebtoolsIcons(): array {
+    $logos_url = Settings::get('webtools_wshape.url') ?: static::ICONS_URL;
```
22Alexandra: It fetches icons, not just logos.
```suggestion
    $icons_url = Settings::get('webtools_wshape.url') ?: static::ICONS_URL;
```

22Alexandra (dismissed):

22Alexandra (dismissed):

PR #1722, opened 2026-04-24
EWPP-6790: Update to ECL 5.0.0-RC3.

22Alexandra:


templates/patterns/navigation_menu/navigation_menu.ui_patterns.yml
```diff
@@ -44,7 +44,7 @@ navigation_menu:
       type: "numeric"
       label: "Lines limit for first level items"
       description: "Number of maximum lines for each first level item label. It defaults to 2, set it to zero to remove this behavior."
-      preview: 1
+      preview: 2
```
22Alexandra, line 47: This change was needed to fix the pattern's preview until FRONT-5307 fixes the rendering for max_lines: 1.

tests/src/Kernel/fixtures/rendering.yml
```diff
@@ -4712,11 +4712,11 @@
       'li.ecl-mega-menu__item--has-children:nth-child(2) ul.ecl-mega-menu__sublist li.ecl-mega-menu__subitem:nth-child(1) div.ecl-mega-menu__mega--level-2 div.ecl-mega-menu__featured ul.ecl-mega-menu__featured-list li.ecl-mega-menu__featured-list__item:nth-child(3) picture.ecl-mega-menu__featured-picture img.ecl-mega-menu__featured-image[src="https://loremflickr.com/1000/500/arch"][alt="alternative text of the 3rd featured item"]': 1
       'button.ecl-button.ecl-button--tertiary.ecl-mega-menu__open.ecl-button--icon-only .ecl-button__container:nth-child(1) span:nth-child(1).ecl-icon--hamburger': 1
       'button.ecl-button.ecl-button--tertiary.ecl-mega-menu__open.ecl-button--icon-only .ecl-button__container:nth-child(1) span:nth-child(2).ecl-icon--close': 1
-      'li.ecl-mega-menu__item:nth-child(1) ul.ecl-mega-menu__sublist li a span.ecl-icon.ecl-icon--2xs.ecl-link__icon.ecl-icon--external': 6
+      'li.ecl-mega-menu__item:nth-child(1) ul.ecl-mega-menu__sublist li a span.ecl-icon.ecl-icon--2xs.ecl-link__icon.ecl-icon--external': 2
```
22Alexandra, line 4715: This parent has only 2 external links but the html of @ecl/mega-menu/mega-menu-item.html.twig for level 2 (recursive call for items with grandchildren like Child 2.1, Child 2.3), the `<div class="ecl-mega-menu__mega">` was never properly closed (the `{% if level < 2 %}` block was the only place that added the close).
So, the counts were for all external icons. Same for the cases below.

composer.json
```diff
@@ -8,7 +8,8 @@
     "require": {
         "php": ">=8.1",
         "cweagans/composer-patches": "^2.0",
-        "drupal/core": "^10.3 || ^11",
+        "drupal/core": "~11.2.0",
```
intelektron: We have to remember to remove it before merge

intelektron (dismissed):

composer.json
22Alexandra: I committed by mistake the composer.json changes. I'll drop those.

sergepavle (approved):

intelektron (approved):

PR #1734, opened 2026-05-08
EWPP-6850: Add Animated numbers pattern.

22Alexandra:


templates/patterns/animated_numbers/pattern-animated-numbers.html.twig
```diff
@@ -0,0 +1,49 @@
...
+    value: item.value,
+    suffix: item.suffix,
+    description: item.description,
+  }]) %}
+{% endfor %}
+
+{% include '@ecl-twig/animated-numbers' with {
+  with_background: with_background ?? false,
+  border: border ?? false,
+  counter_color: counter_color ?? true,
```
sergepavle, line 45: Why not to use `|default()` syntax?

22Alexandra: Because when passing false from a template that calls the pattern, the |default filter considers it empty and it would set the default value to true instead of it being false as intended.

22Alexandra: Check here https://twig.symfony.com/doc/3.x/filters/default.html

sergepavle: ok

sergepavle (dismissed):

templates/patterns/animated_numbers/animated_numbers.ui_patterns.yml
```diff
@@ -0,0 +1,100 @@
...
+    border:
+      type: "boolean"
+      label: "Border left"
+      description: "Whether the items should have a left border or not (default false)."
+      preview: true
+      escape: false
+    sources_label:
+      type: "text"
+      label: "Sources label"
+      description: "The label of sources. Defaults to 'Sources'."
```
fmmribeiro: maybe to picky missing ':'
```suggestion
      description: "The label of sources. Defaults to 'Sources:'."
```

PR #1740, opened 2026-05-13
OEL-4458: Fix handling of multiple site owners.

donquixote:
There are different issues with the current implementation.

I am only going to point out one here.

Steps to reproduce:
- Add multiple site owners.
- Switch between eu, ec, standardised, core.

In one of them, we will see raw html in the text output.
This is also shown by the test, where we now replace `->getText()` with `->getHtml()` in the assertion.

### DIFF
--- a/templates/navigation/oe-corporate-blocks-ec-footer--core.html.twig
+++ b/templates/navigation/oe-corporate-blocks-ec-footer--core.html.twig
@@ -12,7 +12,10 @@
 #}
 {# Populate the site info section. #}
 {% if site_owner and site_owner|length > 1 %}
-  {% set site_owner_description = 'This site is co-managed by:<br />'|t ~ site_owner|join('<br />') %}
+  {% set site_owner_description %}
+    {{- 'This site is co-managed by:<br />'|t }}
+    {{- site_owner|safe_join('<br />') }}
+  {%- endset %}
 {% else %}
   {% set site_owner_description = site_owner ? 'This site is managed by:<br />@name'|t({'@name': site_owner|first}) %}
 {% endif %}
--- a/tests/src/Functional/CorporateFooterRenderTest.php
+++ b/tests/src/Functional/CorporateFooterRenderTest.php
@@ -326,7 +326,7 @@ class CorporateFooterRenderTest extends BrowserTestBase {
     ], 'EU Site Name');
     $this->drupalGet('<front>');
     $actual = $assert->elementExists('css', 'div.ecl-site-footer__description');
-    $this->assertEquals('This site is co-managed by:<br />Directorate-General for Budget<br />Directorate-General for Economic and Financial Affairs', $actual->getText());
+    $this->assertEquals('This site is co-managed by:<br>Directorate-General for Budget<br>Directorate-General for Economic and Financial Affairs', $actual->getHtml());
 
     // Test European Union footer standardised block rendering.
     $this->branding = 'standardised';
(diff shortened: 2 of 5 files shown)


templates/navigation/oe-corporate-blocks-ec-footer--core.html.twig
```diff
@@ -12,9 +12,12 @@
...
 {# Populate the site info section. #}
 {% if site_owner and site_owner|length > 1 %}
-  {% set site_owner_description = 'This site is co-managed by:<br />'|t ~ site_owner|join('<br />') %}
+  {% set site_owner_description %}
+    {{- 'This site is co-managed by:<br>'|t }}
+    {{- site_owner|safe_join('<br>') }}
+  {%- endset %}
 {% else %}
-  {% set site_owner_description = site_owner ? 'This site is managed by:<br />@name'|t({'@name': site_owner|first}) %}
+  {% set site_owner_description = site_owner ? 'This site is managed by:<br>@name'|t({'@name': site_owner|first}) %}
```
donquixote: Now that we already change the translatable strings by replacing `<br />` with `<br>`, we can also review which part exactly should be translatable.

E.g. `'This site is managed by:<br>@name'|t` and `'This site is co-managed by:<br>@names'|t`
vs just `'This site is managed by:'|t` and `'This site is co-managed by:'|t`

templates/navigation/oe-corporate-blocks-ec-footer--core.html.twig
```diff
@@ -12,9 +12,12 @@
 #}
 {# Populate the site info section. #}
 {% if site_owner and site_owner|length > 1 %}
-  {% set site_owner_description = 'This site is co-managed by:<br />'|t ~ site_owner|join('<br />') %}
+  {% set site_owner_description %}
+    {{- 'This site is co-managed by:<br>'|t }}
+    {{- site_owner|safe_join('<br>') }}
```
donquixote: The result of `|safe_join(..)` must be printed and then captured to a variable.
It cannot be assigned with `=` or passed as replacement argument to `|t(..)`, or it will lose its being marked as safe, and everything will be escaped again.

templates/navigation/oe-corporate-blocks-ec-footer--core.html.twig
donquixote: Or we could restore the old translatable string and keep the `<br />` at least in that part, if a goal is to keep translations.

22Alexandra: What's the real benefit of changing the `<br />` with `<br>`? Even core still uses `<br />`, so we can keep it like that and avoid updating all the translations unless there is an actual issue or requirement.

templates/navigation/oe-corporate-blocks-ec-footer--core.html.twig
22Alexandra: Replacing the `join` filter with the `safe_join` one could be worth it but the whole set block seems more like matter of style or readability in the template.

22Alexandra:
If there is an actual issue generated by these templates which require the fixes in this PR, please rebase the PR and do the necessary updates in the translation files so the translations still work.
Also, if translations get adapted, consider moving the `<br>` outside the string.

If there is no issue generated by these templates, we can consider not updating the strings because 5.x will shortly stop receiving support but bug fixes only for a while. 

The 6.x branch is the new default branch and the markup might get soon changed to the current ECL markup (span for a single site owner and an unordered list for the multiple site owners).

templates/navigation/oe-corporate-blocks-ec-footer--core.html.twig
donquixote: The "correct" html would be `<br>`, because it's not xml.
But it's ok to leave it if the translatable string remains unchanged otherwise.
I think I mostly did it here because it would feel inconsistent if for the other case we use `<br>`.

sergepavle:
Looks good. Needs to rebase and fix failures in test.

22Alexandra:
This PR contains many changes from the 6.x branch that should not be ported to 5.x.
@donquixote please rebase the PR from 5.x branch so that includes only the changes related to it.

tests/src/Functional/CorporateFooterRenderTest.php
```diff
@@ -321,7 +321,7 @@ public function testCorporateFooterRendering(): void {
     ], 'EU Site Name');
     $this->drupalGet('<front>');
     $actual = $assert->elementExists('css', 'div.ecl-site-footer__description');
-    $this->assertEquals('This site is co-managed by:<br />Directorate-General for Budget<br />Directorate-General for Economic and Financial Affairs', $actual->getText());
+    $this->assertEquals('This site is co-managed by:<br>Directorate-General for Budget<br>Directorate-General for Economic and Financial Affairs', $actual->getHtml());
```
donquixote, line 329: The DOMDocument parser used in `->getHtml()` converts `<br />` to `<br>`.
In the html source we do have `<br />` because we wanted to keep the original translatable strings.

intelektron (approved):

22Alexandra (approved):

22Alexandra (approved):

nagyad (approved):

PR #1769, opened 2026-06-18
EWPP-6952: Better focal point sup in Banners - ECL wrappers and image_anchor

fmmribeiro:


modules/oe_theme_helper/oe_theme_helper.post_update.php
```diff
@@ -820,3 +820,145 @@ function oe_theme_helper_post_update_60001(): void {
...
+  }
+
+  // Add new banner styles.
+  $image_styles = [
+    'oe_theme_full_width_medium' => [
+      'label' => 'Full width medium (1920)',
+      'width' => '1920',
+      'height' => '1920',
+    ],
+    'oe_theme_full_with_small' => [
```
22Alexandra: I think this is a typo and it should have been "width".

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -622,32 +614,16 @@ public function testBanner(): void {
...
-    $this->assertCount(1, $extra_large_media);
-    $this->assertStringContainsString('oe_theme_extra_large_3_1_banner/public/example_1_en.jpeg', $extra_large_media->attr('srcset'));
-    $extra_extra_large_media = $crawler->filter('section.ecl-banner.ecl-banner--font-m.ecl-banner--box-bg-light.ecl-banner--color-dark.ecl-banner--h-center picture source[media="all and (min-width: 1140px)"]');
-    $this->assertCount(1, $extra_extra_large_media);
-    $this->assertStringContainsString('oe_theme_extra_extra_large_4_1_banner/public/example_1_en.jpeg', $extra_extra_large_media->attr('srcset'));
-    $full_width_media = $crawler->filter('section.ecl-banner.ecl-banner--font-m.ecl-banner--box-bg-light.ecl-banner--color-dark.ecl-banner--h-center picture source[media="all and (min-width: 1368px)"]');
-    $this->assertCount(1, $full_width_media);
-    $this->assertStringContainsString('oe_theme_full_width_banner_4_1/public/example_1_en.jpeg', $full_width_media->attr('srcset'));
+
+    // francisco: media breakpoints????
```
22Alexandra: I guess this is either WIP or leftover.

oe_theme.theme
```diff
@@ -2190,7 +2184,7 @@ function oe_theme_preprocess_paragraph__oe_carousel(array &$variables): void {
       continue;
     }
 
-    $responsive_image = _oe_theme_load_from_responsive_image_style_id("oe_theme_{$banner_type}_banner", $uri);
+    $responsive_image = _oe_theme_load_from_responsive_image_style_id("oe_theme_banner", $uri);
```
22Alexandra, line 2194:
```suggestion
    $responsive_image = _oe_theme_load_from_responsive_image_style_id('oe_theme_banner', $uri);
```

This change makes the lines $banner_type redundant so the following code should be removed:
```
// Load needed images styles based on carousel size.
  $banner_type = match ($variables['size']) {
    's' => 'small',
    'm' => 'medium',
    'l' => 'large',
  };
```

modules/oe_theme_helper/oe_theme_helper.post_update.php
```diff
@@ -820,3 +820,145 @@ function oe_theme_helper_post_update_60001(): void {
...
+    if ($style) {
+      continue;
+    }
+    // Create image style.
+    $image_style = ImageStyle::create([
+      'name' => $style_id,
+      'label' => $style_data['label'],
+    ]);
+    // Add scale effect to the image style.
+    $effect = [
```
22Alexandra, line 928: Please add the upscale to false the same like in config/install. False is the default value but we need to be consistent between update and install.

modules/oe_theme_helper/oe_theme_helper.post_update.php
```diff
@@ -820,3 +820,145 @@ function oe_theme_helper_post_update_60001(): void {
...
+    $image_style->addImageEffect($effect);
+    $image_style->save();
+    \Drupal::logger('oe_theme_helper')->notice('Created image style "@id" with label "@label".', [
+      '@id' => $style_id,
+      '@label' => $style_data['label'],
+    ]);
+  }
+
+  // Create responsive image styles for the new banner styles.
+  $config_path = \Drupal::service('extension.list.module')->getPath('oe_theme_helper') . '/config/install';
```
22Alexandra: Let's aim to be consistent in the image styles and responsive image styles creation. Above we use entity API while here the config files are used, why?

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
22Alexandra: Since we don't use different image styles per Banner size, I think it's important to add assertions confirming the same image styles are used for each banner size.

components/banners/banner.twig
```diff
@@ -8,6 +8,7 @@
 {% set _link = [] %}
 {% set _title = [] %}
 {% set _description = [] %}
+{% set _image_anchor = image_anchor|default('50% 50%') %}
```
22Alexandra, line 11: Please add test coverage for this in rendering.yml file of fixtures for components.

22Alexandra (changes requested):

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -1153,15 +1135,12 @@ public function testCarousel(): void {
       'items' => [
         [
           'title' => 'Item 1',
-          'image' => $this->container->get('file_url_generator')->transformRelative($en_file_1_url),
+          'image' => 'oe_theme_full_width/public/example_1_en.jpeg',
```
22Alexandra, line 1129: There are some lines above of unused code that has to be cleaned up. Lines 1034 through 1054.

modules/oe_theme_helper/oe_theme_helper.post_update.php
```diff
@@ -820,3 +820,166 @@ function oe_theme_helper_post_update_60001(): void {
...
+    ],
+    'oe_theme_full_width_small' => [
+      'label' => 'Full width small (768)',
+      'width' => '768',
+      'height' => '768',
+      'upscale' => FALSE,
+    ],
+  ];
+
+  $image_style_storage = \Drupal::entityTypeManager()->getStorage('image_style');
```
22Alexandra: This variable already exists with this value added previously on line 881.

tests/src/Kernel/fixtures/rendering.yml
```diff
@@ -387,19 +381,17 @@
       size: "l"
       font_color: light
       box_background: dark
+      image_anchor: '35% 50%'
```
22Alexandra, line 384: This is not asserted.

22Alexandra (changes requested):

22Alexandra (dismissed):
The .info file should contain the new image styles created and the deprecated ones have to be marked as such in readme.

sergepavle (approved):

intelektron (approved):

PR #1774, opened 2026-06-25
EWPP-7015: Add section_feedback to footer templates.

bircher:


tests/src/Functional/CorporateFooterRenderTest.php
```diff
@@ -934,6 +936,61 @@ public function testCorporateFooterRendering(): void {
...
+    $this->library = $library;
+    $this->branding = $branding;
+    $this->configFactory->getEditable('oe_theme.settings')->set('component_library', $this->library)->save();
+    $this->configFactory->getEditable('oe_theme.settings')->set('branding', $this->branding)->save();
+
+    // Set up a realistic test footer.
+    $data = $this->getFixtureContent($this->library . '_footer.yml');
+    $this->overrideCorporateBlocksFooter($this->library, $data);
+
+    $description = 'Help us improve our ' . $this->library . '-' . $this->branding . ' website';
```
22Alexandra: Why can't we add the testing data inside the `ec_footer.yml` and `eu_footer.yml` fixtures so we don't overcomplicate this and then improve the test above to assert this? There is no need to trigger the preprocess hook.

22Alexandra (changes requested):

22Alexandra (approved):

sergepavle (approved):

22Alexandra (approved):

sergepavle (approved):

PR #1790, opened 2026-08-05
EWPP-7119: Set the configured theme component library to the Webtools smart loader.

22Alexandra:


nagyad (dismissed):

modules/oe_theme_helper/oe_theme_helper.module
```diff
@@ -72,6 +73,69 @@ function oe_theme_helper_page_attachments(array &$page) {
...
+/**
+ * Form submit handler for the theme settings form.
+ *
+ * @param array $form
+ *   The form.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state.
+ */
+function oe_theme_helper_component_library_submit(array &$form, FormStateInterface $form_state): void {
+  if ($form_state->getValue('component_library') === $form['component_library']['#default_value']) {
```
fmmribeiro, line 132: this will only invalidate when swithing from EC => EU but not EU => EC

fmmribeiro: my mistake default_values changes

fmmribeiro (dismissed):

modules/oe_theme_helper/oe_theme_helper.module
22Alexandra: The default_value gets the current component_library setting, so it does work.
EC !== EU is the same thing with EU !== EC

PR #1794, opened 2026-08-12
EWPP-7197: Removing behat and refactoring tests.

upchuk:


tests/src/Functional/CorporateBlocksTest.php
```diff
@@ -0,0 +1,125 @@
...
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  protected $configFactory;
+
+  /**
+   * {@inheritdoc}
+   *
+   * Kept identical to CorporateFooterRenderTest and SiteBrandingTest so the
+   * three share the same CachedDatabaseInstallTrait fingerprint and reuse a
+   * single cached database install.
```
22Alexandra, line 32: This test class and `SiteBrandingTest` are in batch6 but `CorporateFooterRenderTest` is in batch8, so do they actually keep the same fingerprint?

tests/src/Functional/PageHeaderTest.php
```diff
@@ -0,0 +1,148 @@
...
+    $this->drupalGet('user/register');
+    $this->assertPageHeaderTitle('Create new account');
+    // The site identity is not shown on the standard page header.
+    $this->assertSession()->elementNotExists('css', '.ecl-page-header h2.ecl-u-type-heading-2');
+    $this->assertBreadcrumbTrail(['Home']);
+    $this->assertBreadcrumbActiveElement('Create new account');
+
+    // Editing the title updates the page header accordingly.
+    $node = $this->getNodeByTitle('Robots are everywhere');
+    $node->setTitle('Robots are everywhere nowadays')->save();
```
22Alexandra, line 70: We should do it through the form like in behat cause the cache invalidation is not tested like this, no?

tests/src/Functional/ThemeShowcaseTest.php
```diff
@@ -0,0 +1,129 @@
...
+    $assert->elementExists('css', '.ecl-site-header--has-menu .ecl-menu');
+  }
+
+  /**
+   * Asserts that the site header matches the given ECL branding.
+   *
+   * @param string $branding
+   *   The ECL branding, either 'core' or 'standardised'.
+   */
+  protected function assertSiteHeader(string $branding): void {
```
22Alexandra: `$branding` is not used here. I know it's the same like the Behat step but we should probably drop it if there is no difference anymore between the brandings.

tests/src/Functional/ThemeShowcaseTest.php
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'block',
+    'oe_search',
+    'oe_multilingual',
+    'oe_theme_helper',
+    'oe_theme_demo',
+    'oe_theme_helper',
```
22Alexandra: Duplicates line 33.
```suggestion
```

tests/src/Kernel/RenderingOneTest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_theme\Kernel;
+
+/**
+ * Tests rendering of elements (first third of the fixtures).
```
22Alexandra:
```suggestion
 * Tests rendering of elements (first quarter of the fixtures).
```

tests/src/Kernel/RenderingTwoTest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_theme\Kernel;
+
+/**
+ * Tests rendering of elements (second third of the fixtures).
```
22Alexandra:
```suggestion
 * Tests rendering of elements (second quarter of the fixtures).
```

tests/src/Kernel/RenderingThreeTest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_theme\Kernel;
+
+/**
+ * Tests rendering of elements (third third of the fixtures).
```
22Alexandra:
```suggestion
 * Tests rendering of elements (third quarter of the fixtures).
```

tests/src/FunctionalJavascript/ThemeShowcaseTest.php
```diff
@@ -0,0 +1,204 @@
...
+  public function testJsThemeShowcase(): void {
+    // Use the European Union component library so the language selector button
+    // shows the full language name.
+    $this->configFactory->getEditable('oe_theme.settings')->set('component_library', 'eu')->save();
+
+    // Force a desktop viewport: below the ECL desktop breakpoint the navigation
+    // collapses into the mobile hamburger menu, where the top-level links are
+    // hidden and cannot be hovered. For some reason, it doesn't get set by
+    // phpunit at this size.
+    $this->getSession()->resizeWindow(1920, 1080);
```
22Alexandra, line 84: Yeah, the parent class sets it to a different value in `initFrontPage`:
Image

So, let's add the class and do the resizing there.

tests/src/Functional/ThemeShowcaseTest.php
```diff
@@ -0,0 +1,129 @@
...
+
+    $about = $assert->elementExists('css', '.ecl-menu__item:nth-child(2) .ecl-menu__mega');
+    foreach (['Commission at work', 'Departments'] as $link) {
+      $this->assertNotNull($about->findLink($link), sprintf('Link "%s" not found in the about dropdown menu.', $link));
+    }
+
+    // Changing the ECL branding switches the site header style. The demo
+    // defaults to the standardised branding.
+    $this->assertSiteHeader('standardised');
+    $assert->elementExists('css', '.ecl-site-header--has-menu .ecl-menu');
```
22Alexandra: I guess this can be moved inside the `assertSiteHeader` method.

tests/src/Functional/CorporateBlocksTest.php
```diff
@@ -0,0 +1,125 @@
...
+
+    $pages = [
+      '<front>',
+      'user/register',
+    ];
+
+    foreach ($pages as $page) {
+      // By default, the European Commission footer is displayed.
+      $this->drupalGet($page);
+      $this->assertCorporateFooter('https://commission.europa.eu/index_en', 'European Commission logo');
```
22Alexandra, line 80: The footer is already tested in here no? \Drupal\Tests\oe_theme\Functional\CorporateFooterRenderTest::testCorporateFooterRendering

tests/src/Functional/SiteBrandingTest.php
```diff
@@ -0,0 +1,185 @@
...
+    $this->assertStringContainsString('oe_theme/dist/eu/images/logo/condensed-version/positive/logo-eu--' . $langcode . '.svg', $image->getAttribute('src'));
+  }
+
+  /**
+   * Sets the theme's component library.
+   *
+   * @param string $component_library
+   *   The component library machine name, either 'ec' or 'eu'.
+   */
+  protected function setComponentLibrary(string $component_library): void {
```
22Alexandra, line 125: We should move this into a trait and reuse it cause in `CorporateBlocksTest`, `SiteHeaderTest` and `ThemeShowcaseTest` you do it inline multiple times.

tests/src/Functional/SiteBrandingTest.php
```diff
@@ -0,0 +1,185 @@
...
+      'en' => 'https://european-union.europa.eu/index_en',
+      'bg' => 'https://european-union.europa.eu/index_bg',
+      'fr' => 'https://european-union.europa.eu/index_fr',
+      // Non-EU language falls back to the language-neutral link.
+      'is' => 'https://european-union.europa.eu',
+    ];
+    foreach ($eu_logos as $langcode => $link) {
+      $this->drupalGetInLanguage('<front>', $langcode);
+      $this->assertHeaderLogo($link, 'Home - European Union', 'European Union flag', 'European Union');
+      $this->assertLocalisedEuLogos($langcode);
```
22Alexandra: The logo is asserted here \Drupal\Tests\oe_theme\Functional\SiteHeaderTest::testSiteHeader although not localized but still not worth the duplication. I would assert it in the existing class localized instead.

22Alexandra (changes requested):

tests/src/Functional/CorporateBlocksTest.php
upchuk: Yes yes, the files are shared, they are on disk, all batches use them.

tests/src/Functional/PageHeaderTest.php
upchuk: Cache gets invalidated also via the API, not just the form

tests/src/FunctionalJavascript/ThemeShowcaseTest.php
upchuk: No need, we can keep everything as standard and resize where needed. Otherwise, to be consistent, we'd have to resize in all our test classes and it wasnt needed.

tests/src/Functional/CorporateBlocksTest.php
upchuk: Perhaps, but difficult to say if entirely. In any case, it's a simple subset of assertions, doesn;t impact the timing or memory

tests/src/Functional/SiteBrandingTest.php
upchuk: Boh, it's a simple config set, might be overkill, lots of other config is being set in various tests that can repeat.

22Alexandra (dismissed):

tests/src/FunctionalJavascript/ThemeShowcaseTest.php
```diff
@@ -0,0 +1,204 @@
...
+    $assert = $this->assertSession();
+    $page = $this->getSession()->getPage();
+
+    // The navigation menu features dropdown menus. The submenu links are not
+    // visible by default.
+    $this->assertLinkNotVisible('Commission at work');
+    $this->assertLinkNotVisible('Democratic change');
+
+    // Hovering over "About" reveals its submenu, but not the priorities one.
+    $page->findLink('About')->mouseOver();
```
nagyad, line 95: In the behat, we have 1 sec of waiting time for ECL's .25s mega-menu transition delay. This could give us false negatives.

modules/oe_theme_content_event/tests/src/FunctionalJavascript/InfoDisclosureExtraFieldTest.php
```diff
@@ -32,15 +32,45 @@ class InfoDisclosureExtraFieldTest extends WebDriverTestBase {
 
   /**
    * {@inheritdoc}
+   *
+   * This intentionally mirrors the module set of
+   * \Drupal\Tests\oe_theme\Functional\ContentRenderTestBase so this test shares
+   * that suite's cached database install. Keep the two lists in sync; if they
+   * diverge, this test installs its own DB.
```
nagyad, line 39: The CachedDatabaseInstallTrait computes the fingerprint first, then calls parent::installDrupal() during that call core's WebDriverTestBase appends js_testing_ajax_request_test etc. to the static $modules. 
So even with perfectly synced lists, the JS test installs more modules than its fingerprint claims. Every test that runs after it in the same batch will have a polluted list, misses the cache and silently reinstalls. 

Also the other way around, if a Functional test writes the dump first, the JS test restores a site that is missing the js_testing_* modules core expects to be installed.

nagyad (changes requested):

modules/oe_theme_content_event/tests/src/FunctionalJavascript/InfoDisclosureExtraFieldTest.php
nagyad: 1. We should never share dump between JS and Functional tests. 
2. In CachedDatabaseInstallTrait We should reset the polluted static before every install.

tests/src/FunctionalJavascript/ThemeShowcaseTest.php
upchuk: Hasn't been a problem so far.

modules/oe_theme_content_event/tests/src/FunctionalJavascript/InfoDisclosureExtraFieldTest.php
upchuk: Indeed. However, I dont agree we shouldnt share the DBs between types of suites, defeats the whole purpose, we get too many install. I just made sure we install also the JS modules if the functional test is the first to set the DB and ensure both use the same fingerprint, so no static pollution.

nagyad:
Remove the debug commit from the list, please.

runner.yml.dist
```diff
@@ -113,12 +110,24 @@ toolkit:
...
           # Build ECL twig templates if not present. In CI, each batch test job runs in a
           # separate container and only receives the build/ artifacts, so ecl_components/
           # (gitignored) must be rebuilt before tests can resolve @ecl-twig/... includes.
           - { task: "exec", command: "[ -d ecl_components ] || bash ./scripts/build-ecl-pipeline.sh" }
+          # Wait until the shared MySQL server is up AND the "drupal" database is
+          # queryable before launching the parallel batches. This is to ensure that Kernel tests have
+          # a database if they are first in the batch.
+          - { task: "exec", command: "for i in $(seq 1 30); do mysql -h mysql -P 3306 -u root -e 'USE drupal; SELECT 1' >/dev/null 2>&1 && break; echo \"waiting for drupal database ($i/30)\"; sleep 2; done; mysql -h mysql -P 3306 -u root -e 'USE drupal; SELECT 1'" }
 
   ci:
```
hernani: This can be removed now

composer.json
```diff
@@ -107,7 +104,6 @@
         "_readme": [
             "autoload-dev: explicitly mapping EntityReferenceFieldCreationTrait to prevent failures.",
             "Locking extra_field to 3.0.0-beta1 due to issues, will be fixed in EWPP-6001.",
-            "Locking drupal/datetime_testing to 1.0.0-beta6 since beta7 removed a behat step still required in the project.",
             "Locking openeuropa/oe_content to 4.14.0 since behat got replaced with phpunit in latest."
```
hernani: Shouldn't be unlocked now?

hernani (approved):

PR #1805, opened 2026-09-10
EWPP-7220: Do not render the Subject field values on Event anymore.

22Alexandra:


sergepavle (approved):
Not sure if it is fully safe to hide the subject field at the component level, as it could impact non-EWPP sites. But suppose it is fine.

intelektron (approved):

